PR #9, opened 2021-01-29
Move all optimizers to Optimisers.jl

darsnack:
This PR finishes the already good work in this repo to move all the optimizers from Flux to Optimisers.jl. The main advantage of the implementation in Optimisers.jl is the use of functors and explicit state. This eliminates the need for any `IdDict`s and simplifies how compositions of basic optimizers work.

~~One notable exception: this implementation is mutable. The simple reason for this is that immutable updates will copy entire models and result in high memory usage. Until we have a compiler optimization to address this, this PR opts to have mutable updates for now. You can read more about the reasoning [in this design note](https://github.com/FluxML/ML-Coordination-Tracker/discussions/22).~~

Remaining tasks:
- [x] docstrings
- [x] clean up tests

cc @DhairyaLGandhi @CarloLucibello @mcabbott @ToucheSir

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -7,6 +7,6 @@ include("rules.jl")
 
 export Descent, ADAM, Momentum, Nesterov, RMSProp,
        ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
-       WeightDecay, ChainOptimiser
+       WeightDecay, OptimiserChain
 
 end # module
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -7,6 +7,6 @@ include("rules.jl")
 
 export Descent, ADAM, Momentum, Nesterov, RMSProp,
        ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
-       WeightDecay, Sequence
+       WeightDecay, SequenceOptimiser
 
 end # module
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -6,7 +6,8 @@ include("interface.jl")
 include("rules.jl")
 
 export init, update!
-export Descent, Momentum, Nesterov, RMSProp,
-       ADAM
+export Descent, ADAM, Momentum, Nesterov, RMSProp,
+       ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
+       WeightDecay, Sequence
 
 end # module
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -5,7 +5,6 @@ using Functors: functor, fmap, isleaf
 include("interface.jl")
 include("rules.jl")
 
-export init, update!
 export Descent, ADAM, Momentum, Nesterov, RMSProp,
        ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
        WeightDecay, Sequence
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -7,6 +7,6 @@ include("rules.jl")
 
 export Descent, ADAM, Momentum, Nesterov, RMSProp,
        ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
-       WeightDecay, SequenceOptimiser
+       WeightDecay, ChainOptimiser
 
 end # module
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -5,6 +5,7 @@ using Functors: functor, fmap, isleaf
 include("interface.jl")
 include("rules.jl")
 
+export init, update!
 export Descent, Momentum, Nesterov, RMSProp,
        ADAM
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -13,7 +13,7 @@ function init(o, x)
 end
 
 function _update!(o, x, x̄, st)
-  x̄, st = apply!(o, x̄, st)
+  x̄, st = apply!(o, x, x̄, st)
   return patch!(x, x̄), st
 end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,7 +1,4 @@
-function patch!(x, x̄)
-  @. x .-= x̄
-  return x
-end
+patch(x, x̄) = x .- x̄
 
 function init(o, x)
   if isleaf(x)
@@ -12,20 +9,20 @@ function init(o, x)
   end
 end
 
-function _update!(o, x, x̄, st)
-  x̄, st = apply!(o, x, x̄, st)
-  return patch!(x, x̄), st
+function _update(o, x, x̄, st)
+  x̄, st = apply(o, x, x̄, st)
+  return patch(x, x̄), st
 end
 
-function update!(o, x::T, x̄, state) where T
+function update(o, x::T, x̄, state) where T
   if x̄ === nothing
     return x, state
   elseif isleaf(x)
-    _update!(o, x, x̄, state)
+    _update(o, x, x̄, state)
   else
     x̄, _  = functor(typeof(x), x̄)
     x, restructure = functor(typeof(x), x)
-    xstate = map((x, x̄, state) -> update!(o, x, x̄, state), x, x̄, state)
+    xstate = map((x, x̄, state) -> update(o, x, x̄, state), x, x̄, state)
     restructure(map(first, xstate)), map(x -> x[2], xstate)
   end
 end
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,11 +1,11 @@
 patch(x, x̄) = x .- x̄
 
-function init(o, x)
+function state(o, x)
   if isleaf(x)
     return init(o, x)
   else
     x, _ = functor(x)
-    return map(x -> init(o, x), x)
+    return map(x -> state(o, x), x)
   end
 end
 
@@ -18,11 +18,11 @@ function update(o, x::T, x̄, state) where T
   if x̄ === nothing
     return x, state
   elseif isleaf(x)
-    _update(o, x, x̄, state)
+    return _update(o, x, x̄, state)
   else
     x̄, _  = functor(typeof(x), x̄)
     x, restructure = functor(typeof(x), x)
     xstate = map((x, x̄, state) -> update(o, x, x̄, state), x, x̄, state)
-    restructure(map(first, xstate)), map(x -> x[2], xstate)
+    return restructure(map(first, xstate)), map(x -> x[2], xstate)
   end
 end
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,30 +1,31 @@
-function patch(x, x̄)
-  return x .- x̄
+function patch!(x, x̄)
+  @. x .-= x̄
+  return x
 end
 
-function state(o, x)
+function init(o, x)
   if isleaf(x)
     return init(o, x)
   else
     x, _ = functor(x)
-    map(x -> state(o, x), x)
+    return map(x -> init(o, x), x)
   end
 end
 
-function _update(o, x, x̄, st)
-  x̄, st = apply(o, x, x̄, st)
-  return patch(x, x̄), st
+function _update!(o, x, x̄, st)
+  x̄, st = apply!(o, x̄, st)
+  return patch!(x, x̄), st
 end
 
-function update(o, x::T, x̄, state) where T
+function update!(o, x::T, x̄, state) where T
   if x̄ === nothing
     return x, state
   elseif isleaf(x)
-    _update(o, x, x̄, state)
+    _update!(o, x, x̄, state)
   else
     x̄, _  = functor(typeof(x), x̄)
-    x, re = functor(typeof(x), x)
-    xstate = map((x, x̄, state) -> update(o, x, x̄, state), x, x̄, state)
-    re(map(first, xstate)), map(x -> x[2], xstate)
+    x, restructure = functor(typeof(x), x)
+    xstate = map((x, x̄, state) -> update!(o, x, x̄, state), x, x̄, state)
+    restructure(map(first, xstate)), map(x -> x[2], xstate)
   end
 end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -430,7 +430,7 @@ weight decay regularization.
                          (no need to change default)
 """
 ADAMW(η = 1f-3, β = (9f-1, 9.99f-1), γ = 0, ϵ = eps(typeof(η))) =
-  SequenceOptimiser(ADAM(η, β, ϵ), WeightDecay(γ))
+  ChainOptimiser(ADAM(η, β, ϵ), WeightDecay(γ))
 
 """
     AdaBelief(η = 1f-3, β = (9f-1, 9.99f-1), ϵ = eps(typeof(η)))
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -1,5 +1,5 @@
 """
-    Descent(; η = 0.1)
+    Descent(η = 0.1)
 
 Classic gradient descent optimiser with learning rate `η`.
 For each parameter `p` and its gradient `dp`, this runs `p -= η*dp`.
@@ -11,7 +11,7 @@ For each parameter `p` and its gradient `dp`, this runs `p -= η*dp`.
 struct Descent{T}
   eta::T
 end
-Descent(; η = 0.1) = Descent(η)
+Descent(η = 0.1) = Descent(η)
 
 init(o::Descent, x::AbstractArray) = nothing
 
@@ -25,7 +25,7 @@ end
 (o::Descent)(m, dm, st) = update(o, m, dm, st)
 
 """
-    Momentum(; η = 0.01, ρ = 0.9)
+    Momentum(η = 0.01, ρ = 0.9)
 
 Gradient descent optimizer with learning rate `η` and momentum `ρ`.
 
@@ -39,7 +39,7 @@ struct Momentum{T}
   eta::T
   rho::T
 end
-Momentum(; η = 0.01, ρ = 0.9) = Momentum(η, ρ)
+Momentum(η = 0.01, ρ = 0.9) = Momentum(η, ρ)
 
 init(o::Momentum, x::AbstractArray) = (velocity = zero(x),)
 
@@ -53,7 +53,7 @@ end
 (o::Momentum)(m, dm, state) = update(o, m, dm, state)
 
 """
-    Nesterov(; η = 0.001, ρ = 0.9)
+    Nesterov(η = 0.001, ρ = 0.9)
 
 Gradient descent optimizer with learning rate `η` and Nesterov momentum `ρ`.
 
@@ -67,7 +67,7 @@ struct Nesterov{T}
   eta::T
   rho::T
 end
-Nesterov(; η = 0.001, ρ = 0.9) = Nesterov(η, ρ)
+Nesterov(η = 0.001, ρ = 0.9) = Nesterov(η, ρ)
 
 init(o::Nesterov, x::AbstractArray) = (velocity = zero(x),)
 
@@ -82,7 +82,7 @@ function apply(o::Nesterov, x, dx, state)
 end
 
 """
-    RMSProp(; η = 0.001, ρ = 0.9, ϵ = eps(typeof(η)))
+    RMSProp(η = 0.001, ρ = 0.9, ϵ = eps(typeof(η)))
 
 Optimizer using the
 [RMSProp](https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf)
@@ -102,7 +102,7 @@ struct RMSProp{T}
   rho::T
   epsilon::T
 end
-RMSProp(; η = 0.001, ρ = 0.9, ϵ = eps(typeof(η))) = RMSProp(η, ρ, ϵ)
+RMSProp(η = 0.001, ρ = 0.9, ϵ = eps(typeof(η))) = RMSProp(η, ρ, ϵ)
 
 init(o::RMSProp, x::AbstractArray) = (acceleration = zero(x),)
 
@@ -117,7 +117,7 @@ end
 (o::RMSProp)(m, dm, state) = update(o, m, dm, state)
 
 """
-    ADAM(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
+    ADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
 
 [ADAM](https://arxiv.org/abs/1412.6980) optimiser.
 
@@ -134,7 +134,7 @@ struct ADAM{T}
   beta::Tuple{T, T}
   epsilon::T
 end
-ADAM(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = ADAM(η, β, ϵ)
+ADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = ADAM(η, β, ϵ)
 
 init(o::ADAM, x::AbstractArray) = (moments = (zero(x), zero(x)), decays = o.beta)
 
@@ -152,7 +152,7 @@ function apply(o::ADAM{T}, x, dx, state) where T
 end
 
 """
-    RADAM(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
+    RADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
 
 [Rectified ADAM](https://arxiv.org/abs/1908.03265) optimizer.
 
@@ -169,7 +169,7 @@ struct RADAM{T}
   beta::Tuple{T, T}
   epsilon::T
 end
-RADAM(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = RADAM(η, β, ϵ)
+RADAM(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = RADAM(η, β, ϵ)
 
 init(o::RADAM, x::AbstractArray) = (moments = (zero(x), zero(x)), decays = o.beta, t = 1)
 
@@ -196,7 +196,7 @@ function apply(o::RADAM, x, dx, state)
 end
 
 """
-    AdaMax(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
+    AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η)))
 
 [AdaMax](https://arxiv.org/abs/1412.6980) is a variant of ADAM based on the ∞-norm.
 
@@ -213,7 +213,7 @@ struct AdaMax{T}
   beta::Tuple{T, T}
   epsilon::T
 end
-AdaMax(; η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = AdaMax(η, β, ϵ)
+AdaMax(η = 0.001, β = (0.9, 0.999), ϵ = eps(typeof(η))) = AdaMax(η, β, ϵ)
 
 init(o::AdaMax, x::AbstractArray) = (moments = (zero(x), zero(x)), decays = o.beta)
 
@@ -233,7 +233,7 @@ function apply(o::AdaMax, x, dx, state)
 end
 
 """
-    OADAM(; η = 0.001, β = (0.5, 0.9), ϵ = eps(typeof(η)))
+    OADAM(η = 0.001, β = (0.5, 0.9), ϵ = eps(typeof(η)))
 
 [OADAM](https://arxiv.org/abs/1711.00141) (Optimistic ADAM)
 is a variant of ADAM adding an "optimistic" term suitable for adversarial training.
@@ -251,7 +251,7 @@ struct OADAM{T}
   beta::Tuple{T, T}
   epsilon::T
 end
-OADAM(; η = 0.001, β = (0.5, 0.9), ϵ = eps(typeof(η))) = OADAM(η, β, ϵ)
+OADAM(η = 0.001, β = (0.5, 0.9), ϵ = eps(typeof(η))) = OADAM(η, β, ϵ)
 
 init(o::OADAM, x::AbstractArray) = (moments = (zero(x), zero(x)), decays = o.beta, dx = zero(x))
 
@@ -273,7 +273,7 @@ function apply(o::OADAM, x, dx, state)
 end
 
 """
-    ADAGrad(; η = 0.1, ϵ = eps(typeof(η)))
+    ADAGrad(η = 0.1, ϵ = eps(typeof(η)))
 
 [ADAGrad](http://www.jmlr.org/papers/volume12/duchi11a/duchi11a.pdf) optimizer. It has
 parameter specific learning rates based on how frequently it is updated.
@@ -289,7 +289,7 @@ struct ADAGrad{T}
   eta::T
   epsilon::T
 end
-ADAGrad(; η = 0.1, ϵ = eps(typeof(η))) = ADAGrad(η, ϵ)
+ADAGrad(η = 0.1, ϵ = eps(typeof(η))) = ADAGrad(η, ϵ)
 
 init(o::ADAGrad, x::AbstractArray) = (acceleration = fill!(similar(x), o.epsilon),)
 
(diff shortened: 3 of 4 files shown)


src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+function apply!(o::Descent, x, dx, state)
+  η = convert(eltype(dx), o.eta)
+  dx .*= η
+  
+  return dx, state
 end
 
-function (o::Descent)(m, m̄)
-  update(o, m, m̄, state(o, m))[1]
+(o::Descent)(m, dm, st) = update!(o, m, dm, st)
```
ToucheSir, line 27: Following from https://github.com/FluxML/ML-Coordination-Tracker/discussions/22#discussioncomment-317446, I'm not sure we should make these structs callable at all. Given _none_ of methods do anything more than forward to `update(!)` with the optimizer object itself, why not get rid of them altogether?

ToucheSir (changes requested):

src/rules.jl
darsnack: I agree but left them in the PR so we could discuss and make a decision

darsnack: It also (wrongly) makes it look like the optimizer has completely internal state even though this will still return an updated state.

darsnack:
Sorry for the mass cc's but I figured this was an important enough change that we should have a full discussion

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+  rho::S
 end
+Momentum(;η = 0.01, ρ = 0.9) = Momentum(η, ρ)
 
-function (o::Descent)(m, m̄, st)
-  update(o, m, m̄, st)
+init(o::Momentum, x::AbstractArray) = (velocity = zero(x),)
+
+function apply!(o::Momentum, x, dx, state)
+  η, ρ, v = o.eta, o.rho, state.velocity
```
ToucheSir, line 49: This and many of the subsequent implementations will require type conversion in `apply` like `Descent`.

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+Gradient descent optimizer with learning rate `η` and momentum `ρ`.
+
+# Parameters
+- Learning rate (`η`): Amount by which gradients are discounted before updating
+                       the weights.
+- Momentum (`ρ`): Controls the acceleration of gradient descent in the
+                  prominent direction, in effect dampening oscillations.
+"""
+mutable struct Momentum{T,S}
+  eta::T
```
ToucheSir, line 41: This is more of a question for Flux core, but why are fields spelled out while locals (e.g. in `apply`) use symbols? Flux layers are certainly not shy to use field names like `σ` and `β`.

src/rules.jl
darsnack: Good point. Now that the optimizers are parameterized, I think my preference would be to avoid conversions and make the default type parameter `Float32`.

src/interface.jl
```diff
@@ -1,30 +1,31 @@
-function patch(x, x̄)
-  return x .- x̄
+function patch!(x, x̄)
+  @. x .-= x̄
```
ToucheSir, line 2: Back to the topic of composition and not double-scaling, perhaps LR scaling should be done in or before (i.e. in `_update`) patch? That would ensure it only happens once, while eliminating any risk of it not happening as long as one is using `update` (which 99% of users should be).

src/rules.jl
ToucheSir: I think we'll still need them for reduced and mixed-precision, but perhaps in that case it would be easier to make the optimizers themselves Functors and provide functions like `f64`.

src/interface.jl
darsnack: That would mean removing LR from the optimizers and having a "scale-by optimizer." I have mixed feelings on this issue, but on a more practical note, I would suggest that we resolve it in a separate (later) PR.

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+                       the weights.
+- Momentum (`ρ`): Controls the acceleration of gradient descent in the
+                  prominent direction, in effect dampening oscillations.
+- Machine epsilon (`ϵ::Float32`): Constant to prevent division by zero
+                  (no need to change default)
+"""
+mutable struct RMSProp{T,S}
   eta::T
   rho::S
+  epsilon::Float32
```
darsnack, line 107: I would like to make this parameterized as well and make the default value `eps(T)` instead of hard coded to `1f-8`.

src/interface.jl
ToucheSir: I'm fine with that as long as it's resolved before this gets registered. For example, I remembered that https://github.com/deepmind/optax/blob/master/optax/_src/alias.py#L107 exists, which takes care of most of the concerns around ergonomics we discussed in https://github.com/FluxML/ML-Coordination-Tracker/discussions/22#discussioncomment-269839 but would require some invasive API changes to implement.

src/Optimisers.jl
```diff
@@ -5,7 +5,9 @@ using Functors: functor, fmap, isleaf
 include("interface.jl")
 include("rules.jl")
 
-export Descent, Momentum, Nesterov, RMSProp,
-       ADAM
+export init, update!
```
DhairyaLGandhi, line 8: These names are too generic to export

src/interface.jl
DhairyaLGandhi: We can't do inplace updates

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+end
+
+"""
+    Sequence(opts...)
+
+Compose a sequence of optimisers so that each `opt` in `opts`
+updates the gradient in the order specified.
+"""
+struct Sequence{O}
+  opts::O
```
DhairyaLGandhi, line 509: I'm definitely unsure about the naming of this.

DhairyaLGandhi (commented):
Thanks so much for looking into this!

I haven't gone through the entire pr yet but with optimisers.jl, we can't do inplace updates and have a slightly different design to allow for things like xla (and even the GPU compilation work in Julia) to do more aggressive optimisations.

darsnack:
Okay but I don't want to halt other progress until XLA/GPUCompiler is ready. We could always support both in-place and immutable updates for now.

darsnack:
I'm happy to switch this PR to be completely immutable if we already have something ready to address the allocation issue.

src/rules.jl
darsnack: What you prefer? I have no strong opinions, but I think `Optimiser` was a poor name from Flux.

DhairyaLGandhi:
Supporting both doesn't seem worth it, and the work will immediately be useful in 1.6, so it's not halting progress there. We should be alright

darsnack:
Do you mean optimization passes through AbstractInterpreter?

DhairyaLGandhi:
Not really, I mean the gpu codegen work that has already been merged in base.

darsnack:
Sorry, do you mind linking PRs? So this will make all the out-of-place updates mutating as an optimization for code involving `CuArray`s?

DhairyaLGandhi:
There isn't one pr directly linked with it, unfortunately. I guess I could link to the recent work in [GPU work](https://github.com/JuliaLang/julia/pulls?q=is%3Apr+label%3Agpu) in the language. It's not something that will enable universally better code I don't think, but it fits in line with the programming model of GPUs where you do end up wanting to free up memory as soon as possible.

Allocations can be improved by improving the function calls just as much, so not inherently something to hold this up on.

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+# Parameters
+- Learning rate (`η`): Amount by which gradients are discounted before updating
+                       the weights.
 """
-mutable struct Descent
-  eta::Float64
+mutable struct Descent{T}
+  eta::T
 end
+Descent(;η = 0.1) = Descent(η)
```
CarloLucibello, line 16:
```suggestion
Descent(; η = 0.1) = Descent(η)
```

src/rules.jl
```diff
@@ -1,106 +1,523 @@
...
+  
+  return dx, state
 end
 
-function (o::Descent)(m, m̄)
-  update(o, m, m̄, state(o, m))[1]
+(o::Descent)(m, dm, st) = update!(o, m, dm, st)
+
+"""
+    Momentum(;η = 0.01, ρ = 0.9)
```
CarloLucibello, line 30:
```suggestion
    Momentum(; η = 0.01, ρ = 0.9)
```

CarloLucibello:
This is breaking the current convention of passing the lr as the first positional argument, it is going to break basically all existing code.  Did we already have some discussion on that?

src/rules.jl
DhairyaLGandhi: I had thought of `ComposedOptimiser` but that was too long, but maybe that's not too bad.

DhairyaLGandhi:
I would leave out the changes in interface.jl for the purposes of this PR.

It may make things faster to model structure as in #3

darsnack:
> This is breaking the current convention of passing the lr as the first positional argument, it is going to break basically all existing code. Did we already have some discussion on that?

No, and sorry about that! TBH in all my time using Flux, I never realized that the 1st arg was always LR. I always end up looking it up, because it is easier to double check than two have my training not work cause I set the momentum instead of the LR. Thus, the change to keyword args where it is easier to remember. That being said, I have no strong feelings on this change and happy to go back to avoid breaking things.

----

The changes in interface.jl are mostly the mutating stuff. I feel like we should discuss this, because the key feature of functional optimizers is explicit state not immutability. If there is something ready to roll out in a few weeks that make immutable updates mutating as an optimization, then I'll just drop the changes to interface.jl and make everything immutable. 

But if it will be months before we get such features ready for use, then I would strongly recommend we consider keeping mutability right now. At the end of the day, there is no world in which we can have optimizers that make copies of all the model parameters/gradients (even on Flux#master). It seems silly to not provide what we think is a better optimizer and scheduling just because we also want feature X. Why not make feature X part of a later PR if it will be ready later?

----

Re: ParameterSchedulers.jl: @DhairyaLGandhi you mentioned you wanted to discuss here. What are the design assumptions that need to be addressed? We can fix ParameterSchedulers.jl to account for them.

src/rules.jl
darsnack: The problem I had with `ComoposedOptimiser` or similar names was that it doesn't communicate how the sub-optimizers are composed. Something like `Sequence` makes it clear that the optimizers will be applied sequentially to the gradients.

DhairyaLGandhi:
> explicit state not immutability

Both are

> be months before we get such features ready for use, then I would strongly recommend we consider keeping mutability right now

Not sure which features you're talking about here? But pretty sure immutable updates here is an explicit design, and we would maintain that.

Re: ParameterSchedulers.jl: apologies, I meant that we should move the discussion and implementation from https://github.com/FluxML/Flux.jl/pull/1487 to https://github.com/FluxML/Flux.jl/pull/1481, and (as a follow up) the scheduling interface should be an implementation of the optimisers+training without needing additional handling on either side. I'm other words, we need to generalise the interface around these methods to allow for scheduling.

DhairyaLGandhi:
Either way, the optimisation around mutability is orthogonal, and would not need weeks if it's required urgently.

darsnack:
> Not sure which features you're talking about here?

I'm talking about the feature where `x, st = update(o, x, dx, st)` does not make copies of `x` or `dx` even though `update` is immutable.

> the scheduling interface should be an implementation of the optimisers+training without needing additional handling on either side. I'm other words, we need to generalise the interface around these methods to allow for scheduling.

The explicit state is that generalization. That was the purpose of this PR.

Based on this discussion, I think we have two options. Either make a PR to Flux.jl itself that makes all the optimizers use explicit state and leave Optimisers.jl alone. Or merge some version of this PR into Optimisers.jl with the work around mutability for the future.

DhairyaLGandhi:
Also I just pushed some code I had implemented some time back for the remaining optimisers in `dg/adams` which we can merge, it's tested and everything.

DhairyaLGandhi:
> The explicit state is that generalization. That was the purpose of this PR.

Agreed, but it's incomplete in and of itself. As an example, making some information around training available to the scheduler would be very helpful. 

> Either make a PR to Flux.jl itself that makes all the optimizers use explicit state

Maybe I'm not following this properly, apologies if I am, but I'm pretty sure that optimisers.jl *is* that

src/rules.jl
DhairyaLGandhi: So the way I see it, that much is self documenting because users would be expected to give a tuple-like object (ie immutable) to construct this, and therefore the order matters. The difference is that it acts as multiple optimisers each with their own scheduling/ decays etc. `Sequence` doesn't communicate that. My intuition would fail to recognise what this is a sequence of, is it generic?, could I give it things other than optimisers?, what information does the sequence pass from one element to the next, is this the same as chain?

darsnack:
> As an example, making some information around training available to the scheduler would be very helpful.

What information would that be? Scheduling policies require the current iteration which is easily specified through the explicit state.

> Maybe I'm not following this properly, apologies if I am, but I'm pretty sure that optimisers.jl is that

That's what I'm trying to say!

As a requirement for doing scheduling right, it became clear that the optimizers needed to move to explicit state. So, I have opened the [discussion topic](https://github.com/FluxML/ML-Coordination-Tracker/discussions/22), then I prepared this PR in light of that discussion. This PR follows the design principles of Optimisers.jl to use explicit state. What isn't required (right now) is immutability. So, I am suggesting we save the immutability design goal for later so that we can have explicit state optimizers in Flux now.

src/rules.jl
darsnack: Sure, I see your point. Then what about `SequenceOptimiser`? "Compose" can mean sequential in the sense of function composition, but it is a vague term. I could also want to "compose" optimizers by applying each in parallel to the gradient then averaging their updates. Just cause I specified the arguments to the composition as a tuple doesn't mean that their ordering is automatically significant. `SequenceOptimizer` communicates that the ordering *is* significant.

DhairyaLGandhi:
> What information would that be? 

The Epoch of training for example.

Immutability is part of the design and interface, and I'd keep that. I would much prefer an issue to follow up rather than block the current work.

darsnack:
> The Epoch of training for example.

The epoch of training wouldn't necessarily be information in `Flux.train!`. In any case, ParameterSchedulers.jl already has usage patterns in place when you want to schedule by epoch, iteration, or some mixture. We can save the discussion on scheduling for later maybe.

> Immutability is part of the design and interface, and I'd keep that. I would much prefer an issue to follow up rather than block the current work.

That's why I suggested that if we can't budge on immutability in Optimisers.jl, then I can submit a PR to Flux.jl that implements explicit state but not immutability.

src/rules.jl
DhairyaLGandhi: If sequence doesn't add enough and compose is ambiguous, then we are left with `Optimiser`.

> Just cause I specified the arguments to the composition as a tuple doesn't mean that their ordering is automatically significant.

I mean, putting significance to argument ordering is what APIs exist to do, so I don't think that's a great rebuttal. `map`, `tuple`, `string`, even `Chain` etc are examples

DhairyaLGandhi:
I'm under the assumption that you feel strongly about keeping mutation since it's the default in Flux?

darsnack:
No, I don't care whether it is immutable or mutable *right now*. I care about whether it will be usable on Flux#master in a reasonable time frame. In fact, I completely agree with the design goal of immutability, and I would very much want it to be how future Flux optimizers are designed. All I'm suggesting is that we progressively add the  features of explicit state first then immutability later.

The reason for adding these features progressively is that explicit state *soon* will give us a much nicer optimizers interface on which to build important tools like optimizer composition and scheduling. Immutability is not required to build these tools, but I 100% support moving towards the goal of immutability on a separate time scale from explicit state.

DhairyaLGandhi:
Since this package was designed as the next step we see the optimisers taking see https://github.com/FluxML/Flux.jl/issues/637 for reference, this is the first step. Mutation as an implementation would build up on top of this. I'm happy to do the correct thing first and go from there.

src/rules.jl
darsnack: `map`, `tuple`, and `string` are all standard programming constructs that enforce ordering. The fact that we don't need to be more specific about ordering is because the ordering enforcement is well understood.

`Chain` is actually a good example. The word "chain" implies ordering because the links in a physical object chain are ordered one after then other. `Parallel` is another layer that composes several sub-layers, except that unlike `Chain`, it evaluates the sub-layers in parallel. If we have meaningless names to these two layers, like `Optimiser`, then `Chain` would be called `Layer` and `Parallel` would also want to be called `Layer`, and none of it would make any sense.

The only reason I didn't name this type of optimizer composition as `Chain` is because we already have something called `Chain`. The names `Sequence`, `SequenceOptimiser`, `ChainOptimiser`, etc. are all okay by me. `Optimiser` is not because it doesn't mean anything other than this is a collection of sub-optimizers. It says nothing about how those sub-optimizers will be used or evaluated.

DhairyaLGandhi:
I'm confused then, how are the immutable updates relevant to the schedulers?

darsnack:
> I'm confused then, how are the immutable updates relevant to the schedulers?

They are not. That's my point. We can have explicit state and build nice tools that our users want (like scheduling) on top of explicit state alone.

> As I mentioned, there's a branch already with both features that we can merge right now which should take care of the timescale. Would that be more acceptable?

Yes, that would be great. I'm happy to do that then. Ping me on that branch's PR, and let's make the final push to get it merged!

ToucheSir:
Just a couple questions to really pin down this mutable vs immutable thread:

1. What is the memory overhead of a pure `apply` vs mutating `apply!`. e.g. given `2*|model params|` (model params + gradients) `apply(o::Descent, ...)` allocates `|model params|` and `apply!(o::Descent, ...)` allocates nothing. This number will obviously be different depending on whether an optimizer uses temporary arrays as well, but I think those can be mostly ignored since the relative magnitude of overhead is what we're primarily concerned about.
2. If there is an overhead, what is the acceptable limit? At what point does it become a performance or usability regression because models that used to train are now OOMing or thrashing VRAM?
3. As much as I'm a fan of the pure non-mutating approach, how many Flux users will actually require it any time soon? AFAICT the new XLA.jl is the only one, and that hasn't received any updates in over a month on the still very WIP libtpu branch.

I think these should be sufficient to reach a consensus about what approach/tradeoffs to take, but LMK if there's anything else you think should be added to the list above.

DhairyaLGandhi:
1. About 10%
2. haven't noticed ooms, I think memory wise it has the same performance for standard models (you should be able to train with the same batchsize

darsnack:
If (1) is about 10%, then is that because you only allocate when patching a leaf node then that memory can be reclaimed?

DhairyaLGandhi:
I think it's a good idea to test them, the final version might be with functors, and an out of place update step, but the ci shows them converging as expected

darsnack:
Okay based on this discussion, @DhairyaLGandhi it sounds like you are fairly confident that the memory consumption of immutable updates is manageable without additional steps.

So, I have removed all the mutability changes to interface.jl from this PR. It should match #10 exactly. I would still like to address some of the other comments on this PR. I'll summarize them:
1. `Descent` converts `eta` to the `eltype(dx)`. Instead of conversion everywhere, since all the optimizers are type parameterized, can we make the default constants all `Float32`? This would match the pattern in most of Flux where everything defaults to `Float32`.
2. Can the `epsilon` constant used to prevent divide-by-zero errors be defaulted to `eps(T)` where `T` is the `eltype(dx)`? AFAICT, `1f-8` is an artifact from other languages to where people just hardcode `eps(T)`.
3. I renamed `Sequence` to `SequenceOptimiser`. Is there another (better) name that we want?

DhairyaLGandhi:
To be clear, I am not against mutation, far from it, i don't appreciate any performance drop, but this had more to do with development cycles. We would need to support mutation anyway, it's easier to do it like this.

darsnack:
Copying over @DhairyaLGandhi comment so we can keep the discussion in one place. In response to how to use array traits to apply mutable updates:
> we would travel the tree with functors to choose the update function based on the type of parameters. Array interface would tell us which version to choose. We can optionally have mutating updates that optimisers may want to implement, but that's ideally not going to be the case

This basically my means we'll need to write `apply!` right?

darsnack:
Though I guess we can save the discussion on enabling mutating updates for another issue/PR. If we can just have consensus on [these questions](https://github.com/FluxML/Optimisers.jl/pull/9#issuecomment-769933834) above, then I think we can merge this PR.

src/rules.jl
DhairyaLGandhi: Why not `espilon::T` then?

src/interface.jl
```diff
@@ -1,13 +1,11 @@
-function patch(x, x̄)
-  return x .- x̄
-end
+patch(x, x̄) = x .- x̄
 
-function state(o, x)
+function init(o, x)
```
DhairyaLGandhi, line 3: Can we leave this name change here

src/rules.jl
```diff
@@ -1,106 +1,521 @@
...
+end
+SequenceOptimiser(opts...) = SequenceOptimiser(opts)
+
+init(o::SequenceOptimiser, x::AbstractArray) = [init(opt, x) for opt in o.opts]
+
+(o::SequenceOptimiser)(m, dm, state) = update(o, m, dm, state)
+
+function apply(o::SequenceOptimiser, x, dx, states)
+  for (i, (opt, state)) in enumerate(zip(o.opts, states))
+    dx, states[i] = apply(opt, x, dx, state)
```
DhairyaLGandhi, line 517: We should construct the states array rather than set into an array

DhairyaLGandhi (commented):
All the structs can be made immutable as well. ref https://github.com/FluxML/Optimisers.jl/pull/3#discussion_r441474631

darsnack:
If the structs are immutable, then how do you do scheduling? You'd need to separate the LR from the rest of the optimizer.

ToucheSir:
Re-init the optimizer struct? If we standardize on LR as the first param (or as a kwarg), then it shouldn't cause any major overhead since immutable structs can be allocated on the stack.

darsnack:
You could return a new optimizer with a new LR, but I don't see how that would work nicely with a loop. Seems like a messy design to me.

darsnack:
What I mean by this is that only the scope that created the initial reference can update it, right? You could initialize a new optimizer, but you couldn't make the original reference point to it.

ToucheSir:
I would think it plays nicer because it removes the need for `update_func` in ParameterSchedulers? Instead you could simply pass the constructor and have the scheduler inject hparams -> return the initialized optimizer at every loop iteration.

> You could initialize a new optimizer, but you couldn't make the original reference point to it.

Why not create a new `Descent` inside the loop? I can't think of many use-cases where one needs to hold on to an optimizer reference after the inner loop. The only reason you have to do so in PyTorch and current Flux is because the optimizers are coupled to/carry around their parameters.

darsnack:
Oh yeah I like that a lot. Sounds good! Will make the all the requested changes later today.

src/rules.jl
```diff
@@ -1,106 +1,521 @@
...
+"""
+mutable struct Momentum{T,S}
+  eta::T
+  rho::S
 end
+Momentum(; η = 0.01, ρ = 0.9) = Momentum(η, ρ)
 
-function (o::Descent)(m, m̄, st)
-  update(o, m, m̄, st)
+init(o::Momentum, x::AbstractArray) = (velocity = zero(x),)
```
DhairyaLGandhi, line 46: Can we also get rid of the namedtuples and use regular tuples. This is a clever trick but we don't want to impose that when creating the base design. With decoupling the state, we can message to users that they can use their own custom structs in there if they so wish, but let's keep it simple as we develop the design.

darsnack: The choice of what structure is used for state is self-contained to each optimiser. For example,`Descent` doesn't need to use a named tuple just cause `ADAM` does. No one is forced to follow it, but people will probably look to the default optimizers as examples.

I would argue in favor of the named tuples here. Named tuples are nice for state like this, because it prevents ordering mismatch bugs. And it is cleaner description what the state of the optimizer is.

src/rules.jl
darsnack: That's the change I want to make, yes.

src/rules.jl
darsnack: Just to clarify, you basically want to create a new array. Or is it a specific problem with `setindex!`? Would this be what you're looking for?
```julia
function apply(o::SequenceOptimiser, x, dx, states)
  new_states = similar(states)
  for (i, (opt, state)) in enumerate(zip(o.opts, states))
    dx, new_states[i] = apply(opt, x, dx, state)
  end

  return dx, new_states
end
```

darsnack:
@CarloLucibello any more thoughts on the keyword arguments? I am also thinking `Momentum(η = 0.01; ρ = 0.9)` would be an option.

CarloLucibello:
For the long term i don't have a clear vision, but i would go with whatever we currently have right now for the time being, any change would cause widespread breakage so we have to think it throughly and separetly from this pr. Sorry I am from the cellphone and can't do proper assessment right now

DhairyaLGandhi:
Can the constructors represent the state better, for ex, forcing unicode to make a param explicit isn't ideal. I would prefer positional args here.

src/rules.jl
DhairyaLGandhi: That may be well, but the package itself is relatively early in its development and I would not impose a pattern of use until we are clear how the design would mature. We might have need for a separate init struct,for example, but that isn't clear yet, so want to reduce the verbiage for the initial set of rules we create here.

src/rules.jl
DhairyaLGandhi: Something like this, but without the call to similar


I very rough analogy would be push!, But something that would accumulate into a new array rather than set into it.

src/rules.jl
darsnack: Okay sure

src/rules.jl
darsnack: I'm not sure I follow. Why is `push!` only an analogy?

src/rules.jl
```diff
@@ -1,106 +1,516 @@
...
+
+Decay weights by `γ`.
+
+# Parameters
+- Weight decay (`γ`): Decay applied to weights during optimisation.
+"""
+struct WeightDecay{T}
+  wd::T
+end
+WeightDecay() = WeightDecay(1e-4)
```
DhairyaLGandhi, line 482: Give default value?

darsnack: Sorry what do you mean?

DhairyaLGandhi: Sorry I mistyped, I wanted to mention that type it as f32?

darsnack: On that front, can we type all the default parameters as f32?

DhairyaLGandhi: I think so. Just as default, but users could do whatever suits them.

src/rules.jl
CarloLucibello: Maybe `OptChain`? Anyways, I don't think we should do this name change here, it's another part of the API we don't wanna break right now

CarloLucibello:
This looks fine, except for the fact the composite optimizer should be renamed. 
I'm worried about the breakage though. Everyone's code with custom training loops (i.e. everything using `update!`, like all my flux code) is going to be broken. Of minor importance, custom optimizers overloading `apply!` will fail. 
Do we have a deprecation path?
Maybe we should temporarily add a `state` field to each optimizer here so that we can repurpose `update!` in Flux while deprecating it at the same time.

DhairyaLGandhi:
All good concerns, thanks!

I think the naming is not the end of the world, but I would definitely prefer a better name. I dont think we want to add a state field though, this is a different design and we shouldn't be too worried about the breakages that can be managed elsewhere (see the corresponding PR to Flux where I have updated the commented out `update!`)

We should definitely come up with a smoother breakage path though, although I'm not sure we can get around everything.

darsnack:
AFAICT, even with the commented `update!` in https://github.com/FluxML/Flux.jl/pull/1481, there is still a breaking change since the explicit state must be passed into `update!`.

I don't think there is really any clean way around it. The only way to avoid passing around the state would be to store it in some structure. The only structures available to `update!` are the optimizer, `Params`, and `Grads`. `Params`/`Grads` are clearly poor candidates leaving only the optimizer like @CarloLucibello suggested.

We could do something like
```julia
struct OptAndState{O, S}
  opt::O
  state::S
end
```
Then use `@forward` to push things onto the optimizer. All this code would be in Flux.jl not Optimisers.jl, so we can save the discussion for https://github.com/FluxML/Flux.jl/pull/1481. Can we agree that the soft-breakage code doesn't belong in this repo?

src/rules.jl
ToucheSir: We talked about this on Tuesday's call, but I don't recall what was agreed on. WRT breaking, I think it's fine to use a name other than `Optimiser` here, since we can always add an alias `Optimiser = ...` or something in Flux itself.

ToucheSir:
Will `update!` and `Optimiser` get deprecation warnings as part of this bringup?

CarloLucibello:
@darsnack this is a basic script that I'd like to see still working after the change

```julia
using Flux
using Flux.Optimise

opt = ADAM(0.1)
x = rand(2)

for t=1:10
    g = gradient(x -> sum(x.^2), x)
    update!(opt, x, g[1])
    opt.eta /= 2
end
```
If we keep this PR as it is, in order to provide a deprecation path in Flux we would have to create a parallel family of optimizers, ADAM, Nesterov, etc... where

```julia
struct ADAM
   opt::Optimisers.ADAM
   state   
end
```
then `@forward` a few methods and also overload `setfield` and `getfield`. This could be done, but wouldn't be more practical to just add a `state` to the optimizers here?

DhairyaLGandhi:
In that case I'd rather just have a breaking change.

darsnack:
Okay renamed to `ChainOptimiser` from `SequenceOptimiser` as was agreed on the last ML community call.

I disagree with a breaking change. This is such a fundamental piece of the framework, that we shouldn't play fast and loose with people's code.

@CarloLucibello I think we could have deprecation path code in this repo in a `flux.jl` file, but I still feel that the code belongs in Flux.jl itself. Materially, I don't think it matters too much, since Flux.jl will probably be the only downstream user of Optimiser.jl. But I don't think any users of Optimisers.jl should be allowed to use implicit state. It was never a part of the package, so I think we shouldn't introduce it only to deprecate it. The only users who can expect implicit state are Flux.jl users, so that's where I think the deprecation path belongs.

If there is a concern about code complexity, then I think something like the following should resolve it:
```julia
for $Opt in (:Descent, :ADAM) # we would have all in the real code
  @eval begin
    mutable struct $(Opt){T}
      opt::Optimisers.$Opt
      state::T
    end

    Base.getproperty(o::$Opt, p) = getproperty(o.opt, p)
    Base.setproperty!(o::$Opt, p, v) = setproperty!(o.opt, p, v)

    function apply!(o::$Opt, x, dx)
      dx, s = apply(o.opt, x, dx, o.state)
      o.state = s
      
      return dx
    end
  end
end

**Comment by CarloLucibello** [comment]
ok, we'll cook something like that in flux, handling `init` as well.

Two more things (feel free to ignore both)
- the chain in `ChainOptimizer` feels like a verb, maybe better OptimizersChain or OptChain?
- can anyone do or point me to a very short summarization of the advantages of the external state approach compared to the IdDict one? One disadvantage I see is that having to init the state and pass it around is slightly annoying

**Comment by DhairyaLGandhi** [comment]
I don't think we would play fast and lose but instead deprecate the current ones as a whole, pointing to the new explicit state approach, and handle both of them inside Flux for a technically breaking change. That way we can preserve custom loops for one more release cycle.

**Comment by darsnack** [comment]
What about `ChainedOptimiser`? Happy to do `OptimiserChain` too.

There are probably more reasons than I list here for explicit state, so @ToucheSir and @DhairyaLGandhi can chime in:

With `IdDict`s (or any form of implicit state), the state is an implementation detail instead of part of the interface. This is an issue for people trying to extend the interface, since using state is a requirement. For example, if you are writing a scheduler, you might do:
```julia
struct Scheduler{S, O}
  opt::O
  schedule::S
  iter::Int
end

function apply!(o, x, dx)
  o.opt.eta = o.schedule[o.iter]
  o.iter += 1
  return apply!(o.opt, x, dx)
end
```
This seems like a perfectly reasonable and logical thing to write, but it won't work unless `o` is optimizing exactly one array `x`. This is because the state which is associated with each `x` that `o` is applied to is not accounted for anywhere. This can be rectified by replacing `iter` with an `IdDict` and doing `o.iter[x]`, but it wouldn't be obvious to anyone extending the interface at first glance.

So, state is a requirement for almost every optimizer. Aside for a complicated paragraph in the docs, the best way to make this clear to users is by making state part of the interface if it is a requirement and not an implementation detail. It also allows anyone writing an optimizer to write the natural code (w.r.t. a single `x`), then we manage passing around the state for complicated data structures.

I think largely, the benefits are that this is a clearer, more extensible design. It should make for safer code with easier to detect errors. There's also the XLA-side of this where explicit state and immutability can allow compiler optimizations, but those are further down the line.

darsnack:
It also means that the code for an optimizer is stable. When we want to special case how we handle the multiple states associated with a data structure, it only needs to be done once, and none of the optimizer rules need to change their code.

darsnack:
I think everything should be resolved in the latest commit. I was convinced that `OptimiserChain` was the best name.

Docstrings are in place, but we should make setting up the docs a separate PR.

CarloLucibello (approved):

CarloLucibello:
@DhairyaLGandhi can I have write access to this repo (actually, if it could be made org-wide it would be much more convenient)

darsnack:
Bump can we get the write-access for this repo to match the org? I think this can be merged.

DhairyaLGandhi (approved):

DhairyaLGandhi:
Thanks!

PR #10, opened 2021-01-29
Add more optimisers

DhairyaLGandhi:
- [x] Add to tests
- [x] copy over docs

cc @darsnack

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -109,7 +109,7 @@ struct RADAM{T,S}
   beta::S
 end
 
-init(o::RADAM, x::AbstractArray) = (zero(x), zero(x), [β[1], β[2]], 1)
+init(o::RADAM, x::AbstractArray) = (zero(x), zero(x), [o.beta[1], o.beta[2]], 1)
 
 function apply(o::RADAM, x, Δ, st)
   η, β = o.eta, o.beta
@@ -137,7 +137,7 @@ struct AdaMax{T,S}
   beta::S
 end
 
-init(o::AdaMax, x::AbstractArray) = (zero(x), zero(x), [β[1], β[2]])
+init(o::AdaMax, x::AbstractArray) = (zero(x), zero(x), [o.beta[1], o.beta[2]])
 
 function apply(o::AdaMax, x, Δ, st)
   η, β = o.eta, o.beta
@@ -157,7 +157,7 @@ struct OADAM{T,S}
   beta::S
 end
 
-init(o::OADAM, x::AbstractArray) = (zero(x), zero(x), zero(x), [β[1], β[2]])
+init(o::OADAM, x::AbstractArray) = (zero(x), zero(x), zero(x), [o.beta[1], o.beta[2]])
 
 function apply(o::OADAM, x, Δ, st)
   η, β = o.eta, o.beta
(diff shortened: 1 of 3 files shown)


darsnack:
Thanks! This adds the optimizers like #9 but I don't see what's there to avoid making copies of all the parameters.

DhairyaLGandhi:
This is the immutable version with explicit state which was what we wanted; the inplace updates are essentially implementing https://github.com/SciML/ArrayInterface.jl#ismutablex, if that's alright with you.

That change is still separate, and should be in a separate PR.

darsnack:
Okay so just to help me understand things...every optimizer implements `apply` which is immutable. `ismutablex` let's us know if an array is mutable. How are you planning to put those pieces together to update the gradient without making a copy?

DhairyaLGandhi:
Basically we would travel the tree with functors to choose the update function based on the type of parameters. Array interface would tell us which version to choose. We can optionally have mutating updates that optimisers may want to implement, but that's ideally not going to be the case

DhairyaLGandhi:
So this is done, we can merge this now if its alright with you

darsnack:
Generally yes it's fine by me, but can we instead address and merge #9 since it has unanswered comments on the same code?

darsnack:
I was initially thinking that this PR was going to be some solution to the memory overhead problem, but it is just an immutable version of #9. So it isn't needed since #9 is also immutable now, and if it's all the same, maybe we can just use #9 since people have commented on that code already.

DhairyaLGandhi:
Any specific concerns? I don't see any that would correspond to this PR, since I would leave out the changes to interface.jl from that pr.

darsnack:
I mean I don't really see the point of merging this then rebasing #9 but if that's what you want then go ahead.

DhairyaLGandhi:
Oh no, of we merge #9 then I'll close this

src/rules.jl
```diff
@@ -1,8 +1,27 @@
...
+
+opt = Descent(0.3)
+
+ps = params(model)
+
+gs = gradient(ps) do
+    loss(x, y)
+end
+
+Flux.Optimise.update!(opt, ps, gs)
```
darsnack, line 23: I'm not sure Flux/Zygote code belongs in the doc string for this package even though they are the main consumer.

More appropriate in the docs as tutorials instead of a do string.

src/rules.jl
```diff
@@ -104,3 +198,393 @@ function apply(o::ADAM, x, Δ, st)
...
+function apply(o::ExpDecay, x, Δ, st)
+  η, s, decay = o.eta, o.step, o.decay
+  n, = st .+ 1
+  if n%s == 0 && count(x -> x%s == 0, st) == 1
+    η = max(η * decay, o.clip)
+    o.eta = η
+  end
+  Δ = Δ * η
+  Δ, (n,)
+end
```
darsnack, line 570: These aren't really optimizers. They are scheduling policies implemented as optimizers, so I would prefer to leave them out for now until we have a discussion about scheduling.

CarloLucibello:
this should be closed. @DhairyaLGandhi please grant us the write rights we already requested multiple times

PR #16, opened 2021-02-24
RFC: Switch interface for higher-order optimizers

darsnack:
In response to [this discussion](https://github.com/FluxML/ML-Coordination-Tracker/discussions/22#discussioncomment-342967), this PR makes a simple change to the order of the arguments in our interface. Functionally, no changes are made.

Higher-order optimizers (e.g. AdaHessian) consume the first and second derivatives as arguments to the rule. This change allows us to support arbitrary number of derivative terms in the future (though it is limited to only the gradient for now).

Summary of changes:
- `apply(o, x, dx, st)` -> `apply(o, st, x, dx)`
- `update(o, x, dx, st)` -> `update(o, st, x, dx)`
- `o(x, dx, st)` -> `o(st, x, dx)`

The idea is that a future second-order rule would define `apply(o, st, x, grad, hessian) = ...`. And the `update` would be changed to support `Vararg`:
```julia
function update(o, state, x::T, x̄s...) where T
  if all(isnothing, x̄s)
    return x, state
  elseif isleaf(x)
    return _update(o, state, x, x̄s...)
  else
    # don't know if this is correct, but it's just a sketch
    x̄s = foreach(x̄ -> functor(typeof(x), x̄)[1], x̄s)
    x, restructure = functor(typeof(x), x)
    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
    return restructure(map(first, xstate)), map(x -> x[2], xstate)
  end
end
```

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -9,20 +9,23 @@ function state(o, x)
   end
 end
 
-function _update(o, st, x, x̄)
-  x̄, st = apply(o, st, x, x̄)
-  return patch(x, x̄), st
+function _update(o, st, x, x̄s...)
+  st, x̄ = apply(o, st, x, x̄s...)
+  return st, patch(x, x̄)
 end
 
-function update(o, state, x::T, x̄) where T
-  if x̄ === nothing
-    return x, state
+function update(o, state, x::T, x̄s...) where T
+  if all(isnothing, x̄s)
+    return state, x
   elseif isleaf(x)
-    return _update(o, state, x, x̄)
+    return _update(o, state, x, x̄s...)
   else
-    x̄, _  = functor(typeof(x), x̄)
+    x̄s = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
     x, restructure = functor(typeof(x), x)
-    xstate = map((x, x̄, state) -> update(o, state, x, x̄), x, x̄, state)
-    return restructure(map(first, xstate)), map(x -> x[2], xstate)
+    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
+    return map(first, xstate), restructure(map(last, xstate))
   end
 end
+
+# default all rules to first order calls
+apply(o, state, x, dx, dxs...) = apply(o, state, x, dx)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -9,20 +9,20 @@ function state(o, x)
   end
 end
 
-function _update(o, x, x̄, st)
-  x̄, st = apply(o, x, x̄, st)
+function _update(o, st, x, x̄)
+  x̄, st = apply(o, st, x, x̄)
   return patch(x, x̄), st
 end
 
-function update(o, x::T, x̄, state) where T
+function update(o, state, x::T, x̄) where T
   if x̄ === nothing
     return x, state
   elseif isleaf(x)
-    return _update(o, x, x̄, state)
+    return _update(o, state, x, x̄)
   else
     x̄, _  = functor(typeof(x), x̄)
     x, restructure = functor(typeof(x), x)
-    xstate = map((x, x̄, state) -> update(o, x, x̄, state), x, x̄, state)
+    xstate = map((x, x̄, state) -> update(o, state, x, x̄), x, x̄, state)
     return restructure(map(first, xstate)), map(x -> x[2], xstate)
   end
 end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -15,13 +15,13 @@ Descent() = Descent(1f-1)
 
 init(o::Descent, x::AbstractArray) = nothing
 
-function apply(o::Descent, x, dx, state)
+function apply(o::Descent, state, x, dx)
   η = convert(eltype(dx), o.eta)
   
   return dx .* η, state
 end
 
-(o::Descent)(m, dm, st) = update(o, m, dm, st)
+(o::Descent)(state, m, dm) = update(o, state, m, dm)
 
 """
     Momentum(η = 1f-2, ρ = 9f-1)
@@ -42,14 +42,14 @@ Momentum(η = 1f-2, ρ = 9f-1) = Momentum{typeof(η)}(η, ρ)
 
 init(o::Momentum, x::AbstractArray) = zero(x)
 
-function apply(o::Momentum, x, dx, state)
+function apply(o::Momentum, state, x, dx)
   η, ρ, v = o.eta, o.rho, state
   @. v = ρ * v - η * dx
   
   return -v, v
 end
 
-(o::Momentum)(m, dm, state) = update(o, m, dm, state)
+(o::Momentum)(state, m, dm) = update(o, state, m, dm)
 
 """
     Nesterov(η = 1f-3, ρ = 9f-1)
@@ -70,9 +70,9 @@ Nesterov(η = 1f-3, ρ = 9f-1) = Nesterov{typeof(η)}(η, ρ)
 
 init(o::Nesterov, x::AbstractArray) = zero(x)
 
-(o::Nesterov)(m, dm, state) = update(o, m, dm, state)
+(o::Nesterov)(state, m, dm) = update(o, state, m, dm)
 
-function apply(o::Nesterov, x, dx, state)
+function apply(o::Nesterov, state, x, dx)
   η, ρ, v = o.eta, o.rho, state
   d = @. ρ^2 * v - (1+ρ) * η * dx
   @. v = ρ * v - η * dx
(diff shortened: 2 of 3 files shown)


darsnack:
Actually, the following passes the tests:
```julia
function _update(o, st, x, x̄s...)
  x̄, st = apply(o, st, x, x̄s...)
  return patch(x, x̄), st
end

function update(o, state, x::T, x̄s...) where T
  if all(isnothing, x̄s)
    return x, state
  elseif isleaf(x)
    return _update(o, state, x, x̄s...)
  else
    x̄s = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
    x, restructure = functor(typeof(x), x)
    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
    return restructure(map(first, xstate)), map(x -> x[2], xstate)
  end
end
```

So we could add support for higher-order now even though there are no higher-order rules defined.

darsnack:
One other concern is composition. Here's a sketch for `OptimiserChain`:
```julia
function apply(o::OptimiserChain, states, x, dx, dxs...)
  new_states = similar(states)
  for (i, (opt, state)) in enumerate(zip(o.opts, states))
    dx, new_states[i] = apply(opt, state, x, dx, dxs...)
  end

  return dx, new_states
end
```

As you can see, this forces the use of the first `dx` which represented the update or "patch" to `x`. The higher-order terms don't get transformed. This would exclude zero-th order optimizers from composition, but I don't have a clear picture of what composition even means in that case.

It also assumes that all the composed optimizers are of the same order. This is something I would like to fix. You could do it like `apply(o::Descent, state, x, xs...) = apply(o, state, x, first(xs))`, but I have yet to figure out how to make that generic.

ToucheSir:
One crude approach would be to have first-order optimizers take in a vararg `dxs...`, slice out `dx = dxs[1]` and concat that onto the rest when returning. In other words, each optimizer assumes `length(dxs) >= [optimizer order]` and doesn't touch anything higher. There are probably weird interactions that would come out of gradient transforms like clipping for this, but it's the best I can think of at the moment.

darsnack:
Yeah I wanted to enable something that allowed writing rules with a fixed number of `dxs` based on order (syntactically the clearest for the rule author). The other thought was that while a rule can accept an arbitrary number of derivative terms, it produces a single transformed `Δx` that gets `patch`ed onto `x`.

If there was an interface for the optimizer to specify its order, then we could handle calling `apply` with the right number of arguments. For example, if we have `order(o::Descent) = 1` and `order(o::AdaHessian) = 2`:
```julia
function apply(o::OptimiserChain, states, x, dx, dxs...)
  new_states = similar(states)
  for (i, (opt, state)) in enumerate(zip(o.opts, states))
    nterms = order(opt)
    dx, new_states[i] = apply(opt, state, x, dx, dxs[1:nterms])
  end

  return dx, new_states
end
```
If we had an abstract type, you could encode the order into the type.

ToucheSir:
> Yeah I wanted to enable something that allowed writing rules with a fixed number of dxs based on order (syntactically the clearest for the rule author)

I was also thinking about how he might design the interface for this. Some design ideas below:
```julia
# 0-order
apply(o, states, x) = states # or x, ()

# >= 2-order
function apply(o, states, x, dx, dxs...)
  states, dx = apply(o, states, x, dx)
  return states, (dx, dxs...)
end
```
This removes the boilerplate for 1st-order optimizers. Given a way to query optimizer order, this could be generalized to any order.

src/rules.jl
```diff
@@ -15,14 +15,14 @@ Descent() = Descent(1f-1)
 
 init(o::Descent, x::AbstractArray) = nothing
 
-function apply(o::Descent, x, dx, state)
+function apply(o::Descent, state, x, dx)
   η = convert(eltype(dx), o.eta)
   dx .*= η
   
   return dx .* η, state
```
ToucheSir, line 22: Should the return order be reversed as well to stay consistent? Otherwise I think we'd need to enforce the first item being a tuple as well since `OptimiserChain` does `dx, state = apply(o, ...)`

darsnack: Yeah that makes sense

src/rules.jl
```diff
@@ -504,15 +504,15 @@ OptimiserChain(opts...) = OptimiserChain(opts)
 
 init(o::OptimiserChain, x::AbstractArray) = [init(opt, x) for opt in o.opts]
 
-(o::OptimiserChain)(m, dm, state) = update(o, m, dm, state)
+(o::OptimiserChain)(state, m, dm) = update(o, state, m, dm)
 
-function apply(o::OptimiserChain, x, dx, states)
+function apply(o::OptimiserChain, states, x, dx)
```
ToucheSir, line 509: Should this and the `dx` on L512 be varargs/splats? Do we need to come to a consensus about how to handle order != 1 optimizers first?

darsnack: That was an omission by typo on my part. My overall thought process is this:
- `OptimiserChain` works for order >= 1 optimizers (since I don't know how you can compose zero order optimizers if they don't accept a `Δx`...we could make the distinction between `Δx` and `dx`, but I'm not inclined to do that)
- Every optimizer can consume an arbitrary number of derivative terms, but it produces a single "patch" `Δx`
- `OptimiserChain` handles the multiple states and passing the transformed "patch" between rules, calling each one as if it vararg number of terms
- By default, if a higher order rule doesn't exist, the dispatch will route to a first order rule

darsnack:
Okay, so I am pretty happy with where we are heading. I think we've managed to support arbitrary order optimizers with no extra boilerplate, and a rule write can just add more terms onto `apply` if they need them. Here's a crude (i.e. not good) demonstration of `AdaHessian` (I ignore the hessian estimation part of the algorithm...just assume that `D` is correct).
```julia
using Optimisers
using Statistics: mean

function spatial_smoothing(D, b)
    Dout = similar(D)
    d = length(D)
    for i in 0:(d ÷ b - 1)
        for j in 1:b
            Dout[i * b + j] = sum(D[i * b + k] for k in 1:b) / b
        end
    end

    return Dout
end

struct AdaHessian{T, S<:Integer}
    eta::T
    beta::Tuple{T, T}
    power::T
    block::S
    epsilon::T
end
AdaHessian(η = 1.5f-1, β = (9f-1, 9.99f-1), k = 0.5, b = 1, ϵ = eps(typeof(η))) =
    AdaHessian{typeof(η), typeof(b)}(η, β, k, b, ϵ)

Optimisers.init(o::AdaHessian, x::AbstractArray) = (zero(x), zero(x), o.beta)

function Optimisers.apply(o::AdaHessian, state, x, dx, D)
    η, β, k, b, ϵ = o.eta, o.beta, o.power, o.block, o.epsilon
    mt, vt, βt = state

    Ds = spatial_smoothing(D, b)
    @. mt = β[1] * mt + (1 - β[1]) * dx
    @. vt = β[2] * vt + (1 - β[2]) * Ds^2
    dx = @. mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)^k * η

    return (mt, vt, βt .* β), dx
end

w, b = rand(3, 4), rand(3)

o = AdaHessian()
ps = (w, b)
gs = map(x -> rand(size(x)...), ps)
Ds = map(x -> rand(size(x)...), ps)
st = Optimisers.state(o, ps)
st, ps = Optimisers.update(o, st, ps, gs, Ds)
```

It all runs, and I made no changes to the existing rules other than switching the order of the arguments/returns for `apply` and `update`. Pretty nice!

ToucheSir:
Bump. I think this should be good to go after a rebase?

darsnack:
Thanks for the reminder. I'll probably only get time to work on this toward the end of this week. I think it needs a rebase, and I want to make sure for all the combinations (1st order + higher order / with + without a rule defined) that we hit the right defaults.

darsnack:
Maybe @CarloLucibello or @DhairyaLGandhi want to take a look over this, but I think it is good to go. This is a simple change that make no functional difference for the current optimizers, but it gives us an API that can handle higher-order optimizers in the future.

ToucheSir (approved):
LGTM!

src/interface.jl
```diff
@@ -9,20 +9,23 @@ function state(o, x)
...
-    return _update(o, x, x̄, state)
+    return _update(o, state, x, x̄s...)
   else
-    x̄, _  = functor(typeof(x), x̄)
+    x̄s = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
     x, restructure = functor(typeof(x), x)
-    xstate = map((x, x̄, state) -> update(o, x, x̄, state), x, x̄, state)
-    return restructure(map(first, xstate)), map(x -> x[2], xstate)
+    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
+    return map(first, xstate), restructure(map(last, xstate))
```
CarloLucibello, line 26: why restructure changed position here?

CarloLucibello: ah ok, seems fine

CarloLucibello (approved):

DhairyaLGandhi:
A better interface would be to expect the higher order rules to expect Tuples instead which was already the case. It is more explicit to have the tuple and avoids this major a change for all the other optimisers. I am thinking 

```Julia
update(opt, x, (dx1, dx2, ...), st)
vs.
update(opt, st, x, dx1, dx2, ...)
```

It seems to be more consistent and is in line with thinking of the higher order derivs as partials (and how things like `Tangent` work). Maybe let's do that instead. It would make some stuff for Flux and DDP easier to handle as well.

PR #27, opened 2022-01-27
Add gradient clipping

mcabbott:


src/rules.jl
```diff
@@ -490,6 +490,58 @@ function apply(o::WeightDecay, state, x, dx)
...
+ClipNorm(ω = 10f0, p = 2; throw::Bool = true) = ClipNorm{throw,typeof(ω)}(ω, p)
+
+init(o::ClipNorm, x::AbstractArray) = nothing
+
+(o::ClipNorm)(state::Nothing, m, dm) = update(o, state, m, dm)
+
+function apply(o::ClipNorm{E}, state, x, dx) where {E}
+  nrm = norm(dx, o.p)
+  if E && !isfinite(nrm)
+    throw(DomainError("gradient has $(o.p)-norm $nrm, for array $(summary(x))"))
```
mcabbott, line 538: Unlike the Flux version, this allows p-norm, and (by default) throws an error on infinite or NaN.

test/runtests.jl
```diff
@@ -18,25 +19,50 @@ using Statistics
...
+    for t = 1:10^5
+      x = rand(10)
+      gs = gradient(w -> loss(x, w, w′), w)
+      st, w = Optimisers.update(opt, st, w, gs...)
+    end
+    @test loss(rand(10, 10), w, w′) < 0.01
+  end
+
+  @testset "gradient clipping" begin
+    @test_skip m = (α = ([0], sin), γ = rand(3))
```
mcabbott, line 40: At present a Optimises can't handle non-arrays when setting up the state.

src/rules.jl
```diff
@@ -490,6 +490,58 @@ function apply(o::WeightDecay, state, x, dx)
   return state, dx
 end
 
+"""
+    ClipGrad(δ = 10f0)
+
+Restricts every gradient component to live in `-δ ≤ dx ≤ δ`.
```
darsnack, line 496:
```suggestion
Restricts every gradient component to stay in `-δ ≤ dx ≤ δ`.
```

src/rules.jl
```diff
@@ -490,6 +490,58 @@ function apply(o::WeightDecay, state, x, dx)
...
+
+function apply(o::ClipGrad, state, x, dx)
+  δ = convert(eltype(dx), o.delta)
+  dx′ = @. clamp(dx, -δ, δ)
+
+  return state, dx′
+end
+
+"""
+    ClipNorm(ω = 10f0, p = 2; throw = true)
```
darsnack, line 517: Looks like there's a mismatch between ω and μ.

src/rules.jl
```diff
@@ -490,6 +490,58 @@ function apply(o::WeightDecay, state, x, dx)
...
+
+Scales any gradient array for which `norm(dx, p) > μ` to
+bring this p-norm down to the given threshold.
```
darsnack, line 520:
```suggestion
Scales any gradient array for which `norm(dx, p) > μ`
such that this threshold is no longer violated.
```

src/rules.jl
```diff
@@ -490,6 +490,58 @@ function apply(o::WeightDecay, state, x, dx)
...
+Throws an error if the norm is infinite or `NaN`.
+
+See also [`ClipGrad`](@ref).
+"""
+struct ClipNorm{E,T<:Real}
+  omega::T
+  p::T
+end
+ClipNorm(ω = 10f0, p = 2; throw::Bool = true) = ClipNorm{throw,typeof(ω)}(ω, p)
```
darsnack, line 529: Any advantage to storing `throw` in the type when the struct is immutable? I think `o.throw == true` is more informative than `ClipNorm{true}`.

test/runtests.jl
darsnack: https://github.com/FluxML/Optimisers.jl/issues/28

src/rules.jl
mcabbott: No great reason, that might be cleaner. 

Thoughts on unconditionally throwing? That's certainly simpler.

src/rules.jl
mcabbott: I wrote that and then worried about whether the norm is exactly equal to threshold afterwards... will think a bit. 

I guess this is also not true for 0-norm.

src/rules.jl
darsnack: There might be cases where you use `ClipNorm` to prevent a spurious infinite gradient? It's hard to imagine a case where clipping the gradient for a single step would solve the run away training. Any non-ML cases where this could happen?

mcabbott: I guess there are possible cases where an infinte norm does not mean the result will have NaN, which means it might not be crazy to cary on without throwing:
```
julia> x = [1, 2, 3f38, 3f38];

julia> norm(x)
Inf32

julia> x * (10 / norm(x))
4-element Vector{Float32}:
 0.0
 0.0
 0.0
 0.0
```
No idea if that really happens.

darsnack: I guess having the option and throwing by default is the right call.

src/rules.jl
mcabbott: Now I said obey, since strictly speaking you stay inside an interval not an equation...

src/rules.jl
mcabbott: This should be good to go, if you don't hate the current wording.

mcabbott:
Bump?

src/rules.jl
```diff
@@ -490,6 +490,59 @@ function apply(o::WeightDecay, state, x, dx)
...
+
+  return state, dx′
+end
+
+"""
+    ClipNorm(ω = 10f0, p = 2; throw = true)
+
+Scales any gradient array for which `norm(dx, p) > ω`
+to stay at this threshold (unless `p==0`).
+Throws an error if the norm is infinite or `NaN`.
```
darsnack, line 521:
```suggestion
Throws an error if the norm is infinite or `NaN` when `throw == true`.
```

darsnack (approved):
LGTM!

Issue #28, opened 2022-01-27
Handle non-array leaves in `state(o, x)`


darsnack:
Ref https://github.com/FluxML/Optimisers.jl/pull/27/files#r793219036.

PR #29, opened 2022-01-27
Optimise only at `isnumeric` leaves

mcabbott:
Closes https://github.com/FluxML/Optimisers.jl/issues/28

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,3 +1,13 @@
+
+struct State{T} backing::T end
+State(st::State) = st
+backing(st::State) = getfield(st, :backing)
+function Base.show(io::IO, st::State)
+  print(io, "State(")
+  show(IOContext(io, :compact => true), backing(st))
+  print(io, ")")
+end
+
 patch(x, x̄) = x .- x̄
 
 function state(o, x)
@@ -7,27 +17,29 @@ function state(o, x)
     return nothing
   else
     x, _ = functor(x)
-    return map(x -> state(o, x), x)
+    return State(map(x -> state(o, x), x))
   end
 end
 
-function _update(o, st, x, x̄s...)
+function update(o, st, x, x̄s...)
   st, x̄ = apply(o, st, x, x̄s...)
   return st, patch(x, x̄)
 end
 
-function update(o, state, x::T, x̄s...) where T
+function update(o, state::State, x::T, x̄s...) where T
   if all(isnothing, x̄s)
     return state, x
   elseif isnumeric(x)
-    return _update(o, state, x, x̄s...)
+    return update(o, backing(state), x, x̄s...)
   else
     x̄s = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
     x, restructure = functor(typeof(x), x)
-    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
-    return map(first, xstate), restructure(map(last, xstate))
+    xstate = map((state, x, x̄s...) -> update(o, State(state), x, x̄s...), backing(state), x, x̄s...)
+    return State(map(first, xstate)), restructure(map(last, xstate))
   end
 end
+
+update(::State, args...) = throw(ArgumentError("the optimiser state must be the 2nd argument"))
 
 # default all rules to first order calls
 apply(o, state, x, dx, dxs...) = apply(o, state, x, dx)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -6,14 +6,14 @@ function state(o, x)
   elseif isleaf(x)
     return nothing
   else
-    x, _ = functor(x)
-    return map(x -> state(o, x), x)
+    x′, _ = functor(x)
+    return map(xᵢ -> state(o, xᵢ), x′)
   end
 end
 
 function _update(o, st, x, x̄s...)
-  st, x̄ = apply(o, st, x, x̄s...)
-  return st, patch(x, x̄)
+  st′, x̄′ = apply(o, st, x, x̄s...)
+  return st′, patch(x, x̄′)
 end
 
 function update(o, state, x::T, x̄s...) where T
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -8,15 +8,31 @@ using Statistics
...
+    # Original example
+    w = (α = 5rand(3, 3), β = rand(3, 3))
     st = Optimisers.state(o, w)
     loss(x, y) = mean((x.α .* x.β .- y.α .* y.β) .^ 2)
-    l = loss(w, w′)
+    @test loss(w, w′) > 1
     for i = 1:10^4
       gs = gradient(x -> loss(x, w′), w)
-      st, w = o(st, w, gs...)
+      st, w = Optimisers.update(o, st, w, gs...)
```
mcabbott, line 19: Now tests both notations.

If you write, by mistake,
```
st, w = Optimisers.update(st, w, gs...)
```
it still runs without error, except the loss doesn't decline so fast. That's awful.

mcabbott: One idea would be to always wrap the state tree in some marker struct, so that it's clearly distinct.  https://github.com/FluxML/Optimisers.jl/pull/29/commits/cfde5a7eace2042962dc248b8be805074e0d97cd changes this silent bug into this:
```
julia> st, w = Optimisers.update(st, w, gs...)
ERROR: MethodError: no method matching update(::Optimisers.State{NamedTuple{(:α, :β, :γ), Tuple{Tuple{Vector{Float64}, Vector{Float64}, Tuple{Float32, Float32}}, Optimisers.State{NamedTuple{(:parent,), Tuple{Tuple{Matrix{Float64}, Matrix{Float64}, Tuple{Float32, Float32}}}}}, Optimisers.State{Tuple{Tuple{Vector{Float64}, Vector{Float64}, Tuple{Float32, Float32}}, Nothing}}}}}, ::NamedTuple{(:α, :β), Tuple{Matrix{Float64}, Matrix{Float64}}}, ::NamedTuple{(:α, :β), Tuple{Matrix{Float64}, Matrix{Float64}}})
Closest candidates are:
  update(::Any, ::Any, ::AbstractArray, ::Any...) at ~/.julia/dev/Optimisers/src/interface.jl:24
  update(::Any, ::Optimisers.State, ::T, ::Any...) where T at ~/.julia/dev/Optimisers/src/interface.jl:29
```

ToucheSir: Optax uses named state structs for this, e.g. https://optax.readthedocs.io/en/latest/api.html?highlight=state#optax.ScaleByAdamState

mcabbott: I've reverted what I had here, which made `Optimisers.state` wrap the outermost object (and many others) in some State struct, just as a marker to make the signature `update(::Any, ::Optimisers.State, ::T, ::Any...)`. 

But I now wonder why it doesn't wrap the state (for each trainable array) in some `State(optimiser, state)`. Then you wouldn't have to pass the optimiser around in addition to the state, and would write just:
```julia
st = Optimisers.state(ADAM(), w)
st, w = Optimisers.update(st, w, gs...)
```
Having `update` take 3 parameters not 4 seems less confusing. Are there any downsides to this? It would still not require anyone implementing a new optimiser to inherit any types from this package. But, it can be a separate PR.

ToucheSir: Nothing I can think of right now. It would require the user include the optimizer in the state in order to access the hparams (or copy the hparams over to the state), but that's probably not a big deal.

src/interface.jl
```diff
@@ -29,3 +31,7 @@ end
 
 # default all rules to first order calls
 apply(o, state, x, dx, dxs...) = apply(o, state, x, dx)
+
+isnumeric(x::AbstractArray{<:Number}) = isleaf(x)  # isleaf to allow for e.g. transposed shared weights
+isnumeric(x::AbstractArray{<:Bool}) = false  # convention of ChainRules is that Bool is non-differentiable
+isnumeric(x) = false
```
mcabbott, line 37: BTW, I think these functions would ideally move up to Functors, since they are also what you want for a flatten function:  https://github.com/FluxML/Functors.jl/pull/31

test/runtests.jl
ToucheSir: We've talked about doing away with that short version completely, so I wouldn't weigh it into the decision.

src/interface.jl
ToucheSir: This ties into the more philosophical question of how much Functors.jl is a generic structural traversal library vs a specialized module system for NNs. That said, I don't see any harm in moving `isnumeric` over.

test/runtests.jl
mcabbott: There is also this (undocumented?) path to update optimiser parameters by keyword:
```julia
julia> o = ADAM()
ADAM{Float32}(0.001f0, (0.9f0, 0.999f0), 1.1920929f-7)

julia> ADAM(o, eta=1/3)
ADAM{Float64}(0.3333333333333333, (0.8999999761581421, 0.9990000128746033), 1.1920928955078125e-7)
```
That also seems a bit obscure. If you do need it, perhaps it should be a walk over the state, like
```
newstate = Optimisers.state(ADAM(1/3), st, w)
```

ToucheSir: I have a vague recollection of this, cc @darsnack in case he remembers more. As I recall the plan for scheduling hparams was just to create a new opt struct at every schedule update. I still think co-locating opt and state make sense, as traversal seems to be cheap and it gets us optimizer mixing + matching for free.

src/interface.jl
```diff
@@ -1,31 +1,37 @@
...
     return _update(o, state, x, x̄s...)
   else
-    x̄s = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
-    x, restructure = functor(typeof(x), x)
-    xstate = map((state, x, x̄s...) -> update(o, state, x, x̄s...), state, x, x̄s...)
-    return map(first, xstate), restructure(map(last, xstate))
+    x̄s′ = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
+    x′, re = functor(typeof(x), x)
+    xstate = map((stᵢ, xᵢ, x̄sᵢ...) -> update(o, stᵢ, xᵢ, x̄sᵢ...), state, x′, x̄s′...)
+    return map(first, xstate), re(map(last, xstate))
```
mcabbott, line 28: Last commit should change nothing, it just gives new names to new variables, so as to be less confusing. I think occasionally this helps type-inference not to be confused, too, but haven't tested.

test/runtests.jl
darsnack: I've always been against the `o(st, w, gs...)` syntax

mcabbott: Excellent. I'll make another PR, but perhaps this one is ready to go?

src/interface.jl
mcabbott: Yes. No rush I guess; Optimisers certainly needs this concept somewhere.

ToucheSir (approved):
:shipit:

test/runtests.jl
darsnack: I think some way to say "give me o but with o.eta = x" is a good idea, but the current ADAM(o, eta = x) way is very confusing. I would be in favor of eliminating it for now and agreeing upon a syntax later when we bring in schedulers. There is no need to consider schedulers for the design here.

ToucheSir: Push comes to shove we point users to Setfield or Accessors in the docs.

PR #30, opened 2022-01-27
Wrap optimiser and state in a struct

mcabbott:
As discussed here https://github.com/FluxML/Optimisers.jl/pull/29#discussion_r793256133, this removes the need to explicitly pass the optimiser itself, by wrapping it up with the state. Then `Optimisers.update` takes 3 arguments by default, not 4.

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -97,7 +97,7 @@ Uses the optimiser and the gradient to change the trainable parameters in the mo
 Returns the improved model, and the optimiser states needed for the next update.
 The initial tree of states comes from [`setup`](@ref).
 
-This is used in exactly the same manner as [`update`](@ref), but it because it may mutate
+This is used in exactly the same manner as [`update`](@ref), but because it may mutate
 arrays within the old model (and the old state), it will be faster for models of ordinary
 `Array`s or `CuArray`s. However, you should not rely on the old model being fully updated.
 """
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -10,4 +10,97 @@ export Descent, ADAM, Momentum, Nesterov, RMSProp,
        ADAGrad, AdaMax, ADADelta, AMSGrad, NADAM, ADAMW, RADAM, OADAM, AdaBelief,
        WeightDecay, ClipGrad, ClipNorm, OptimiserChain
 
+"""
+    Optimisers.apply!(rule::RuleType, state, parameters, gradient) -> (state, gradient)
+
+This defines the action of any optimisation rule. It should return the modified gradient
+which will be subtracted from the parameters, and the updated state (if any) for use at
+the next iteration, as a tuple `(state, gradient)`.
+
+For efficiency it is free to mutate the old state, but only what is returned will be used.
+Ideally this should check `iswriteable(x)`, which the built-in rules do via [`@..`](@ref).
+
+The initial state is `init(rule::RuleType, parameters)`.
+
+# Example
+```jldoctest
+julia> Optimisers.init(Descent(0.1), [1,2,3]) === nothing
+true
+
+julia> Optimisers.apply!(Descent(0.1), nothing, [1,2,3], [4,5,6])
+(nothing, Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}}(*, ([4, 5, 6], 0.1)))
+```
+"""
+apply!
+
+"""
+    Optimisers.init(rule::RuleType, parameters) -> state
+
+Sets up the initial state for a given optimisation rule, and an array of parameters.
+This and [`apply!`](@ref) are the two functions which any new optimisation rule must define.
+
+# Examples
+```jldoctest
+julia> Optimisers.init(Descent(), [1,2,3])  # is `nothing`
+
+julia> Optimisers.init(Momentum(), [1.0, 2.0])
+2-element Vector{Float64}:
+ 0.0
+ 0.0
+```
+"""
+init
+
+"""
+    Optimisers.setup(rule, model) -> tree
+
+Initialises the given optimiser for every trainable parameter within the model.
+Returns a tree of the relevant states, which must be passed to [`update`](@ref)
+or [`update!`](@ref).
+
+# Example
+```jldoctest
+julia> Optimisers.setup(Descent(0.1f0), (x = rand(3), y = (true, false), z = tanh))
+(x = Leaf(Descent{Float32}(0.1), nothing), y = (nothing, nothing), z = nothing)
+```
+"""
+setup
+
+"""
+    Optimisers.update(tree, model, gradient) -> (tree, model)
+
+Uses the optimiser and the gradient to change the trainable parameters in the model.
+Returns the improved model, and the optimiser states needed for the next update.
+The initial tree of states comes from [`setup`](@ref).
+
+See also [`update!`](@ref), which will be faster for models of ordinary `Array`s or `CuArray`s.
+
+# Example
+```jldoctest
+julia> m = (x = Float32[1,2,3], y = tanh);
+
+julia> t = Optimisers.setup(Descent(0.1f0), m)
+(x = Leaf(Descent{Float32}(0.1), nothing), y = nothing)
+
+julia> g = (x = [1,1,1], y = nothing);  # fake gradient
+
+julia> Optimisers.update(t, m, g)
+((x = Leaf(Descent{Float32}(0.1), nothing), y = nothing), (x = Float32[0.9, 1.9, 2.9], y = tanh))
+```
+"""
+update
+
+"""
+    Optimisers.update!(tree, model, gradient) -> (tree, model)
+
+Uses the optimiser and the gradient to change the trainable parameters in the model.
+Returns the improved model, and the optimiser states needed for the next update.
+The initial tree of states comes from [`setup`](@ref).
+
+This is used in exactly the same manner as [`update`](@ref), but it because it may mutate
+arrays within the old model (and the old state), it will be faster for models of ordinary
+`Array`s or `CuArray`s. However, you should not rely on the old model being fully updated.
+"""
+update!
+
 end # module
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -4,14 +4,16 @@ struct Leaf{R,S}
   state::S
 end
 
-function setup(rule, x)
+function setup(rule, x; seen = Base.IdSet())
   if isnumeric(x)
+    x in seen && throw(ArgumentError("Optimisers.jl does not at present handle tied weights, sorry."))
+    isbits(x) || push!(seen, x)
     return Leaf(rule, init(rule, x))
   elseif isleaf(x)
     return nothing
   else
     x′, _ = functor(x)
-    return map(xᵢ -> setup(rule, xᵢ), x′)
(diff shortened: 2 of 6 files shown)


src/interface.jl
```diff
@@ -11,20 +18,23 @@ function state(o, x)
...
+  return State(o, st′), patch(x, x̄′)
 end
 
-function update(o, state, x::T, x̄s...) where T
+function update(state, x::T, x̄s...) where T
   if all(isnothing, x̄s)
     return state, x
   elseif isnumeric(x)
-    return _update(o, state, x, x̄s...)
+    return _update(state, x, x̄s...)
```
mcabbott, line 33: It's tempting to delete the underscore and have a method `update(s::State, x, x̄s...)` which is called when there is state, which implies x isnumeric. That will miss the `isnothing` check for zero gradient, though, so it would have to be duplicated.

darsnack:
Slight bike shedding comment: what about the framing being `Optimiser(rule, state)`? So, `ADAM` etc. are "update rules" that when combined state initialized on some `x` makes a complete "optimizer."

mcabbott:
So you want to reserve the word "state" for what `init` returns, and call ADAM a "rule", and call the combined thing an "optimiser".

That might be better. It is a bit weird to use "state" to mean two things.

The current rule descriptions all say things like "version of the ADAM optimiser". Maybe "optimiser" isn't the perfect word for the combined thing, but what is?

darsnack:
> Maybe opt = Optimisers.setup(ADAM(), model) would be nice?

This is currently there in form of `init` + `state`. `init` is the low level function that actually returns the state for a leaf. `state` is the user facing function (equivalent to your `setup`). I do think `setup` makes a bit more sense.

> The current rule descriptions all say things like "version of the ADAM optimiser". Maybe "optimiser" isn't the perfect word for the combined thing, but what is?

I think it makes sense. In the current non-functional understanding, our optimizers are just rules + state.

@ToucheSir and I have always pushed that these functional optimizers are really more like gradient transformation rules. Under this framing, things like `WeightDecay` or `ClipGrad` make sense.

mcabbott:
How about "store" for the name of the complete thing, rule + state?

darsnack:
Hmm seems like a verb being squeezed into a noun. With `setup`, I suppose no one needs to type this name out. It could just be `RuleAndState` or `RuleState`

ToucheSir:
+1 for `RuleState`.

mcabbott:
I was thinking took "store" as a noun (larder, storehouse, stockpile). RuleState is also OK. It could also be `Setup`, since that's also a noun, perhaps that's even cleaner?

src/interface.jl
mcabbott: https://github.com/FluxML/Optimisers.jl/pull/30/commits/395652610c84f267014c754d4a891ac6cc66e203 makes this change. Maybe it's cleaner?

darsnack:
> I was thinking took "store" as a noun (larder, storehouse, stockpile).

I guess this is murky due to colloquialisms 😄.

`Setup` is also okay by me if we want something short.

ToucheSir:
This is getting into ultra bikeshed territory, but "`update` with `Setup`" reads more awkwardly to me than "`update` with `RuleState`".

mcabbott:
Ok, we think a bit. Most importantly, what single letter variable name does it get?

ToucheSir:
If I may so bold to suggest an multi-letter initialism instead. So `st`/`sp` for `Setup`, or `rs` for `RuleState`/`RuleAndState`.

src/interface.jl
```diff
@@ -1,31 +1,44 @@
-patch(x, x̄) = x .- x̄
 
-function state(o, x)
+struct Leaf{R,S}
+  rule::R
+  state::S
+end
+function Base.show(io::IO, ℓ::Leaf)
+  ioc = IOContext(io, :compact => true)
+  print(io, "Leaf("); show(ioc, ℓ.rule); print(io, ", "); show(ioc, ℓ.state); print(io, ")")
```
ToucheSir, line 8: Not worth skipping the linebreaks here IMO.

ToucheSir:
Want to give Kyle a chance to walk through this, but otherwise looks good. Do we want to include tweaks to https://github.com/FluxML/Optimisers.jl/blob/master/docs/src/index.md in this PR or a follow-up?

src/interface.jl
mcabbott: Yes probably. I moved it to the bottom of the file, as it really isn't essential to the first read.

mcabbott:
I was fiddling a little with the docs. I guess trying to explain better that you *define* array-level functions, and *call* model-level functions. I made up a stupid optimiser which has at least some state, maybe there's a bette one. Far from perfect, but see what you think.

src/interface.jl
```diff
@@ -1,39 +1,46 @@
...
+  if all(isnothing, x̄s)
+    return ℓ, x
+  else
+    s′, x̄′ = apply!(ℓ.rule, ℓ.state, x, x̄s...)
+    return Leaf(ℓ.rule, s′), subtract!(x, x̄′)
+  end
 end
 
-function update!(o, state, x, x̄s...)
+function update!(tree, x, x̄s...)
```
mcabbott, line 29: Re the bikeshed committee... Everything used to be called state. This method which walks... maybe its argument should be `tree`, in which case maybe the other method, which applies, acts on leaves?

It's not super-explicit, but it's not misleading. `RuleAndState` does not spark joy... I always feel a bit like we're vandalising Mike's haiku...

ToucheSir: Py(Tree) is what JAX uses too. Technically it can be a DAG, but probably in our best interest not to acknowledge that ;)

mcabbott: If I'm thinking correctly, this one really should be a tree, even if the model is not. Once everything is working... tied parameters will have only one optimiser node.

mcabbott: Oh and I meant to explicitly check there aren't duplicates, until we can handle them.

src/interface.jl
```diff
@@ -4,14 +4,16 @@ struct Leaf{R,S}
   state::S
 end
 
-function setup(rule, x)
+function setup(rule, x; seen = Base.IdSet())
   if isnumeric(x)
+    x in seen && throw(ArgumentError("Optimisers.jl does not at present handle tied weights, sorry."))
```
ToucheSir, line 9:
```suggestion
    iscachesafe(x) && x in seen && throw(ArgumentError("Optimisers.jl does not at present handle tied weights, sorry."))
```
Where:
```julia
iscachesafe(x) = !isbits(x) && ismutable(x)
iscachesafe(::Union{String,Symbol}) = false
```
To prevent false positives on value types.

ToucheSir: This should probably be in functors and used liberally whenever a cache is in play.

mcabbott: Oh I forgot about strings. I don't think ismutable tells you much, as an immutable wrapper can hold a mutable object.

mcabbott: No it's fine. Strings can't be `isnumeric`. And I test `isbits` before pushing into the cache.

ToucheSir: I was backfilling this implementation from work on functors and missed the `isnumeric` guard above. Would recommend checking for `isbits` first before doing anything requiring `objectid`, as `objectid` is significantly slower for value types (involves a ccall and hashing):
```julia-repl
julia> x = ntuple(identity, 100);

julia> @btime objectid($x)
  99.999 ns (0 allocations: 0 bytes)
0x31d9b387e6e6f021

julia> @btime isbits($x)
  1.152 ns (0 allocations: 0 bytes)
```
How about:
```julia
if !isbits(x)
  x in seen && throw(ArgumentError("Optimisers.jl does not at present handle tied weights, sorry."))
  push!(seen, x)
end
```

src/interface.jl
mcabbott: When setup learns to look for duplicates, and write down their positions for the necessary grad accumulation steps, it will have to record these somewhere. So probably we will ultimately need another struct to hold such things alongside the nested-tuples matching the model. For which Tree is then the obvious name.

src/interface.jl
mcabbott: Agree this is faster, but... Zygote doesn't really work with SArrays right now, so it's just numbers, and this runs exactly once while Functors's `fmap` might run several times per iteration... so my vote is that this doesn't really merit much optimisation yet.

src/Optimisers.jl
```diff
@@ -10,4 +10,97 @@ export Descent, ADAM, Momentum, Nesterov, RMSProp,
...
+# Example
+```jldoctest
+julia> Optimisers.init(Descent(0.1), [1,2,3]) === nothing
+true
+
+julia> Optimisers.apply!(Descent(0.1), nothing, [1,2,3], [4,5,6])
+(nothing, Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}}(*, ([4, 5, 6], 0.1)))
+```
+"""
+apply!
```
mcabbott, line 34: Doesn't have to be the same PR, but I added docstrings for all the basic functions. They are also in the manual, but it's good to have something at the repl too, I think.

I put them all in the main file. Seems more neutral than attaching them to `apply!(::Descent, ...)` or something.

src/interface.jl
ToucheSir: I thought Zygote could project back out to SArrays now? But yes, `fmap`s cache usage is a better optimization target and may have already caused performance issues (there's an issue about slow `fcollect` somewhere).

mcabbott: It tries, and wraps many in SizedArray. But IIRC many things fail, e.g. broadcasting uses `sum(SA[1 2; 3 4]; dims=(1,3))` (ndims+1 as a hack to get type-stable dims).

src/Optimisers.jl
```diff
@@ -10,4 +10,97 @@ export Descent, ADAM, Momentum, Nesterov, RMSProp,
...
+update
+
+"""
+    Optimisers.update!(tree, model, gradient) -> (tree, model)
+
+Uses the optimiser and the gradient to change the trainable parameters in the model.
+Returns the improved model, and the optimiser states needed for the next update.
+The initial tree of states comes from [`setup`](@ref).
+
+This is used in exactly the same manner as [`update`](@ref), but it because it may mutate
```
darsnack, line 100:
```suggestion
This is used in exactly the same manner as [`update`](@ref), but because it may mutate
```

src/Optimisers.jl
```diff
@@ -10,4 +10,97 @@ export Descent, ADAM, Momentum, Nesterov, RMSProp,
...
+# Example
+```jldoctest
+julia> Optimisers.setup(Descent(0.1f0), (x = rand(3), y = (true, false), z = tanh))
+(x = Leaf(Descent{Float32}(0.1), nothing), y = (nothing, nothing), z = nothing)
+```
+"""
+setup
+
+"""
+    Optimisers.update(tree, model, gradient) -> (tree, model)
```
darsnack, line 70: Should we mention higher order terms in the signature?

darsnack (approved):
Looks good to me!

src/Optimisers.jl
mcabbott: My vote is no, at least for now. There could be a section somewhere (once we have an example? tests?) but I'd like to first cover the basic who-does-what picture.

PR #31, opened 2022-01-27
Add lazy broadcasting

mcabbott:
This lets the each optimiser step build a `Broadcasted` object, which is ultimately evaluated once, in the usual way. They should never escape, but they do save memory:
```
julia> m = (α = rand(100,100), β = rand(100,100));

julia> o = OptimiserChain(WeightDecay(), ADAM());

julia> st = Optimisers.state(o, m);

julia> gr = (α = rand(100,100), β = rand(100,100)); # fake gradient

julia> @btime Optimisers.update($o, $st, $m, $gr);
  min 115.250 μs, mean 180.870 μs (55 allocations, 470.42 KiB)  # before
  min 108.875 μs, mean 126.485 μs (46 allocations, 158.27 KiB)  # used for optimisers
  min 43.750 μs, mean 47.868 μs (46 allocations, 3.86 KiB)      # with in-place model update
```
The same macro also inserts a check that it is possible for optimisers to write into their own state arrays, before doing so. Right now they always try, and they make arrays like `zero(x)`, so will fail for e.g. StaticArrays.

We could similarly use that to mutate the arrays of the model, when possible. But perhaps that wants to be optional even if it's the default, maybe it's `update(opt, state, model, grad; copy=false)`?

It's possible that this option ought to propagate down to every rule, to also control whether they mutate their state arrays. That would mean passing it to every `apply` method, which seems messy.


src/rules.jl
```diff
@@ -44,9 +44,9 @@ init(o::Momentum, x::AbstractArray) = zero(x)
 
 function apply(o::Momentum, state, x, dx)
   η, ρ, v = o.eta, o.rho, state
-  @. v = ρ * v - η * dx
+  v′ = @.. v = ρ * v - η * dx
```
ToucheSir, line 47: It always felt odd to me that state was unconditionally mutated even in the non `!` version of `apply`. Perhaps we should take an "all or nothing" approach where `update!`/`apply!` mutate both gradients and state, while `update`/`apply` mutate nothing. The holy grail would be figuring out how to do that with minimal code duplication.

src/interface.jl
```diff
@@ -35,3 +35,38 @@ apply(o, state, x, dx, dxs...) = apply(o, state, x, dx)
 isnumeric(x::AbstractArray{<:Number}) = isleaf(x)  # isleaf to allow for e.g. transposed shared weights
 isnumeric(x::AbstractArray{<:Bool}) = false  # convention of ChainRules is that Bool is non-differentiable
 isnumeric(x) = false
+
+iswriteable(::DenseArray{<:AbstractFloat}) = true
+iswriteable(_) = false
```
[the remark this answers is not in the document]
darsnack: Same, for all the `is*` methods, I think it is best to use ArrayInterface.jl or even build on top of it if we need redirection.

src/rules.jl
mcabbott: One way you might achieve this is by specifying it at `state` or `setup` or whatever call. If that wraps each `x` in some NoWriteWrapper, then maybe `zero(x)` would be so too? Maybe it won't stick though, or if it propagates, will land up in your model too.

Or the macro could be fancier, it could test a local variable called `copy` which every `apply` function would have as a (keyword) argument. You'd still have to add that but wouldn't have to add `if else` one by one.

My guess is that the mutate-everything version will be the main one, at least for now? You get the new model back but must regard the old as junk. But at least you write things in a way which will work with immutables.

darsnack:
If we are going to "maybe mutate" then I suggest appending `!` to the function names to signal that.

src/interface.jl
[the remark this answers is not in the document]
mcabbott: I'm not keen on this, it's a huge package. I believe ChainRulesCore also has such a function, which it might be reasonably to expect anyone doing AD to care about?

But for now I vote to leave it low-tech -- I don't think this can lead to an error, it just won't write into MArrays.

mcabbott:
Having `update!` and `update` would be fine as an interface, but I'm keen to avoid having `apply!` and `apply` for every optimiser. There should only be one place you can make sign errors.

src/rules.jl
ToucheSir: Agreed on mutate-everything being the default. We can always provide a `function update` (sans colon) later that does `update!(fmap(copy_if_mutable, state), model, fmap(copy_if_mutable, grads))` later. Then only `apply!` would be required. If ever in future we get `ImmutableArray`s or `XLAArray`s, those should just work :tm: because of the `ismutable` trait.

darsnack:
I agree, there should only be `apply!`.

src/interface.jl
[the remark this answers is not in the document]
ToucheSir: I believe just this section should handle both MArrays and CuArrays: https://github.com/JuliaArrays/ArrayInterface.jl/blob/2a5777ff2f512e6f9e7946e651bf40367da24d49/src/ArrayInterface.jl#L124-L148. We definitely want in-place updates for GPU arrays.

darsnack:
Would it be slightly less spooky if we use something global to the package like `use_cuda` in Flux? Having a dangling keyword argument in the interface is not so great for people wanting to write their own rules.

src/rules.jl
mcabbott: OK last commit adds the magic copy idea. Maybe too magic, not sure. It will try to write into Arrays, and not write into ranges etc. But if you set `let copy=true` in the macro's scope, then it will never copy. So we must add a `copy` keyword to every `apply` function, but that's all.

mcabbott:
So the last commit makes the interface `update` fully out of place, and `update!` fully overwriting (when possible). 

It imposes on anyone else writing a rule that they must accept this keyword (and ideally obey it), but nothing too weird will happen -- you will get an ordinary MethodError if your new `apply` cannot be called, same as if you made other mistakes.

ToucheSir:
My thought in https://github.com/FluxML/Optimisers.jl/pull/31#discussion_r794057657 was to have rules always try mutation, then provide a higher-level interface that defensively copies any mutable arrays in state/gradients before calling `update!` and `apply!`. Thoughts?

darsnack:
I would prefer that approach or checking `Optimisers.use_mutation[]` in the macro to the magic local keyword. Just a little too magical for me.

mcabbott:
For defensive copies, I guess it's a reasonable assumption that any optimiser planned to mutate any array it put in its state. What you do need is a way to access them, i.e. you need to assume that some `fmap(maybecopy, state)` is going to work. Which you can document. The typical `apply!` will then contain things like `v′ = @.. v = ρ * v` with the (less magical) macro mutating v if mutable -- with the same source of truth as the defensive `maybecopy` process.

src/interface.jl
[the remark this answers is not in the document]
mcabbott: I believe CuArrays are DenseArrays, so they should already work. Mine also restricts to floats, so that if you happen to initialise with an array of integers, it won't try to write numbers from the gradient in there, even if in some sense it is mutable. 

I don't know how much work beyond that is justified, until we have a use case that cares.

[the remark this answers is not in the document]
ToucheSir: Unfortunately host-side CuArrays are not (ref. https://github.com/JuliaGPU/CUDA.jl/blob/master/src/array.jl#L27), so unless we want to use Requires (or make Flux implement `Optimisers.iswriteable(::CuArray) = true`) a more thorough check is necessary. So adding a check for wrappers + whether or not the (wrapped) type is mutable would probably be sufficient.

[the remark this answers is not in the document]
mcabbott: But that's still DenseArray? https://github.com/JuliaGPU/GPUArrays.jl/blob/35f25abfa22ddd3814b20034e0011de21aa44194/src/host/abstractarray.jl#L12

src/interface.jl
```diff
@@ -11,27 +11,66 @@ function state(o, x)
...
 
-function update(o, state, x::T, x̄s...) where T
+function update(o, state, x::T, x̄s...; copy::Bool = true) where T
   if all(isnothing, x̄s)
     return state, x
   elseif isnumeric(x)
-    return _update(o, state, x, x̄s...)
+    return _update(o, state, x, x̄s...; copy)
   else
     x̄s′ = map(x̄ -> functor(typeof(x), x̄)[1], x̄s)
```
ToucheSir, line 26:
```suggestion
    x̄s′ = map(x̄ -> functor(T, x̄)[1], x̄s)
```
Or remove the type param entirely

ToucheSir:
I skimmed the stdlib, StaticArrays and CUDA.jl. All three seem to have functioning `copy` functions, so here's a sketch of `update`:
```julia
maybecopy(x) = iswriteable(x) ? copy(x) : x
update(o, state, x, x̄s...) = update!(o, fmap(maybecopy, state, exclude=isnumeric), x, fmap(maybecopy, x̄s, exclude=isnumeric)...)
```

mcabbott:
Oh that's another way. Maybe tidier than doing it at the `_update!` stage which acts on one trainable array, and you only pay for one `fmap` cache...

We could go further and simply demand that the state be a tuple or a NamedTuple, always, and call `map`. That might be less mysterious. E.g. if your state has some struct and you forgot to `@functor` it, then these other ways may silently give you unintentional mutation... but `map` will loudly fail the first time you try.

src/interface.jl
mcabbott: Oh right. Actually that's one of the lines I'm still not sure I understand. (Because exactly what `@functor` defines is cleverly hidden within a function which calls `eval`...)

ToucheSir:
For better or worse, `fmap` is still needed for handling nested namedtuples. Un-functored structs should be safe as long as we guard with `isleaf(x) && !isnumeric(x)` (the default definition of `functor` returns an empty tuple of children for this reason).

mcabbott:
Maybe that's fine, `update` is going to allocate so much that an `IdDict` is really no concern. I've adopted this `fmap` strategy. Maybe it's even a good thing if it walks not just the trainable arrays (once we restore that distinction). But `fmap` with `exclude` isn't in the tagged version?

And I updated all the rules. Besides tests it might be worth someone else looking over them in case there are mistakes.

Project.toml
```diff
@@ -1,19 +1,17 @@
 name = "Optimisers"
 uuid = "3bd65402-5787-11e9-1adc-39752487f4e2"
 authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
-version = "0.1.0"
+version = "0.2.0"
```
mcabbott, line 4: Thoughts on versions? v0.1 is tagged, but so broken that surely nobody can use it.

src/interface.jl
ToucheSir: Unfortunately the macro expansion is obscured behind an intermediate function, but the translation is pretty straightforward:
```julia
struct Foo
  a
  b
  c
end
@functor Foo (a, c)

# becomes
Functors.functor(::Type{Foo}, x) = (a=x.a, c=x.c), y -> Foo(y[1], b=x.b, y[2])
```
If there's a cleaner and clearer way to pull this off, we should definitely do it.

Project.toml
ToucheSir: Rip off the band-aid, I say :laughing:

src/interface.jl
```diff
@@ -11,27 +10,68 @@ function state(o, x)
...
 
-function _update(o, st, x, x̄s...)
-  st′, x̄′ = apply(o, st, x, x̄s...)
-  return st′, patch(x, x̄′)
+patch!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : (x .- x̄)
+
+function _update!(o, st, x, x̄s...)
+  st′, x̄′ = apply!(o, st, x, x̄s...)
+  return st′, patch!(x, x̄′)
 end
```
ToucheSir, line 18: I think you mentioned it at some point, but does this function still have to exist? Material for a follow-up PR?

mcabbott: #30 wants to delete the underscore, and have this method (which acts on on array) invoked by dispatch instead of the method just below which walks the tree.

ToucheSir (approved):
One non-blocking question, but otherwise everything makes sense and the savings speak for themselves!

test/runtests.jl
```diff
@@ -20,7 +38,11 @@ Random.seed!(84)
       st, w = Optimisers.update(o, st, w, gs...)
     end
     lw = loss(w, w′)
-    @test lw < 0.001
+    if o isa ADADelta
+      @test_broken lw < 0.001
```
mcabbott, line 42: This one goes to NaN, not sure why. Only on 1.6, with the original seed. Now on nightly too, with a different seed.

src/interface.jl
ToucheSir: Alrighty, let's move the discussion there after this is merged.

src/interface.jl
mcabbott: Ok. This whole recursion isn't quite right yet, fails on the Resnet example from the docs, but I haven't isolated why yet.

test/runtests.jl
ToucheSir: Not sure either. Let's leave this for a separate PR that uses StableRNGs instead.

PR #39, opened 2022-01-29
Add some tests using StaticArrays

mcabbott:
Testing with StaticArrays exposes some bugs (?) in how this handles element type:
* ~~Some `OptimiserChain` combinations promote to Float64.~~ when they have explicit Float64 parameters
* Most optimisers promote Float16, as their default parameters are Float32.

The PR also divides the tests into two files, one of which has loops over all optimisers, the other tries to test that the framework is doing what it should do.

Many of the optimiser convergence tests simply mark broken all that fail to converge. It's possible that some of these are real bugs. I'm surprised that minimising `sum(abs2, x + x')` isn't easy for all of them.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -16,7 +16,7 @@ Descent() = Descent(1f-1)
 init(o::Descent, x::AbstractArray) = nothing
 
 function apply!(o::Descent, state, x, dx)
-  η = convert(float(eltype(dx)), o.eta)
+  η = convert(float(eltype(x)), o.eta)
   
   return state, @.. dx * η
 end
@@ -477,7 +477,7 @@ ClipGrad() = ClipGrad(10f0)
 init(o::ClipGrad, x::AbstractArray) = nothing
 
 function apply!(o::ClipGrad, state, x, dx)
-  δ = convert(float(eltype(dx)), o.delta)
+  δ = convert(float(eltype(x)), o.delta)
   dx′ = @.. clamp(dx, -δ, δ)
 
   return state, dx′
--- a/test/rules.jl
+++ b/test/rules.jl
@@ -97,3 +97,34 @@ end
     end
   end
 end
+
+@testset verbose=true "type promotion" begin
+  @testset "$(name(o))" for o in RULES
+    marray = (f32 = Float32[1,2], f64 = Float64[3,4]) #, f16 = Float16[5,6])
+    types = map(eltype, marray)
+
+    # This is a weak test, as it copies & then does `update!`
+    uparray = Optimisers.update(Optimisers.setup(o, marray), marray, marray)[2]
+    map(eltype, uparray) == types
+
+    # Static version is truly out-of-place:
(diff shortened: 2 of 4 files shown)


test/runtests.jl
```diff
@@ -13,6 +13,7 @@ Functors.@functor TwoThirds (a, c)
 Optimisers.trainable(x::TwoThirds) = (a = x.a,)
 
 @testset verbose=true "Optimisers.jl" begin
+@testset verbose=true "Features" begin
```
mcabbott, line 17: The indents are wrong due to this testset.

test/rules.jl
```diff
@@ -0,0 +1,130 @@
...
+    # This is a weak test, as it copies & then does `update!`
+    uparray = Optimisers.update(Optimisers.setup(o, marray), marray, marray)[2]
+    map(eltype, uparray) == types
+
+    # Static version is truly out-of-place:
+    mstatic = (f32 = SA{Float32}[1,2], f64 = SA{Float64}[3,4]) # , f16 = SA{Float16}[5,6])  with Float16, all fail
+    upstatic = Optimisers.update(Optimisers.setup(o, mstatic), mstatic, mstatic)[2]
+    if o isa OptimiserChain && o.opts[2] isa ADAM  # These promote to Float64
+      @test_broken map(eltype, upstatic) == types
+    else
```
mcabbott, line 115: These make Float64:
```
      WeightDecay → ADAM  |    2       1      3     3.4s
      ClipNorm → ADAM     |    2       1      3     2.2s
 ```

File: test/rules.jl [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+    for _ in 1:10^5
+      g = gradient(x -> sum(abs2, x + x'), m)[1]
+      s, m = Optimisers.update!(s, m, g)
+    end
+    # @test sum(m) < sum(1:64)
+    if sum(m) < 1
+      @test sum(m) < 1
+    else
+      @show name(o) sum(m)/sum(1:64)
+      @test_broken sum(m) < 1
```
mcabbott, line 48: ADAGrad, ADADelta, AMSGrad fail. 

With fewer steps, more fail.

src/rules.jl
```diff
@@ -16,7 +16,7 @@ Descent() = Descent(1f-1)
 init(o::Descent, x::AbstractArray) = nothing
 
 function apply!(o::Descent, state, x, dx)
-  η = convert(float(eltype(dx)), o.eta)
+  η = convert(float(eltype(x)), o.eta)
```
mcabbott, line 19: The lazy broadcast does not have eltype, so this has to come from `x`.

test/rules.jl
mcabbott: Because it's ADAM(0.001) explicitly.

test/rules.jl
```diff
@@ -0,0 +1,157 @@
...
+      @test_broken lw < 0.001
+    else
+      @test lw < 0.001
+    end
+  end
+end
+
+@testset verbose=true "StaticArrays" begin
+  @testset "$(name(o))" for o in RULES
+
```
ToucheSir, line 78:
```suggestion
```

ToucheSir (approved):

PR #41, opened 2022-01-30
Allow types from ChainRules

### LINKED ISSUE #32: Support ChainRules types
Mosting handling `NoTangent` and `ZeroTangent`.

mcabbott:
This is a fairly minimal change to allow use with e.g. Diffractor. It just converts all Tangents back to NamedTuples.

It replaces some `all(isnothing)` with dispatch, as this seems a little tidier. (Instead of writing a new `isZero`, to allow also CRC's zero types.)

~~Needs tests.~~ Has a few tests. I tried some more complicated ones with Diffractor, but that's not registered. More later?

Closes #32

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,4 +1,9 @@
 
+using ChainRulesCore: canonicalize, backing, Tangent, AbstractZero
+base(dx::Tangent) = backing(canonicalize(dx))
+base(dx) = dx
+const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
+
 struct Leaf{R,S}
   rule::R
   state::S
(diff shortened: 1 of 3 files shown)


src/interface.jl
```diff
@@ -18,26 +23,21 @@ end
 
 subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : (x .- x̄)
 
+update!(::Nothing, x, ::Zero...) = nothing, x
 update!(::Nothing, x, x̄s...) = nothing, x
 
+         update!(ℓ::Leaf, x, ::Zero...) = ℓ, x
 function update!(ℓ::Leaf, x, x̄s...)
```
mcabbott, line 30: If we had a formatting robot, it would certainly hate my weird indents here. Possibly too weird?

ToucheSir: Yes, I'd say so. One way to ensure alignment would be changing to 
```julia
function update!(ℓ::Leaf, x, ::Zero...)
  return ℓ, x
end
```
But that's overkill compared to keeping the one-liner and stomaching the mismatched indentation.

mcabbott: Turns out you can also write
```
function ff(x::Int) "stuff" end
```
but I'm not sure if that's better or worse.

ToucheSir: VSCode's formatter will change that to a multiline function, though there may be a way to stop it using JuliaFormatter config options. Another technically feasible option (but one I'd prefer to avoid for consistency reasons) would be changing `function update!(ℓ::Leaf, x, x̄s...)` to `update!(ℓ::Leaf, x, x̄s...) = begin` on the following line.

mcabbott: Yes. I don't know. Leave it for now and see if we hate it? 

There is some chance this is a performance disaster, I haven't checked. I mean everything Functors.jl is type-unstable, and perhaps TTFG is more important than compiled speed... but I haven't set up any cases to time & compare.

ToucheSir: To be honest I already don't like the cut of its jib, so "for now" is a bit of a tough sell :smile:. How about grouping the 2 one-liners (leaf and tree) together?

Unfortunately Functors wasn't really designed with type stability in mind, so trying to fix that is hard without an extensive redesign of the internals. I have a mind to do so, but it won't be quick since everything is constraint by backwards compat.

mcabbott: Ok, I really don't mind much. I've pushed them to the left now.

ToucheSir (approved):

src/interface.jl
mcabbott: Re type stability, it's not even obvious that one should fix it. Setfield et al aim to be useful for small structs in tight loops, but huge nested Flux models maybe you get exponential compilation time for no discernible gain?

ToucheSir: That is true, though if we're embracing type instability I'd rather be purposeful about it with e.g. `@nospecialize`. It's also not clear to me whether issues like https://github.com/FluxML/NNlib.jl/pull/370 will ever manifest themselves here. Certainly profiling for networks big (e.g. MetalHead) and small (e.g. whatever AlphaZero.jl uses) would be in order before taking on the latency dragon.

PR #42, opened 2022-01-30
Handle tied weights in `update!`

mcabbott:
This alters `setup` to record the "address" of any tied weights, and then `update!` to first add the gradient of the second to the first (of each pair), then update as normal, and finally re-create the tie in the updated model.

I've tried to match the existing `update!` as much as possible. The format in which the "address" is stored is just a tuple of property names. The function to "pick out" a gradient component based on this is easy. The one to "place back" the modified one is trickier, as it needs to re-create missing branches -- and not just the minimal branch, but all the other empty fields. So it walks the model in parallel to the gradient, a lot like the existing `update!`. This isn't something that e.g. Setfield.jl thinks about.

Maybe this could all be abstracted away somehow, moved up into Functors? We seem to need quite a few patterns which aren't like `fmap`.

Surely the "address" could be stored in a more compile-away-able format, alla Setfield.jl. None of this is type stable, and it takes a few μs. I think that's true of everything in Functors.jl too. And it might even be desirable, for startup speed with deeply nested models. 

~~This is on top of #41.~~ Uses https://github.com/FluxML/Functors.jl/pull/33, which is copied in here for now.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -64,7 +64,7 @@ function update!(t::Tied, x, x̄)
   # accumulate tied gradients:
   for (dup, orig) in t.ties
     x̄ = place(x, x̄, orig) do x̄ᵢ
-      x̄ᵢᵢ = pick(x̄, dup)
+      x̄ᵢᵢ = pick(x, x̄, dup)
       Broadcast.broadcasted(+, base(x̄ᵢ), base(x̄ᵢᵢ))
     end
   end
@@ -98,7 +98,7 @@ iswriteable(_) = false
 ids(x::NamedTuple{names}) where names = NamedTuple{names}(names)  # a map-friendly version of pairs
 ids(x::Tuple) = propertynames(x)
 
-function pick(x, addr::Tuple)  # used for both x and x̄
+function pick(x, addr::Tuple)
   (isempty(addr) || x isa Zero) && return x
   x′, _ = functor(x)
   pick(get(x′, addr[1], nothing), tail(addr))
@@ -108,6 +108,14 @@ place(f, x, addr::Tuple{}) = f()
 function place(f, x, addr::Tuple)
   x′, re = functor(x)
   re(map((xᵢ, i) -> i == addr[1] ? place(f, xᵢ, tail(addr)) : xᵢ, x′, ids(x′)))
+end
+
+# This function needs to see x::Transpose to handle x̄::Matrix, etc.
+function pick(x, x̄, addr::Tuple)
+  (isempty(addr) || x̄ isa Zero) && return x̄
+  x̄′, _ = functor(typeof(x), base(x̄))
+  x′, _ = functor(typeof(x), x)
+  pick(get(x′, addr[1], nothing), get(x̄′, addr[1], nothing), tail(addr))
 end
 
 place(f, x, x̄, addr::Tuple{}) = f(x̄)  # placed into x̄
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -17,7 +17,7 @@ function Base.show(io::IO, ℓ::Leaf)  # show method is mostly to hide its long 
   print(io, ")")
 end
 
-struct Tied; ties; tree; end
+struct Tied; ties; tree; end  # nothing about shared parameters is type-stable
 
 function setup(rule, x; ties = Pair[], cache = IdDict())
   rule isa AbstractRule || Base.depwarn("In future, all optimisation rules should be <: AbstractRule", :setup)
@@ -26,15 +26,12 @@ function setup(rule, x; ties = Pair[], cache = IdDict())
 end
 
 function _setup(rule, x, addr; ties, cache)
-  usecache = !isbits(x) && cache !== false
   if isnumeric(x)
-    if usecache && haskey(cache, x)
+    if haskey(cache, x)
       push!(ties, addr => cache[x])
       return nothing
     else
-      if usecache
-        cache[x] = addr
-      end
+      cache[x] = addr  # unlike the Functors cache, this one is only used for isnumeric objects
       return Leaf(rule, init(rule, x))
     end
   elseif isleaf(x)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -59,25 +59,26 @@ function update!(tree, x, x̄s...)
   map(first, xtree), re(map(last, xtree))
 end
 
-update!(t::Tied, x, ::Zero) = t, x
+update!(t::Tied, x, ::Zero, ::Zero...) = t, x
 function update!(t::Tied, x, x̄)
   # accumulate tied gradients:
-  for (dup, orig) in t.ties
-    x̄ = place(x, x̄, orig) do x̄ᵢ
-      x̄ᵢᵢ = pick(x, x̄, dup)
-      Broadcast.broadcasted(+, base(x̄ᵢ), base(x̄ᵢᵢ))
+  for (β, α) in t.ties
+    x̄ = place(x, x̄, α) do x̄ₐ
+      x̄ᵦ = pick(x, x̄, β)
+      Broadcast.broadcasted(+, base(x̄ₐ), base(x̄ᵦ))
     end
   end
   # run the optimisers:
   t′, x′ = update!(t.tree, x, x̄)
   # restore tied weights:
-  for (dup, orig) in t.ties
-    x′ = place(x′, dup) do
-      pick(x′, orig)
+  for (β, α) in t.ties
+    x′ = place(x′, β) do
+      pick(x′, α)
     end
   end
   Tied(t.ties, t′), x′
 end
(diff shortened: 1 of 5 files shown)


ToucheSir:
Following from our discussion on design, how does composition of states work with this approach? Since `ties`, `pick` and `place` work on absolute addresses from the tree root, it's not clear to me how to keep things in sync when optimizing a subset (e.g. gradual unfreezing) or a superset (e.g. adding a classifier head and fine-tuning) of the original state.

mcabbott:
Good question. These are all from the root. You can't accidentally take a branch of the state & model and lose them, since the outermost Tree type will mean the types don't line up. If instead you pasted these, whole, into some larger model's pair, I bet they would just work? Unless doing that created further ties which neither half saw before.

Explicitly messing with the state tree with your bare hands seems obviously at-own-risk. Better tools for explicitly feezing etc... I guess these would need to interact with this, re-run the check for ties which `setup` does here?

src/interface.jl
```diff
@@ -2,22 +2,38 @@
...
   state::S
 end
 
-function setup(rule, x; seen = Base.IdSet())
+struct Tree; ties; leaves; end
+
+function setup(rule, x; ties = Pair[], cache = IdDict())
+  tree = _setup(rule, x, (); ties, cache)
+  isempty(ties) ? tree : Tree(ties, tree)
+end
```
mcabbott, line 18: Mostly accidentally, this method in fact lets you construct ties where they aren't inferred:
```julia
julia> using StaticArrays

julia> st = Optimisers.setup(Descent(-0.01), (SA[1,2], SA[3,4]), ties = [(2,) => (1,)])
Optimisers.Tree([(2,) => (1,)], (Leaf(Descent{Float64}(-0.01), nothing), Leaf(Descent{Float64}(-0.01), nothing)))

julia> Optimisers.update!(st, (SA[1,2], SA[3,4]), (SA[5,6], SA[7,8]))
(Optimisers.Tree([(2,) => (1,)], (Leaf(Descent{Float64}(-0.01), nothing), Leaf(Descent{Float64}(-0.01), nothing))), ([1.12, 2.14], [1.12, 2.14]))
```

darsnack:
Had the time to go through this and https://github.com/FluxML/Functors.jl/pull/33. I'm happy with both approaches. I had one bike shed comment, but after reading the discussion, I think it's relevant too.

Should `Tied` be something more broad like `InitializedTree`. Something to indicate that this is a tree + auxiliary information? In the future we might consider adding different kinds of auxiliary information. Going over the two cases above:
- Doing a superset like backbone + classifier is probably simple enough to just splat the addresses correctly.
- More complicated is gradual (un)freezing where branches come in and out. In this case, I think it's easier to keep the trees complete and include auxiliary information that says "this branch is frozen, don't recurse or update." This also opens this piece of the design up so that other packages can write iterative freezing algorithms and all they have to do is use utilities like `fsimilar`, etc.

ToucheSir:
I have to admit the implementation is a tad brain-warping, though all that complexity is necessary so there's not much more to say. I agree with Kyle that we should give this a go. If some unforeseen case rears its ugly head then hey, Optimisers.jl isn't anywhere near stable yet and the tied logic is sufficiently decoupled from the core Optimisers functionality.

mcabbott:
Re interaction with other things:

* Storing this information in some struct at the base of the state tree alters its shape, and means that anything else walking this tree needs to know about this storage `struct`, the one which doesn't have a parallel in the model. Perhaps that's an argument against this way of doing it. But storing tie information (say) at the first leaf of each pair instead, seems tricky too --- would the address stored there still be relative to the root? That seems more dangerous, as grafting branches would make it wrong.

* To freeze some weights there's a similar question of where to store the information. It could also be stored at the base (either expanding Tied to have two jobs, or by composing two such things, each with their own `update!` dispatch). Such base storage would need updating on any graft / backbone + classifier  operation. It could also be stored at the leaf, e.g. by adding a "freeze" flag to Leaf. Or it could be stored by inserting a struct at the base of the sub-tree which is entirely frozen. My guess is that per-leaf sounds the simplest in fact. Unlike ties you aren't forced to think about the "address" of particular leaves.

src/interface.jl
```diff
@@ -2,22 +2,38 @@
...
   rule::R
   state::S
 end
 
-function setup(rule, x; seen = Base.IdSet())
+struct Tied; ties; tree; end
+
+function setup(rule, x; ties = Pair[], cache = IdDict())
+  tree = _setup(rule, x, (); ties, cache)
+  isempty(ties) ? tree : Tied(ties, tree)
```
ToucheSir, line 17: Is there a place for a potentially type stable version of `setup`?

mcabbott: I was going to say that all things `fmap` are unstable, but in fact that's not true here, as neither `update` nor `setup` call it. So, maybe. I think it would need you to declare that there are no shared weights, and disable that machine. 

How would it look? `setup(opt, x; cache = false)`, matching (a version of) `fmap`, perhaps too obscure? It will still only run once, vs. update in a loop.

With ties, I thought there is no hope, didn't bother adding types to the `Tied` type since `pick` & `place` are nowhere near type-stable. We could push things into the type domain, e.g. store address as `Tuple{:layers, 1, :weight}`, but I'm scared it will be ugly and take forever to compile and not really save time. At least I'd like to kick that down to road until someone has benchmarks.

ToucheSir: One approach could be changing `struct Tied; ties; tree; end` to `struct Tied{T}; ties::Vector{Pair{Any,Any}}; tree::T; end` and always returning it from `setup`. That would be breaking and might require a rename, but it would be less surprising.

mcabbott: > and always returning it from setup

This sounds like it moves the type instability due to `isempty(ties)` from `setup` which runs once into `update` which runs many times. And the slow part of `setup` isn't this, it's that it needs the cache to identify ties.

One thing we could do is have a fast `setup` path for `isbits` models (or things sufficiently close) which cannot have ties. But I'd like to have benchmarks before bolting on more stuff. 

> That would be breaking 

This package makes no guarantee about what type the thing returned by `setup` is, only that you must pass it to `update`. If, eventually, other tools develop which care, we can think again about whether such changes are breaking.

ToucheSir: I wish there were more defined conventions around the handling of input/return types as they pertain to SemVer, as even though 90% of packages make no guarantees about what return types will be, There's definitely some some implicit moderating effect that prevents them from shifting too much. Though the blast radius should be low, my gut feeling is one of unease here.

mcabbott: It also doesn't really matter if this package creeps up to 0.5 in the first few months of being useful. It doesn't have hundreds of dependents and they have clearly signed up for something young.

ToucheSir: I wish FluxML had a bit more social capital/good reputation when it comes to making breaking changes smoothly (and past history aside, this case is a good argument that we can, assuming it's breaking in the first place), but them's the breaks.

ToucheSir:
Anything you want to add here before we consider letting users have at it?

mcabbott:
It still copies in most of https://github.com/FluxML/Functors.jl/pull/33, here: https://github.com/FluxML/Optimisers.jl/pull/42/files#diff-b000beedbfab2d74910823607528b8c42d0ce41a9955e02a3e572e5d40a9762bR14-R36

src/interface.jl
```diff
@@ -2,22 +2,38 @@
...
+end
+
+function _setup(rule, x, addr; ties, cache)
+  usecache = !isbits(x) && cache !== false
```
mcabbott, line 21: At present this does not use the cache for isbits things, and allows `cache=false` to disable it completely. Ideally this would probably be harmonised with Functors.

ToucheSir:
A thought I had today was whether we couldn't obviate the need for `Tied` by making `Leaf` a mutable struct. Then tying two or more params could be represented by sharing the same leaf instance in the tree. This would not solve the composition problem, but it could reduce quite a bit of internal complexity and would work well with #49.

mcabbott:
> A thought I had today was whether we couldn't obviate the need for `Tied` by making Leaf a mutable struct. Then tying two or more params could be represented by sharing the same leaf instance in the tree. 

I'm not sure I see how this would work. Certainly `setup` could produce such a thing, with a cache alla Functors. But on the walk for `update!`, you need to know when you get to the first of a tied pair, so that you accumulate the two gradients and apply the rule once. To do this, `update!` would I think have to build something much like `Tied` as a first pass, before proceeding much like this.

ToucheSir:
The idea is that `update!` on a non-leaf runs two passes: first accumulate gradients into a cache, and then do the usual traversal with a visited set of leaves to avoid running rules twice. I tried writing some pseudocode for this, but it quickly metastasized into a full-blown branch. Will try to toss up a PR showing off the approach this week.

PR #43, opened 2022-01-30
Trivial cases of `OptimiserChain`

mcabbott:
Previously, using one or no arguments gave something which gave an error about iterating Descent. Now it should just work.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -549,3 +549,10 @@ function apply!(o::OptimiserChain, states, x, dx, dxs...)
 
   return new_states, dx
 end
+
+function Base.show(io::IO, c::OptimiserChain)
+  ioc = IOContext(io, :compact => true)
+  print(ioc, "OptimiserChain(")
+  join(ioc, c.opts, ", ")
+  print(io, ")")
+end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -516,13 +516,28 @@ end
 """
     OptimiserChain(opts...)
 
-Compose a chain (sequence) of optimisers so that each `opt` in `opts`
-updates the gradient in the order specified.
+Compose a sequence of optimisers so that each `opt` in `opts`
+updates the gradient, in the order specified.
+
+# Example
+```jldoctest
+julia> o = OptimiserChain(ClipGrad(1), Descent(0.1));
+
+julia> m = ([0,0,0],);
+
+julia> s = Optimisers.setup(o, m)
+(Leaf(OptimiserChain(ClipGrad{Int64}(1), Descent{Float64}(0.1)), [nothing, nothing]),)
+
+julia> Optimisers.update(s, m, ([0.3, 1, 7],))[2]  # clips before discounting
+([-0.03, -0.1, -0.1],)
+```
 """
-struct OptimiserChain{O}
+struct OptimiserChain{O<:Tuple}
   opts::O
 end
-OptimiserChain(opts...) = OptimiserChain(opts)
+OptimiserChain(o, opts...) = OptimiserChain((o, opts...))
+OptimiserChain(o) = o
+OptimiserChain() = Descent(true)
 
 init(o::OptimiserChain, x::AbstractArray) = [init(opt, x) for opt in o.opts]
 
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -519,6 +519,10 @@ end
 Compose a sequence of optimisers so that each `opt` in `opts`
 updates the gradient, in the order specified.
 
+With an empty sequence, `OptimiserChain()` is the identity,
+so `update!` will subtract the full gradient from the parameters.
+This is equivalent to `Descent(1)`.
+
 # Example
 ```jldoctest
 julia> o = OptimiserChain(ClipGrad(1), Descent(0.1));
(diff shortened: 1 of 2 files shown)


mcabbott:
#39 did not improve time-to-tests-pass...

Do we need 9 tests? This doesn't seem remotely OS-dependent, maybe just linux would be fine?

src/rules.jl
```diff
@@ -534,3 +549,10 @@ function apply!(o::OptimiserChain, states, x, dx, dxs...)
 
   return new_states, dx
 end
+
+function Base.show(io::IO, c::OptimiserChain)
+  ioc = IOContext(io, :compact => true)
+  print(ioc, "OptimiserChain(")
+  join(ioc, c.opts, ", ")
```
mcabbott, line 556: Not sure we want too many show methods; this one just hides the `{Tuple{ClipGrad{Int64}, Descent{Float64}}}` which makes the example ugly. The same information is printed just afterwards.

mcabbott:
```suggestion
function Base.show(io::IO, c::OptimiserChain)
  print(io, "OptimiserChain(")
  join(io, c.opts, ", ")
```

src/rules.jl
```diff
@@ -516,13 +516,28 @@ end
...
+```
 """
-struct OptimiserChain{O}
+struct OptimiserChain{O<:Tuple}
   opts::O
 end
-OptimiserChain(opts...) = OptimiserChain(opts)
+OptimiserChain(o, opts...) = OptimiserChain((o, opts...))
+OptimiserChain(o) = o
+OptimiserChain() = Descent(true)
```
ToucheSir, line 540: I would prefer erroring on an empty chain. A chain with no links is not much of a chain :)
```julia
struct OptimiserChain{O<:Tuple}
  opts::O
  function OptimiserChain(xs::Tuple)
    isempty(xs) && throw(ArgumentError("OptimiserChain cannot be empty"))
    return new{typeof(xs)}(xs)
  end
end
OptimiserChain(opts...) = OptimiserChain(opts)
```

ToucheSir (approved):

src/rules.jl
mcabbott: My thinking was that none of these are things you'd write by hand, but someone will splat a chain of operations in somehow, then no operations should mean it does nothing. And that seems easy to encode.

ToucheSir: Maybe if it acted like https://github.com/FluxML/Flux.jl/blob/master/src/layers/basic.jl#L46, but returning a different type seems too surprising to me.

mcabbott: But then your complaint starts with the 1-arg method. Here I point to this:
```
julia> ∘(sin, cos, tan)
sin ∘ cos ∘ tan

julia> ∘(sin)
sin (generic function with 15 methods)
```

ToucheSir: I guess it's a matter of mental models. If you look at `OptimiserChain` as a composition of multiple steps, then the logical conclusion is that it should work like `\circ`. If you focus on the `Chain` part (or see it as a pipe), then a single-segment chain is still a chain. Moreover, having the thing called Chain in Optimisers behave differently than the one defined in Flux is a potential source of confusion.

mcabbott: Mostly I want to ensure that any time you comment out one link of the chain, things still run. This can of course be achieved with different return types...

ToucheSir: Do have a stacktrace handy for https://github.com/FluxML/Optimisers.jl/pull/43#issue-1118374753? I'm trying to figure out why https://github.com/FluxML/Optimisers.jl/blob/master/src/rules.jl#L529-L536 doesn't work already for the 0- and 1-arg case.

mcabbott: The error was just because the constructor made a OptChain{Descent} instead of OptChain{Tuple{Descent}}, which can't be iterated.

ToucheSir: Thanks, looks like the `{O<:Tuple}` is sufficient to prevent that. Another question I just thought of: should `update!` with an empty OptimiserChain `subtract!`? I can see good arguments both ways.

mcabbott: That's a good question, I don't know. Since these things process the gradient in stages, I assumed that no stages meant no processing. But subtracting the complete gradient is... probably never what you want?

ToucheSir: I would assume no processing too, so if it doesn't blow the complexity budget then that may be the way to go.

mcabbott: Oh I misread that. An empty sequence is (in this PR) currently tested to mean no transformation of the gradient, a straight pipe to `subtract!`. Which is what falls out of the implementation, no steps in the `for` loop.

Performing no update seems more surprising to me. But perhaps you interpret these as making successive changes to the parameters? Clip seems an odd operation in that picture.

Implementation-wise it's one line to do whatever we please.

ToucheSir: I like the doc note and what's in the latest commit is definitely the path of least resistance, so let's stick with that. If someone complains in the future we can revisit.

ToucheSir (approved):

src/rules.jl
mcabbott: Done!

PR #44, opened 2022-01-30
Stop testing on windows + mac

mcabbott:
I don't think anything here is OS dependent. And 10 tests means they get stuck behind each other in the queue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,8 +20,6 @@ jobs:
           - 'nightly'
         os:
           - ubuntu-latest
-          - macOS-latest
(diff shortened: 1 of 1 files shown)


ToucheSir:
I'm generally ok with this. A compromise position would be just testing latest or LTS on Windows and OSX.

darsnack (approved):
Yeah Brian's suggestion is good, or we can just do this and turn it on again once the package stabilizes.

mcabbott:
Ok. If someone is better at GHA, perhaps it could test different versions on different OSes, which would get most of the benefit I presume.

Issue #38, opened 2022-01-28
Port over rule changes from Flux


ToucheSir:
- [x] https://github.com/FluxML/Flux.jl/pull/1776
- [x] https://github.com/FluxML/Flux.jl/pull/1299
- [ ] https://github.com/FluxML/Flux.jl/pull/1612 (not the `InvDecay` and `ExpDecay` parts)

darsnack:
Also, to be clear, `ExpDecay` and `InvDecay` should not be ported. These are not optimizers, and I think it's a mistake to shoehorn them into the same interface.

mcabbott:
This can be closed, 3rd bullet point closed by https://github.com/FluxML/Flux.jl/pull/1868 instead of here.

axsk:
> Also, to be clear, `ExpDecay` and `InvDecay` should not be ported. These are not optimizers, and I think it's a mistake to shoehorn them into the same interface.

How are these supposed to be used now?

ToucheSir:
Have you seen https://fluxml.ai/Flux.jl/stable/training/optimisers/#Scheduling-Optimisers? Basically, use ParameterSchedulers.jl. The Flux docs are actually just out of date, because we added functionality to ParameterSchedulers which makes this easier. http://fluxml.ai/ParameterSchedulers.jl/dev/tutorials/optimizers/ should cover everything.

axsk:
Sorry for crossposting since this probably doesn't belong here, but I don't know which package exactly is responsible.
I had seen ParameterSchedulers and tried to get it to work with the stateful optimisers, i.e. along the lines:
```julia
optstate = Flux.setup!(Optimisers.OptimisersChain(Optimisers.AdamW(), ParameterSchedulers.Exp(1,1)))
# so that i can use update!(optstate, model, grad[1])
```
I tried many combinations of `Flux` or `Optimisers` versions as well as with `ParameterSchedulers.Stateful`, but could not find a working version.

ToucheSir:
That code doesn't look right in many ways: `Flux.setup!` doesn't exist, and ParameterSchedulers schedules should not be used in `OptimisersChain`. Are you sure you've worked through the examples in http://fluxml.ai/ParameterSchedulers.jl/dev/tutorials/optimizers/ that I linked above? If you try them out and they aren't working, please open an issue with a minimal working example (MWE) on ParameterSchedulers.jl.

darsnack:
[...]

PR #46, opened 2022-01-30
Fix ADAMW, and track the loss

mcabbott:
This makes the same changes as https://github.com/FluxML/Flux.jl/pull/1612, xref #38.

It also alters the tests to keep track of the loss. Although it does not fix any of the convergence problems.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -392,8 +392,8 @@ end
 """
     ADAMW(η = 1f-3, β = (9f-1, 9.99f-1), γ = 0, ϵ = eps(typeof(η)))
 
-[ADAMW](https://arxiv.org/abs/1711.05101) is a variant of ADAM fixing (as in repairing) its
-weight decay regularization.
+[ADAMW](https://arxiv.org/abs/1711.05101) is a variant of [ADAM](@ref) fixing 
+(as in repairing) its weight decay regularization.
 
 # Parameters
 - Learning rate (`η`): Amount by which gradients are discounted before updating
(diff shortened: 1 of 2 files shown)


src/rules.jl
```diff
@@ -405,7 +405,7 @@ weight decay regularization.
                          (no need to change default)
 """
 ADAMW(η = 1f-3, β = (9f-1, 9.99f-1), γ = 0, ϵ = eps(typeof(η))) =
-  OptimiserChain(ADAM{typeof(η)}(η, β, ϵ), WeightDecay(γ))
+  OptimiserChain(ADAM{typeof(η)}(1, β, ϵ), WeightDecay{typeof(η)}(γ), Descent(η))
```
ToucheSir, line 408:
```suggestion
  OptimiserChain(ADAM{typeof(η)}(1, β, ϵ), WeightDecay{typeof(η)}(γ), Descent(η))
```
Sorry, I just remembered a previous comment (maybe on Slack? Can't find it now) which noted that our implementation did not match the [paper](https://arxiv.org/abs/1711.05101v3):
image.

TL;DR is that (possibly because of an odd choice of variable names in the paper, they use `α` instead of `η` for the learning rate and `η` for something else), the "fixed" version is wrong and the original is correct. Apologies for the noise.

CarloLucibello: oh damn, Brian is right, https://github.com/FluxML/Flux.jl/pull/1612 should be reverted, I got confused by the nomenclature

CarloLucibello: I'll file a PR in Flux

PR #47, opened 2022-01-30
Complex numbers alla Flux 1776

mcabbott:
This copies https://github.com/FluxML/Flux.jl/pull/1776 ~~but doesn't work. Why?~~

Xref #38

Edit: ~~Now this is on top of #48.~~ The original tests pass, but ADADelta only just passes, only the last few digits of the loss decrease. ~~Not sure if~~ That's true in Flux too. ~~Might be a bug somewhere?~~ Because the parameters chosen are strange ones.

loss

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -103,7 +103,7 @@ init(o::RMSProp, x::AbstractArray) = zero(x)
 function apply!(o::RMSProp, state, x, dx)
   η, ρ, ϵ, acc = o.eta, o.rho, o.epsilon, state
 
-  @.. acc = ρ * acc + (1 - ρ) * dx^2
+  @.. acc = ρ * acc + (1 - ρ) * abs2(dx)
   dx′ = @lazy dx * (η / (sqrt(acc) + ϵ))
   
   return acc, dx′
@@ -136,7 +136,7 @@ function apply!(o::ADAM, state, x, dx)
   mt, vt, βt = state
 
   @.. mt = β[1] * mt + (1 - β[1]) * dx
-  @.. vt = β[2] * vt + (1 - β[2]) * dx ^ 2
+  @.. vt = β[2] * vt + (1 - β[2]) * abs2(dx)
   dx′ = @lazy mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ) * η
 
   return (mt, vt, βt .* β), dx′
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -244,7 +244,7 @@ function apply!(o::OADAM, state, x, dx)
   mt, vt, βt, term = state
 
   @.. mt = β[1] * mt + (1 - β[1]) * dx
-  @.. vt = β[2] * vt + (1 - β[2]) * dx^2
+  @.. vt = β[2] * vt + (1 - β[2]) * abs2(dx)
   prev = copy(term)
   @.. term = η * mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)
   dx′ = @lazy 2 * term - prev
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -171,7 +171,7 @@ function apply!(o::RADAM, state, x, dx)
   mt, vt, βt, t = state
 
   @.. mt = β[1] * mt + (1 - β[1]) * dx
-  @.. vt = β[2] * vt + (1 - β[2]) * dx^2
+  @.. vt = β[2] * vt + (1 - β[2]) * abs2(dx)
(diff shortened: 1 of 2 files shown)


src/rules.jl
```diff
@@ -134,7 +134,7 @@ function apply!(o::ADAM{T}, state, x, dx) where T
   mt, vt, βt = state
 
   mt′ = @.. mt = β[1] * mt + (one(T) - β[1]) * dx
-  vt′ = @.. vt = β[2] * vt + (one(T) - β[2]) * dx ^ 2
+  vt′ = @.. vt = β[2] * vt + (one(T) - β[2]) * abs2(dx)
   dx′ = @.. mt / (one(T) - βt[1]) / (sqrt(vt / (one(T) - βt[2])) + ϵ) * η
 
   return (mt′, vt′, βt .* β), dx′
```
mcabbott, line 140: Possibly related: `vt′ = @.. vt = β[2] * vt + ...` mutates `vt` if possible. But using `vt` on the next line isn't safe, it might be the old one. 

My fault, but maybe we should change the magic macro, so that `@.. vt = β[2] * vt + ...` either mutates `vt` or else re-assigns the same name to the new value. The present state seems like a recipe for such bugs. 

I doubt this is causing the test failure here, since it happens with Array.

ToucheSir: That sounds reasonable to me, can it be pulled off without too much magic?

mcabbott: Yes it's an easy change. The hardest part will be the resulting rebase of every line this PR...

ToucheSir: How about a quick follow-up after this is merged? We're in no rush and that  can keep this PR focused.

mcabbott: I made #48. Partial laziness is... sometimes messy.

src/rules.jl
```diff
@@ -307,13 +306,13 @@ function apply!(o::ADADelta, state, x, dx)
...
+  @.. acc = ρ * acc + (1 - ρ) * abs2(dx)
   # DON'T remove epsilon from numerator
   # or even out of the square roots
   dx′ = @.. dx * sqrt(Δacc + ϵ) / sqrt(acc + ϵ)
-  Δacc′ = @.. Δacc = ρ * Δacc + (1 - ρ) * dx^2
+  @.. Δacc = ρ * Δacc + (1 - ρ) * abs2(dx′)
   
-  return (acc′, Δacc′), dx′
+  return (acc, Δacc), dx′
 end
```
mcabbott, line 316: Source is: 
Screenshot 2022-01-30 at 15 33 58

cossio:
What's the loss function you are using in the plot?

mcabbott:
It's the one from the tests, here https://github.com/FluxML/Optimisers.jl/pull/47/files#diff-9f183063573ccf8f01dd29629bad9986f3faa3ce2fdeeb3f0855eaa1f6450715R214 and the plot commands are in comments there. The same loss is used in the Flux PR's tests.

test/rules.jl
```diff
@@ -166,3 +201,70 @@ end
   end
 end
 
+@testset "with complex numebers: Flux#1776" begin
+  empty!(LOG)
+  @testset "$(name(opt))" for opt in [
+              # The Flux PR had 1e-2 for all. But ADADelta(ρ) needs ρ≈0.9 not small. And it helps to make ε not too small too:
+              ADAM(1e-2), RMSProp(1e-2), RADAM(1e-2), OADAM(1e-2), ADAGrad(1e-2), ADADelta(0.9, 1e-5), NADAM(1e-2), AdaBelief(1e-2),
```
mcabbott, line 208: This is the fix, BTW.

mcabbott:
Good to go?

test/rules.jl
```diff
@@ -19,7 +19,16 @@ RULES = [
 name(o) = typeof(o).name.name
 name(o::OptimiserChain) = join(name.(o.opts), " → ")
 
+LOG = Dict()
+
+loggradient(o) = (f, xs...) -> begin
+  y, dxs = Zygote.withgradient(f, xs...)
+  push!(get!(() -> Float32[], LOG, name(o)), y)
+  dxs  # save the loss, return the gradient
+end
```
ToucheSir, line 28: This (`LOG`) doesn't seem to be in use anywhere, is it still necessary?

mcabbott: It just makes debugging easier if you can plot things form the tests you just ran. It's not strictly necessary but also doesn't really get in the way, I think.

ToucheSir: Fair enough. A comment with what you just described would help then.

ToucheSir (approved):

PR #48, opened 2022-01-30
Separate `@lazy` from `@..` macro, fix some bugs

mcabbott:
This changes the behaviour of `@.. x = rhs` on immutable objects, to attach the label to the new result. While this may mean that the type of `x` changes, it also fixes several bugs where later lines assumed that `x` would be the new one. Should have noticed this on the first pass, sorry. (This was added in #31.)

There are more subtle ways for things to go wrong. Flux's implementation of OADAM uses the mutable gradient array as the place to save an intermediate step: 
https://github.com/FluxML/Flux.jl/blob/05a608b3fb540661bf68c339793cd7083e2ad400/src/optimise/optimisers.jl#L311-L326
That won't work if `Δ == dx` is immutable, but attaching the label to the RHS also won't work, if that gets mutated in the next step. And in general mutating the gradient is unsafe, if could also be someone else's gradient. ~~So for now I make it explicitly materialise the new `dx`, and then update the saved term. Broadcasting a `sqrt` is not super-expensive.~~ Changed now to an explicit copy of `term`, before updating.

Screenshot 2022-01-30 at 13 56 53

In fact, the whole lazy-Broadcasted idea assumes that the function being broadcasted isn't expensive. E.g. `ADADelta` doesn't have correctness bugs, but will compute the sqrt etc. when doing `Δacc =`, and again when `dx′` is ultimately consumed in `subtract!`.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -111,6 +111,9 @@ Broadcast.broadcasted(::typeof(lazy), x) = Lazy(x)
 struct Lazy{T}; bc::T; end
 Broadcast.materialize(x::Lazy) = Broadcast.instantiate(x.bc)
 
+constantlike(λ::T, x::AbstractArray{T}) where T = map(_ -> λ, x)
+constantlike(λ, x::AbstractArray{T}) where T = constantlike(convert(float(T), λ), x)
+
 function Base.show(io::IO, ℓ::Leaf)  # show method is mostly to hide its long type!
   ioc = IOContext(io, :compact => true)
   print(ioc, "Leaf(", ℓ.rule, ", ")
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -78,29 +78,32 @@ end
 
 """
     @.. x = x + y
-    @.. x + y / z
 
-Magic broadcasting macro, for use in `apply!` rules:
-* Applied to assignment `x = rhs` it is like `@.` unless `!iswriteable(x)`,
-  in which case it becomes `x = @. rhs`. Does not check the eltype before writing.
-  Does not know about `x += rhs` etc.
-* Applied to other expressions, it broadcasts like `@.` but does not materialise,
-  returning a `Broadcasted` object for later use. Beware that mutation of arguments
-  will affect the result, and that if it is used in two places work will be done twice.
+Sometimes in-place broadcasting macro, for use in `apply!` rules.
+If `iswriteable(x)` then it is just `@. x = rhs`, but if not, it becomes `x = @. rhs`.
 """
 macro var".."(ex)
-  if Meta.isexpr(ex, :(=))
-    dst = esc(ex.args[1])
-    src = esc(Broadcast.__dot__(ex.args[2]))
-    :($dst = if $iswriteable($dst)
-        $dst .= $src
-      else
-        $src
-      end)
-  else
-    bc = esc(Broadcast.__dot__(ex))
-    :($lazy.($bc))
-  end
+  Meta.isexpr(ex, :(=)) || throw("the macro @.. only accepts assignment, like @.. x = y + z")
+  dst = esc(ex.args[1])
+  src = esc(Broadcast.__dot__(ex.args[2]))
+  :($dst = if $iswriteable($dst)
+      $dst .= $src
+    else
+      $src
+    end)
+end
+
+"""
+    x = @lazy y + z
+
+Lazy broadcasting macro, for use in `apply!` rules. It broadcasts like `@.`
+but does not materialise, returning a `Broadcasted` object for later use.
+Beware that mutation of arguments will affect the result,
+and that if it is used in two places, work will be done twice.
+"""
+macro lazy(ex)
+  bc = esc(Broadcast.__dot__(ex))
+  :($lazy.($bc))
 end
 
 function lazy end
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -111,8 +111,8 @@ Broadcast.broadcasted(::typeof(lazy), x) = Lazy(x)
 struct Lazy{T}; bc::T; end
 Broadcast.materialize(x::Lazy) = Broadcast.instantiate(x.bc)
 
-constantlike(λ::T, x::AbstractArray{T}) where T = map(_ -> λ, x)
-constantlike(λ, x::AbstractArray{T}) where T = constantlike(convert(float(T), λ), x)
+onevalue(λ::T, x::AbstractArray{T}) where T = map(_ -> λ, x)
+onevalue(λ, x::AbstractArray{T}) where T = onevalue(convert(float(T), λ), x)
 
 function Base.show(io::IO, ℓ::Leaf)  # show method is mostly to hide its long type!
   ioc = IOContext(io, :compact => true)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -81,8 +81,8 @@ end
     @.. x + y / z
 
 Magic broadcasting macro, for use in `apply!` rules:
-* Applied to assignment `x = ...` it is like `@.` unless `!iswriteable(x)`,
-  in which case it ignores `x`, and applies `@.` on the right.
+* Applied to assignment `x = rhs` it is like `@.` unless `!iswriteable(x)`,
+  in which case it becomes `x = @. rhs`.
 * Applied to other expressions, it broadcasts like `@.` but does not materialise,
   returning a `Broadcasted` object for later use.
 """
@@ -90,7 +90,7 @@ macro var".."(ex)
   if Meta.isexpr(ex, :(=))
     dst = esc(ex.args[1])
     src = esc(Broadcast.__dot__(ex.args[2]))
-    :(if $iswriteable($dst)
+    :($dst = if $iswriteable($dst)
         $dst .= $src
       else
         $src
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -82,9 +82,11 @@ end
 
 Magic broadcasting macro, for use in `apply!` rules:
 * Applied to assignment `x = rhs` it is like `@.` unless `!iswriteable(x)`,
-  in which case it becomes `x = @. rhs`.
+  in which case it becomes `x = @. rhs`. Does not check the eltype before writing.
+  Does not know about `x += rhs` etc.
 * Applied to other expressions, it broadcasts like `@.` but does not materialise,
-  returning a `Broadcasted` object for later use.
+  returning a `Broadcasted` object for later use. Beware that mutation of arguments
+  will affect the result, and that if it is used in two places work will be done twice.
 """
 macro var".."(ex)
   if Meta.isexpr(ex, :(=))
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -270,7 +270,7 @@ struct ADAGrad{T}
 end
 ADAGrad(η = 1f-1, ϵ = eps(typeof(η))) = ADAGrad{typeof(η)}(η, ϵ)
 
-init(o::ADAGrad, x::AbstractArray) = fill!(similar(x), o.epsilon)
+init(o::ADAGrad, x::AbstractArray) = map(_ -> o.epsilon, x)
 
 function apply!(o::ADAGrad, state, x, dx)
   η, ϵ = o.eta, o.epsilon
@@ -337,7 +337,7 @@ end
 AMSGrad(η = 1f-3, β = (9f-1, 9.99f-1), ϵ = eps(typeof(η))) = AMSGrad{typeof(η)}(η, β, ϵ)
 
 init(o::AMSGrad, x::AbstractArray) =
-  (fill!(similar(x), o.epsilon), fill!(similar(x), o.epsilon), fill!(similar(x), o.epsilon))
+  (map(_ -> o.epsilon, x), map(_ -> o.epsilon, x), map(_ -> o.epsilon, x))
 
 function apply!(o::AMSGrad, state, x, dx)
   η, β, ϵ = o.eta, o.beta, o.epsilon
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -246,7 +246,7 @@ function apply!(o::OADAM, state, x, dx)
   @.. vt = β[2] * vt + (1 - β[2]) * dx^2
   prev = copy(term)
   @.. term = η * mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)
-  dx′ = @. 2 * term - prev
+  dx′ = @.. 2 * term - prev
 
   return (mt, vt, βt .* β, term), dx′
 end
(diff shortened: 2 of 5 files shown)


src/rules.jl
```diff
@@ -271,16 +270,16 @@ struct ADAGrad{T}
 end
 ADAGrad(η = 1f-1, ϵ = eps(typeof(η))) = ADAGrad{typeof(η)}(η, ϵ)
 
-init(o::ADAGrad, x::AbstractArray) = fill!(similar(x), o.epsilon)
+init(o::ADAGrad, x::AbstractArray) = map(_ -> o.epsilon, x)
```
mcabbott, line 273: RFC, but 6f65a28 replaces similar with map. `similar` always makes a mutable container; this should instead make one of a type close to `x`.

src/rules.jl
```diff
@@ -337,7 +337,7 @@ end
 AMSGrad(η = 1f-3, β = (9f-1, 9.99f-1), ϵ = eps(typeof(η))) = AMSGrad{typeof(η)}(η, β, ϵ)
 
 init(o::AMSGrad, x::AbstractArray) =
-  (fill!(similar(x), o.epsilon), fill!(similar(x), o.epsilon), fill!(similar(x), o.epsilon))
+  (map(_ -> o.epsilon, x), map(_ -> o.epsilon, x), map(_ -> o.epsilon, x))
```
ToucheSir, line 340: Part of me wishes we could import Compat.jl and use `map(Returns(o.episilon), x)` instead, but ah well.

mcabbott: This way does save 4 characters...

ToucheSir: The explicit wrapper is more likely to have optimizations behind the scenes and may be kinder to the compiler, but it's true that the impact here would be minimal (unless maybe it can prove `map` returns an immutable value and can run CSE?)

src/interface.jl
```diff
@@ -81,16 +81,16 @@ end
...
 * Applied to other expressions, it broadcasts like `@.` but does not materialise,
   returning a `Broadcasted` object for later use.
 """
 macro var".."(ex)
   if Meta.isexpr(ex, :(=))
     dst = esc(ex.args[1])
     src = esc(Broadcast.__dot__(ex.args[2]))
-    :(if $iswriteable($dst)
+    :($dst = if $iswriteable($dst)
         $dst .= $src
```
mcabbott, line 94: Thinking about complex numbers, this could check that the eltype of `src` fits into `dst`. Not sure how far we want to go making this robust, maybe we wait until it bites someone -- it will not be silently wrong.

src/rules.jl
ToucheSir: On second thought, this removes the possibility of in-place accumulation of state if the parameter and gradients are always immutable. Maybe that's fine on balance, the wildcard here is how ImmutableArrays will behave.

ToucheSir: A similar question already exists for any `init` that currently uses `zero`.

mcabbott: This makes it match `zero`, and should I think always be mutable/not when the model is. That seems like the most straightforward thing, there's just one knob.

ToucheSir:
On laziness: perhaps `@..` should only handle the `@.. x = ...` case and always materialize. Then laziness could be an explicit opt-in with `lazy.(@. ...)`&mdash;not any more verbose and reduces the trickiness involved.

On ADADelta: is there any reason `dx` can't be materialized into directly instead of creating a lazy `dx'`? Seems like this would be worth the extra kernel dispatch + memory usage given it's re-used during `subtract!`.

src/rules.jl
ToucheSir: Or always mutable when possible, but it ought to be consistent and it current isn't. There is merit to the argument that `similar` is more likely to create a dense array unnecessarily.

src/rules.jl
ToucheSir: A potential future optimization would be making these `Fills`.

mcabbott:
I do think we want laziness 99% of the time, so prefer making this compact. It's what lets ClipGrad etc. compose almost for free. 

OADAM actually has a typo which makes it materialise the gradient. It should have `dx′ = @.. 2 * term - prev`. Fusing this 2* and - with `subtract!` seems like what we want, they are free, but a copy is not.

Re ADADelta ... it might be worth some attempt to benchmark this one in isolation?

src/rules.jl
mcabbott: Maybe, but they are only made once, and then perhaps it compiles two specialisations? Also I'm not sure how reliably operations which mix Fill and CuArray stay GPU side.

It's possible we should add a knob to make state mutable when the model is not. It's often more memory than the model itself, right? But for now doing one thing for all optimisers seems cleanest.

ToucheSir:
Would a different macro or even a short alias for `lazy` help there? If the contention is that having `@.. x = ...` and `@.. ...` return different types and observe different behaviour wrt laziness and materialization can cause subtle issues, perhaps that's a sign that they should not have the same name.

mcabbott:
ADADelta times:
```julia
julia> using Optimisers, BenchmarkTools
julia> m, g = ntuple(_ -> rand(Float32, 100,100), 10), ntuple(_ -> rand(Float32, 100,100), 10);
julia> s = Optimisers.setup(ADADelta(), m);

julia> @btime Optimisers.update!($s, $m, $g);
  min 414.583 μs, mean 419.602 μs (0 allocations)  # lazy
  min 334.500 μs, mean 381.109 μs (20 allocations, 391.09 KiB)  # eager

  min 377.166 μs, mean 379.411 μs (0 allocations)  # lazy + @fastmath
  min 180.708 μs, mean 214.607 μs (20 allocations, 391.09 KiB)  # eager + @fastmath

julia> s = Optimisers.setup(ADAM(), m);
julia> @btime Optimisers.update!($s, $m, $g);
  min 265.583 μs, mean 268.664 μs (20 allocations, 320 bytes)

julia> s = Optimisers.setup(Descent(), m);
julia> @btime Optimisers.update!($s, $m, $g);
  min 36.708 μs, mean 37.492 μs (0 allocations)

 julia> m2 = ntuple(_ -> ntuple(_ -> rand(Float32, 1000,1000), 20), 20);
julia> g2 = ntuple(_ -> ntuple(_ -> rand(Float32, 1000,1000), 20), 20);
julia> s2 = Optimisers.setup(ADADelta(), m2);

julia> @btime Optimisers.update!($s2, $m2, $g2);
  min 3.710 s, mean 3.710 s (0 allocations)
  min 3.429 s, mean 3.429 s (800 allocations, 1.49 GiB)
```
On an M1 laptop. I guess checking with CuArrays would be better. 

Re bugs, I think the problem was mostly me adding \prime too many places mechanically without thinking. That I am sure I could have managed even with long descriptive macro names. I will take one more read through this to double-check.

ToucheSir:
All numbers are post warmup.

Lazy:
```julia
julia> CUDA.@sync @btime Optimisers.update!($s, $m, $g);
  215.351 μs (2080 allocations: 267.03 KiB)

julia> let s=s, m=m, g=g; CUDA.@time Optimisers.update!(s, m, g); end;
  0.000329 seconds (2.08 k CPU allocations: 267.359 KiB)
```
Eager:
```julia
julia> CUDA.@sync @btime Optimisers.update!($s, $m, $g);
  184.293 μs (1770 allocations: 166.41 KiB)

julia> let s=s, m=m, g=g; CUDA.@time Optimisers.update!(s, m, g); end;
  0.000280 seconds (1.77 k CPU allocations: 166.734 KiB)
```
Big lazy:
```julia
julia> CUDA.@sync @btime Optimisers.update!($s2, $m2, $g2);
  10.340 ms (83200 allocations: 10.43 MiB)

julia> let s=s2, m=m2, g=g2; CUDA.@time Optimisers.update!(s, m, g); end;
  0.055469 seconds (83.22 k CPU allocations: 10.444 MiB)
```
Big eager:
```julia
julia> CUDA.@sync @btime Optimisers.update!($s2, $m2, $g2);
  21.282 ms (70800 allocations: 6.50 MiB)

julia> let s=s2, m=m2, g=g2; CUDA.@time Optimisers.update!(s, m, g); end;
  0.062318 seconds (70.81 k CPU allocations: 6.513 MiB, 5.08% gc time)
```

mcabbott:
Sorry, I timed some things, & then deleted them as I think the lazy one gets the wrong answer. It captures Δacc which is mutated on the next line. Maybe Nesterov is wrong too.

Besides proofreading this probably wants bolder notation. I think also some of the `dx′` are pretty subtle, about which terms get the old and the new, I wonder if there too we ought to have something bolder.

ToucheSir:
I didn't catch that despite multiple readthroughs as well. How about we stick with eager `@..` for now until to let the design of the lazy path bake? Users who care about memory usage will be using `writeable` arrays anyhow, so it shouldn't create significant allocation overhead.

mcabbott:
Here's one test we should have had:
```
@testset verbose=true "mutation check" begin
  @testset "$(name(o))" for o in RULES
    model = Float64.(rand(Int8, 8))
    s_model = SVector{8}(model)
    grads = [Float64.(rand(Int8, 8)) for t in 1:13]
    s_grads = [SVector{8}(x) for x in grads]
    state = Optimisers.setup(o, model)
    s_state = Optimisers.setup(o, s_model)
    for t in 1:13
      state, model = Optimisers.update!(state, model, grads[t])
      s_state, s_model = Optimisers.update!(s_state, s_model, s_grads[t])
    end
    @test model == s_model 
  end
end
```
This would catch the laziness bug, since mutation of an array you have captured doesn't affect the SArray path, and it does in practice. (In the 2nd digit, not the 16th.) It fails for ADAGrad and AMSGrad, at Float32 resolution, because for these two `map(_ -> o.epsilon, x)` makes an array of Float32, not 64.

Can this miss other such bugs?

src/rules.jl
```diff
@@ -41,10 +41,10 @@ Momentum(η = 1f-2, ρ = 9f-1) = Momentum{typeof(η)}(η, ρ)
 init(o::Momentum, x::AbstractArray) = zero(x)
 
 function apply!(o::Momentum, state, x, dx)
-  η, ρ, v = o.eta, o.rho, state
-  v′ = @.. v = ρ * v - η * dx
+  η, ρ, mv = o.eta, o.rho, state
+  @.. mv = - ρ * mv + η * dx  # Macro @.. broadcasts into v if it can, else @. of rhs.
   
-  return v′, @.. -v′
+  return mv, mv
```
mcabbott, line 47: `mv` is `-v`. By storing this instead, there is no need for a second broadcast of the output.

src/rules.jl
```diff
@@ -133,11 +135,11 @@ function apply!(o::ADAM{T}, state, x, dx) where T
   η, β, ϵ = o.eta, o.beta, o.epsilon
   mt, vt, βt = state
 
-  mt′ = @.. mt = β[1] * mt + (one(T) - β[1]) * dx
-  vt′ = @.. vt = β[2] * vt + (one(T) - β[2]) * dx ^ 2
```
mcabbott, line 137: No need for `one(T)`. The struct constrains `β[2]` to have type T, and 1 will promote to that. As in other rules.

src/rules.jl
```diff
@@ -271,16 +271,16 @@ struct ADAGrad{T}
...
 function apply!(o::ADAGrad, state, x, dx)
   η, ϵ = o.eta, o.epsilon
   acc = state
 
-  acc′ = @.. acc = acc + dx^2
-  dx′ = @.. dx * η / (sqrt(acc) + ϵ)
+  @.. acc = acc + dx^2
+  dx′ = @lazy dx * η / (sqrt(acc) + ϵ)
 
-  return acc′, dx′
```
mcabbott, line 283: Another place my \primes went wrong. At least comparing to Flux, `sqrt(acc)` should be the updated one.

Re notation, `@..` seems non-dangerous now. If you write `@.` you may get an error that you can't write into an immutable array. If you apply it to something which isnt `@.. x = ...` you will get a load error.

mcabbott:
Here's a check that results match Flux exactly: https://gist.github.com/mcabbott/886756eb8a8ee3c8cedd5173ded13aa3

src/interface.jl
```diff
@@ -111,6 +111,9 @@ Broadcast.broadcasted(::typeof(lazy), x) = Lazy(x)
 struct Lazy{T}; bc::T; end
 Broadcast.materialize(x::Lazy) = Broadcast.instantiate(x.bc)
 
+constantlike(λ::T, x::AbstractArray{T}) where T = map(_ -> λ, x)
```
ToucheSir, line 114: thoughts on `filllike` to save a few chars? Too many `L`s?

mcabbott: `withfill(λ, x)`? `fillsame`?

mcabbott: How about `onevalue(λ, x)`?

src/rules.jl
```diff
@@ -67,11 +67,12 @@ Nesterov(η = 1f-3, ρ = 9f-1) = Nesterov{typeof(η)}(η, ρ)
...
-  η, ρ, v = o.eta, o.rho, state
-  d = @.. ρ^2 * v - (1+ρ) * η * dx
-  v′ = @.. v = ρ * v - η * dx
+  η, ρ, vel = o.eta, o.rho, state
+
+  md = @. - ρ^2 * vel + (1+ρ) * η * dx  # Cannot be lazy as this needs the old velocity
+  @.. vel = ρ * vel - η * dx
   
-  return v′, @.. -d
+  return vel, md
```
mcabbott, line 75: BTW, I re-named velocity as `vel` so that single-letter variables are all scalars. `md` is minus-d (compared to Flux) because writing it that way saves the final broadcast.

src/rules.jl
```diff
@@ -41,10 +41,10 @@ Momentum(η = 1f-2, ρ = 9f-1) = Momentum{typeof(η)}(η, ρ)
 init(o::Momentum, x::AbstractArray) = zero(x)
 
 function apply!(o::Momentum, state, x, dx)
-  η, ρ, v = o.eta, o.rho, state
-  v′ = @.. v = ρ * v - η * dx
+  η, ρ, mvel = o.eta, o.rho, state
+  @.. mvel = ρ * mvel + η * dx  # Macro @.. broadcasts into mvel if it can, else @. of rhs.
   
-  return v′, @.. -v′
+  return mvel, mvel
```
mcabbott, line 47: Likewise here, `mvel` is minus-velocity, to avoid the final broadcast. (In Flux this was free, `@. Δ = -v`.)

mcabbott (commented):
Ready to go?

src/rules.jl
ToucheSir: Since `md` has a negative coefficient, how about `negd`?

mcabbott: Or maybe just `newdx`? I'm not sure whether `d` means anything outside of Flux.

mcabbott: I am a bit tempted to delete all the `dx′` primes in fact. Maybe they are just clutter.

ToucheSir: `newdx` sounds good, it clearly communicates the purpose of the value and makes it easier to tell what's being returned on a quick glance.

mcabbott: Done.

ToucheSir (approved):

PR #51, opened 2022-02-05
Centred RMSProp

mcabbott:
Parallel to https://github.com/FluxML/Flux.jl/pull/1778. 

But if "Parameters other than learning rate generally don't need tuning", then having to type them out to get to the boolean one seems awkward. Cleaner to call it a new optimiser?

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -84,7 +84,7 @@ function apply!(o::Nesterov, state, x, dx)
 end
 
 """
-    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centre = false)
+    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred = false)
 
 Optimizer using the
 [RMSProp](https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf)
@@ -101,26 +101,26 @@ gradients by an estimate their variance, instead of their second moment.
                   prominent direction, in effect dampening oscillations.
 - Machine epsilon (`ϵ`): Constant to prevent division by zero
                          (no need to change default)
-- Keyword `centre` (or `center`): Indicates whether to use centred variant
-                                  of the algorithm.
+- Keyword `centred` (or `centered`): Indicates whether to use centred variant
+                                     of the algorithm.
 """
 struct RMSProp{T}
   eta::T
   rho::T
   epsilon::T
-  centre::Bool
+  centred::Bool
 end
-RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centre = false, center = false) =
-  RMSProp{typeof(η)}(η, ρ, ϵ, centre | center)
+RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred::Bool = false, centered::Bool = false) =
+  RMSProp{typeof(η)}(η, ρ, ϵ, centred | centered)
 
-init(o::RMSProp, x::AbstractArray) = (zero(x), o.centre ? zero(x) : false)
+init(o::RMSProp, x::AbstractArray) = (zero(x), o.centred ? zero(x) : false)
 
 function apply!(o::RMSProp, state, x, dx)
   η, ρ, ϵ = o.eta, o.rho, o.epsilon
   quad, lin = state
 
   @.. quad = ρ * quad + (1 - ρ) * abs2(dx)
-  if o.centre
+  if o.centred
     @.. lin = ρ * lin + (1 - ρ) * dx
   end
   dx′ = @lazy dx * η / (sqrt(quad - abs2(lin)) + ϵ)
@@ -132,7 +132,7 @@ function Base.show(io::IO, o::RMSProp)
     show(io, typeof(o))
     print(io, "(")
     join(io, [o.eta, o.rho, o.epsilon], ", ")
-    print(io, "; centre = ", o.centre, ")")
+    print(io, "; centred = ", o.centred, ")")
 end
 
 """
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -84,7 +84,7 @@ function apply!(o::Nesterov, state, x, dx)
 end
 
 """
-    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)))
+    RMSProp(η = 1f-3, ρ = 9f-1, c::Bool, ϵ = eps(typeof(η)))
 
 Optimizer using the
 [RMSProp](https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf)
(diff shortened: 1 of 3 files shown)


ToucheSir:
I have a bit of a hard time believing the no tuning required part...we could always make a kwarg constructor if it gets annoying.

darsnack (approved):

mcabbott:
Maybe a keyword is best, it's different from the others and 4 positional arguments is a lot. Probably https://github.com/FluxML/Flux.jl/pull/1778 should match this.

cossio:
Why `centre` instead of `centered`?

https://github.com/FluxML/Flux.jl/pull/1778 is using `centered`. PyTorch too, https://pytorch.org/docs/stable/generated/torch.optim.RMSprop.html.

src/rules.jl
```diff
@@ -76,7 +76,7 @@ function apply!(o::Nesterov, state, x, dx)
 end
 
 """
-    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)))
+    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centre = false)
```
cossio, line 79:
```suggestion
    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred = false)
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
   eta::T
   rho::T
   epsilon::T
+  centre::Bool
 end
-RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η))) = RMSProp{typeof(η)}(η, ρ, ϵ)
+RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centre = false, center = false) =
+  RMSProp{typeof(η)}(η, ρ, ϵ, centre | center)
```
cossio, line 101:
```suggestion
RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred = false, centered = false) =
  RMSProp{typeof(η)}(η, ρ, ϵ, centred | centered)
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
...
 
-init(o::RMSProp, x::AbstractArray) = zero(x)
+init(o::RMSProp, x::AbstractArray) = (zero(x), o.centre ? zero(x) : false)
 
 function apply!(o::RMSProp, state, x, dx)
-  η, ρ, ϵ, acc = o.eta, o.rho, o.epsilon, state
+  η, ρ, ϵ = o.eta, o.rho, o.epsilon
+  quad, lin = state
```
cossio, line 107:
```suggestion
  st, = state
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
...
 
-  @.. acc = ρ * acc + (1 - ρ) * abs2(dx)
-  dx′ = @lazy dx * (η / (sqrt(acc) + ϵ))
+  @.. quad = ρ * quad + (1 - ρ) * abs2(dx)
+  if o.centre
+    @.. lin = ρ * lin + (1 - ρ) * dx
+  end
```
cossio, line 112:
```suggestion
  if o.centre
    @.. mt = ρ * mt + (1 - ρ) * dx
  end
  @.. st = ρ * st + (1 - ρ) * abs2(dx - mt)
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
...
+  dx′ = @lazy dx * η / (sqrt(quad - abs2(lin)) + ϵ)
```
cossio, line 113:
```suggestion
  dx′ = @lazy dx * η / (sqrt(st + ϵ) + ϵ)
```

cossio:
> Besides matters of taste on naming things, this scatter of suggestions makes some logic changes. Can you please write these (and these alone) clearly in one place with an explanation of what & why? What source is this following? Etc.

Ok I agree, that's why I suggested a separate PR. The naming is of course not essential, but the logic changes amounts to just that now it's carrying an estimate of the variance of the gradient directly instead of the second moment and then subtracting the squared mean. As far as I understand that's how it's implemented by Jax, https://github.com/deepmind/optax/blob/b4aa6657bbf79985279dea76eaf6d53b25d7e8d9/optax/_src/transform.py#L247.

I can make a separate PR since I think that'll make it easier to compare.

mcabbott:
No, one multiple PRs won't make things clearer, but writing 3 lines of formula in one place with a clear description might, for when someone has bandwidth to look closely. Is there a paper we can follow with clear formulas?

cossio:
> Is there a paper we can follow with clear formulas?

I don't think so. Centered RMSprop was introduced by http://arxiv.org/abs/1308.0850 without discussing details and without giving the implementation. There he gives the formulas:

image

which coincide with the current status of this PR (notation: epsilon = minibatch gradient, n_i = gradient second moment, g_i gradient first moment).

However in Jax they implement it by estimating the variance of the gradient directly.

This is the same AdaBelief does:

image

So what I was proposing amounts to setting beta1 = 0 in this AdaBelief pseudo-code. You also need to put epsilon inside the square root because the variance can get numerically negative.

Sorry for the noise with all the scattered suggestions.

However I'm not sure if this is actually better in practice.

cossio:
Actually, looking at Optax code more closely, I now think they do this difference of squares thing instead of estimating the variance directly. See https://github.com/deepmind/optax/blob/a124552d0fc9f81812cd82da0d22528b7a17a847/optax/_src/transform.py#L247.

So then that's probably the way to go here too.  I have removed my previous suggestions.

src/rules.jl
cossio, line 79:
```suggestion
    RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred = false)
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
   eta::T
   rho::T
   epsilon::T
+  centre::Bool
```
cossio, line 98:
```suggestion
  centred::Bool
```

src/rules.jl
cossio, line 101:
```suggestion
RMSProp(η = 1f-3, ρ = 9f-1, ϵ = eps(typeof(η)); centred = false, centered = false) =
  RMSProp{typeof(η)}(η, ρ, ϵ, centred | centered)
```

src/rules.jl
```diff
@@ -95,18 +95,24 @@ struct RMSProp{T}
...
 
 function apply!(o::RMSProp, state, x, dx)
-  η, ρ, ϵ, acc = o.eta, o.rho, o.epsilon, state
+  η, ρ, ϵ = o.eta, o.rho, o.epsilon
+  quad, lin = state
 
-  @.. acc = ρ * acc + (1 - ρ) * abs2(dx)
-  dx′ = @lazy dx * (η / (sqrt(acc) + ϵ))
+  @.. quad = ρ * quad + (1 - ρ) * abs2(dx)
+  if o.centre
```
cossio, line 110:
```suggestion
  if o.centred
```

test/rules.jl
```diff
@@ -14,10 +14,13 @@ RULES = [
   OptimiserChain(ClipNorm(), ADAM(0.001)),
   OptimiserChain(ClipGrad(0.5), Momentum()),
   OptimiserChain(WeightDecay(), OADAM(), ClipGrad(1)),
+  # Not the default:
+  RMSProp(centre = true),
```
cossio, line 18:
```suggestion
  RMSProp(centred = true),
```

test/rules.jl
```diff
@@ -14,10 +14,13 @@ RULES = [
...
 ]
 
 name(o) = typeof(o).name.name  # just for printing testset headings
 name(o::OptimiserChain) = join(name.(o.opts), " → ")
+name(o::RMSProp) = o.centre ? "RMSProp(centre = true)" : :RMSProp
```
cossio, line 23:
```suggestion
name(o::RMSProp) = o.centred ? "RMSProp(centred = true)" : :RMSProp
```

cossio:
I re-added the suggestions for the name (`centre` -> `centred`), without the logic changes.

A suggestion for the docstring (I cannot add this as a Github suggestion bc it's not in the PR):

```
# Parameters
...
- Centred RMSProp: (`centred`): if `false` (default), gradients are normalized by an estimation of their second moments; if `true`, normalizes by the gradient variance instead of the second moment (http://arxiv.org/abs/1308.08500).
```

Perhaps we can merge this?

ToucheSir:
@mcabbott saw you added a couple changes. Is there anything left on the docket or is this good to go?

mcabbott:
Maybe the only Q is whether you give the constructor a verb "centre this" or an adjective "make the centred version".

ToucheSir:
PyTorch and TF both use the adjective form, let's go with that.

mcabbott:
Done. But what is wrong with the tests?

Locally, something else is broken:
```
RMSProp: Test Failed at /Users/me/.julia/dev/Optimisers/test/rules.jl:48
  Expression: iloss(rand(10, 10), w, w′) < 0.01
   Evaluated: 0.07415677254265908 < 0.01
Stacktrace:
 [1] macro expansion
   @ ~/.julia/dev/julia/usr/share/julia/stdlib/v1.9/Test/src/Test.jl:464 [inlined]
 [2] macro expansion
   @ ~/.julia/dev/Optimisers/test/rules.jl:48 [inlined]
...
 [6] top-level scope
   @ ~/.julia/dev/Optimisers/test/rules.jl:36
name(o) = :AdaGrad
sum(m) / sum(1:64) = 0.06019807093228864
```

PR #53, opened 2022-02-05
Shorten leaf printing

mcabbott:
If you print the status of the example in the docs at the readme, this makes it about 4 screenfuls, instead of 20something. But shows a few numbers so that you can see whether the state is all zero. RFC?


src/interface.jl
```diff
@@ -116,8 +116,11 @@ onevalue(λ, x::AbstractArray{T}) where T = onevalue(convert(float(T), λ), x)
...
 function Base.show(io::IO, ℓ::Leaf)  # show method is mostly to hide its long type!
   ioc = IOContext(io, :compact => true)
-  print(ioc, "Leaf(", ℓ.rule, ", ")
-  show(ioc, ℓ.state)
-  print(io, ")")
+  printstyled(io, "Leaf("; color = :green)
+  show(ioc, ℓ.rule)
+  printstyled(", "; color = :green)
+  str = sprint(show, ℓ.state; context = ioc)
+  print(io, length(str) < 70 ? str : first(str, 50) * " … ")
```
ToucheSir, line 123: Is there functionality in Base or InteractiveUtils that can help cut down on magic numbers?

mcabbott: I think you can get terminal size, but don't remember how reliably. However, I'm not sure that's what you want here. The goal is to show the optimiser itself, plus a few numbers of its state (to see if all zero), and that doesn't depend on window size.

If there were always an outermost struct this package owned, then that could arrange to only print one screenful, like matrix printing does. But there isn't right now.

ToucheSir: I noticed that `display` for arrays can insert the `...` in the middle instead of cutting things off at the end based on terminal width. Is there a way to do this but with a custom width like the one specified here?

PR #54, opened 2022-02-06
Add `destructure`, take II

mcabbott:
This adds a `destructure` function, like Flux's, which should handle awkward cases. Reqiures https://github.com/FluxML/Functors.jl/pull/37 . 

Alternative to #40 . This seems like a tidier approach, although not quite as short and elegant as it was before I started adding tests. The key idea is that, on the first walk over the model to flatten it, you can make a tree of vector offsets, which simplifies the reconstruction step and the gradients. The gradient of reconstruction isn't an `fmap` walk, but because it already knows the offsets, it does not care if the walks' orders don't match. 

Should work with numbers too if `isnumeric` is widened to allow them. Should work with mixed element types too, promote for the vector, project back for the reconstruction.

The reason to put it here not in Functors is that this package must already depend on ChainRulesCore, and that this builds in `trainable` deeply enough to make having another version without it a pain. And, Functors at present doesn't have `isnumeric`.

Closes #40, closes https://github.com/FluxML/Functors.jl/pull/31 if it can.

### DIFF
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -2,6 +2,9 @@
 using ChainRulesCore: ChainRulesCore, NoTangent, ProjectTo, unthunk
 const NoT = NoTangent()
 
+base(dx::Tangent{<:Tangent}) = backing(dx).backing  # might be needed for gradient(gradient(destructure))
+base(dx::Tangent{Any, <:NamedTuple{(:backing,)}}) = base(backing(dx).backing)  # Zygote version
+
 """
     destructure(model) -> vector, reconstructor
 
@@ -55,11 +58,11 @@ Base.length(re::Restructure) = re.length
 
 # This flattens a model, and returns a web of offsets for later use:
 function _flatten(x)
-  isnumeric(x) && return vcat(vec(x)), 0, length(x)  # trivial case
+  isnumeric(x) && return vcat(_vec(x)), 0, length(x)  # trivial case
   arrays = AbstractVector[]
   len = Ref(0)
   off = fmap(x; exclude = isnumeric, walk = (f, z) -> map(f, _trainable(z))) do y
-    push!(arrays, vec(y))
+    push!(arrays, _vec(y))
     o = len[]
     len[] = o + length(y)
     o
@@ -67,9 +70,12 @@ function _flatten(x)
   reduce(vcat, arrays), off, len[]
 end
 
+_vec(x::Number) = LinRange(x,x,1)
+_vec(x::AbstractArray) = vec(x)
+
 function ChainRulesCore.rrule(::typeof(_flatten), x)
   flat, off, len = _flatten(x)
-  _flatten_back((dflat, _, _)) = (NoT, _rebuild(x, off, dflat, len; walk = _Tangent_biwalk, prune = NoT))
+  _flatten_back((dflat, _, _)) = (NoT, _rebuild(x, off, unthunk(dflat), len; walk = _Tangent_biwalk, prune = NoT))
   (flat, off, len), _flatten_back
 end
 
@@ -92,7 +98,7 @@ function _trainable_biwalk(f, x, aux)
 end
 
 function _trainmap(f, ch, tr, aux)
-  map(ch, tr, aux) do c, t, a  # isnothing(t) indicates non-trainable field, safe given isnumeric(c)??
+  map(ch, tr, aux) do c, t, a  # isnothing(t) indicates non-trainable field, safe given isnumeric(c)
     isnothing(t) ? c : f(t, a)
   end
 end
@@ -121,7 +127,7 @@ ChainRulesCore.@non_differentiable _zero(x)
 # This is the gradient of model reconstruction, accumulating duplicates:
 function _grad!(x, dx, off, flat::AbstractVector)
   x′, _ = functor(typeof(x), x)
-  dx′, _ = functor(typeof(x), dx)
+  dx′, _ = functor(typeof(x), base(dx))
   off′, _ = functor(typeof(x), off)
   foreach((xᵢ, dxᵢ, oᵢ) -> _grad!(xᵢ, dxᵢ, oᵢ, flat), x′, dx′, off′)
   flat
@@ -134,7 +140,6 @@ _grad!(x, dx::Zero, off, flat::AbstractVector) = dx
 _grad!(x, dx::Zero, off::Integer, flat::AbstractVector) = dx  # ambiguity
 
 function ChainRulesCore.rrule(::typeof(_grad!), x, dx, off, flat)
-  println("grad! fwd ", length(flat))
   _grad_back(dflat) = (NoT, NoT, _rebuild(x, off, unthunk(dflat); walk = _Tangent_biwalk, prune = NoT), NoT, NoT)
   _grad!(x, dx, off, flat), _grad_back
 end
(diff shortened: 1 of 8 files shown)


src/destructure.jl
```diff
@@ -0,0 +1,105 @@
...
+  (flat, off, len), alpha_back
+end
+
+# This reconstructs either a model like x, or a gradient for it:
+function beta(x, off, flat::AbstractVector; len, walk = _trainable_biwalk, kw...)
+  len == length(flat) || error("wrong length")
+  fmap(x, off; exclude = isnumeric, walk, kw...) do y, o
+    _getat(y, o, flat)
+  end
+end
```
ToucheSir, line 60: Can we call these `(un)flatten(_model)`?

src/destructure.jl
```diff
@@ -0,0 +1,105 @@
...
+function alpha(x)
+  isnumeric(x) && return vcat(vec(x)), 0, length(x)  # trivial case
+  arrays = AbstractVector[]
+  len = Ref(0)
+  off = fmap(x; exclude = isnumeric, walk = (f, z) -> map(f, _trainable(z))) do y
+    push!(arrays, vec(y))
+    o = len[]
+    len[] = o + length(y)
+    o
+  end
```
ToucheSir, line 44: Off-topic, but this is a concrete example of where something like https://github.com/FluxML/Functors.jl/pull/32 or `fmapreduce` would help. Instead of writing to a ref, you'd just carry an accumulated total through the traversal.

src/destructure.jl
```diff
@@ -0,0 +1,105 @@
...
+function destructure(x)
+  flat, off, len = alpha(x)
+  flat, Restucture(x, off, len)
+end
+
+struct Restucture{T,S}
+  model::T
+  offsets::S
+  length::Int
+end
```
ToucheSir, line 30: I'm really happy this is being reified. The current implementation in Flux tries to dance around having to explicitly represent the state, but that's a big part of why it's so inflexible.

mcabbott: My main reason to replace this:
```julia
function destructure(x)
   flat, off, len = alpha(x)
   flat, v -> beta(x, off, v; len)
 end
 ```
 with 7 more lines was that the anonymous function's type contains the huge offset struct's type, so it fills your screen...
 
Glad you don't object though! This struct is very much internal still, of course.

Comment (ToucheSir) [reply src/destructure.jl:30]: Nice, that'll be a great QoL change too.

File: src/destructure.jl (hunk 2 above)
Comment (mcabbott) [reply src/destructure.jl:44]: Earlier commits, FWIW, just `append!`ed to a vector, and uses its length instead. This is faster, and delays working out the promoted type instead of needing another basically `fmapreduce` pass. A fancy way to pass the integer forwards might avoid the Ref, but we don't want to do pairwise `vcat` array-by-array.

Comment (ToucheSir) [reply src/destructure.jl:44]: Yes, I was referring to just the total instead of using carried state for the actual array of arrays (or single array). I assume the single array path was nixed because of not knowing the container type without traversing?

Comment (mcabbott) [reply src/destructure.jl:44]: Yes, getting the type needed another walk: https://github.com/FluxML/Optimisers.jl/pull/54/commits/5a7bfc8fe5372c807b857a674febf65f70cb8490#diff-2fc6059e247338b5ac149900b866865ae69bdf3693d9ce37cef19f230ddb8e30L104 
(And forgetting to make that walk non-differentiable gave some surprising bugs...)

File: src/destructure.jl (hunk 3 above)
Comment (darsnack) [reply src/destructure.jl:30]: This should also make it possible to port some of the speedy stuff in DiffEqFlux for `FastChain` etc. easier too. We can consider adding
```julia
(re::Restructure)(x, flat) = re(flat)(x)
```
for that.

darsnack:
Just a +1 comment from me. Brian already covered anything I would have.

src/destructure.jl
mcabbott: No downsides to that. 

One question here is whether we want overwriting versions. Should there be a `destructure!` whose `re` writes back into the same model? And if we want `(c::Chain)(x, p::Vector)` to call something like that, but never make the vector, can these share code?

ToucheSir: I've been thinking a bit around that in light of https://github.com/FluxML/Flux.jl/pull/1855 and https://github.com/FluxML/Flux.jl/pull/1809#issuecomment-1029721533, but it's not clear to me how it could be done somewhat generically within Flux's existing design. This would be a lot easier if layers didn't own their trainable parameters, Flax-style.

src/destructure.jl
mcabbott: Now they have names, with underscores.

test/destructure.jl
```diff
@@ -0,0 +1,166 @@
...
+  @test gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]
+  @test gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]
+
+  @testset "second derivative" begin
+    @test_broken gradient(collect(1:6.0)) do y
+      sum(abs2, gradient(x -> sum(abs2, re2(x)[1]), y)[1])
+    end[1] ≈ [8,16,24,0,0,0]
+    # ERROR: Need an adjoint for constructor ChainRulesCore.Tangent{Any, Tuple{Vector{Float64}, ChainRulesCore.ZeroTangent}}. Gradient is of type Tuple{Vector{Float64}, Vector{Float64}}
+    # with Zygote, which can be fixed by:
+    # Zygote.@adjoint Tangent{T,B}(x::Tuple) where {T,B<:Tuple} = Tangent{T,B}(x), dx -> (dx,)
```
mcabbott, line 128: I think the remaining question on this PR is whether and how much to care about 2nd derivatives. Some work, some don't. I convinced myself there is no bug in the basic logic. But in the details of when to wrap what in a Tangent, or unwrap it for Zygote, there might be bugs, here or upstream.

If we want to be pedantic we could make all 2nd derivatives an error, rather than risk any being wrong. Or a warning.

ToucheSir: At least a warning sounds good to me.

mcabbott: Done!

mcabbott: Good to go?

src/destructure.jl
```diff
@@ -0,0 +1,152 @@
...
+function _grad!(x, dx, off::Integer, flat::AbstractVector)
+  @views flat[off .+ (1:length(x))] .+= dx  # must visit all tied nodes
+  flat
+end
+_grad!(x, dx::Zero, off, flat::AbstractVector) = dx
+_grad!(x, dx::Zero, off::Integer, flat::AbstractVector) = dx  # ambiguity
+
+# These are only needed for 2nd derivatives:
+function ChainRulesCore.rrule(::typeof(_grad!), x, dx, off, flat)
+  @warn "second derivatives of Restructure may not work yet, sorry!" maxlog=3
```
ToucheSir, line 142: Given these were completely busted most of the time before, I don't think we need to apologize so profusely :laughing:

ToucheSir (approved):

src/destructure.jl
mcabbott: Heh. May as well be friendly!

Also, I think the point is that they ought to work, the structure does allow for them. Just it has bugs right now.

PR #56, opened 2022-02-09
Explicitly preserve eltype

### LINKED ISSUE #55: Correct the optimiser's eltype?
Should we add something to adjust the numbers in the optimiser to match the array? This hardly matters with Array etc, as both `update` and `update!` write back into an array of the same type. But with e.g. StaticArrays, a stray Float64 will promote the whole model. 

We could do something like this:
```julia
function retype(::Type{T}, opt::O) where {T,O}
  nt = map(n -> _retype(T, getfield(opt, n)), fieldnames(O)) 
  O.name.wrapper(nt...)
end
_retype(::Type{T}, p::AbstractFloat) where T = convert(T, p)
_retype(::Type{T}, t::Union{Tuple, NamedTuple}) where T = map(p -> _retype(T,p), t)
_retype(::Type, x) = x
```
This requires that the optimiser struct have a default constructor, and will only convert numbers and tuples. But if we add something like this now, it will be obvious from the beginning that your optimiser struct ought to work this way.

mcabbott:
Closes #55

This now rejects arrays of integers as not being things to optimise. Previously, they were, but most optimisers will promote to a float, and `update!` then didn't write in-place, and returned arrays of floats on the first run. I start to think it's cleaner just to declare that `Vector{Int}` cannot be trained.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -21,7 +21,7 @@ function setup(rule, x; seen = Base.IdSet())
   end
 end
 
-subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : (x .- x̄)
+subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
 
 update!(::Nothing, x, ::Zero, ::Zero...) = nothing, x
 update!(::Nothing, x, x̄s...) = nothing, x
@@ -50,10 +50,10 @@ end
 apply!(o, state, x, dx, dxs...) = apply!(o, state, x, dx)
 
 isnumeric(x::AbstractArray{<:Number}) = isleaf(x)  # isleaf to allow for e.g. transposed shared weights
-isnumeric(x::AbstractArray{<:Bool}) = false  # convention of ChainRules is that Bool is non-differentiable
+isnumeric(x::AbstractArray{<:Integer}) = false
 isnumeric(x) = false
 
-iswriteable(::DenseArray{<:AbstractFloat}) = true  # more elaborate versions are possible, wait until needed?
+iswriteable(::DenseArray) = true  # more elaborate versions are possible, wait until needed?
 iswriteable(_) = false
 
 """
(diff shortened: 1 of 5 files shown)


src/interface.jl
```diff
@@ -21,7 +21,7 @@ function setup(rule, x; seen = Base.IdSet())
   end
 end
 
-subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : (x .- x̄)
+subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
```
ToucheSir, line 24:
```suggestion
subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : convert(typeof(x), x .- x̄)
```
Are there any notable cases where this fails? `@less` tells me that it'll take a simpler path for StaticArrays.

mcabbott: I get:
```julia
julia> using StaticArrays

julia> x = @SVector rand(Float32, 100); xb = @SVector rand(100);

julia> @btime Float32.($x .- $xb);
  min 21.919 ns, mean 22.044 ns (0 allocations)

julia> @btime convert.(Float32, $x .- $xb);
  min 43.392 ns, mean 44.969 ns (1 allocation, 16 bytes)
```

mcabbott: Oh now I see you don't write eltype. I think that's worse, it won't fuse:
```
julia> @btime convert(typeof($x), $x .- $xb);
  min 24.682 ns, mean 24.869 ns (0 allocations)

julia> x = rand(Float32, 1000); xb = rand(1000); xnew = rand(Float32, 1000);

julia> @btime $xnew .= Float32.($x .- $xb);
  min 287.054 ns, mean 289.051 ns (0 allocations)

julia> @btime $xnew .= convert.(Float32, $x .- $xb);
  min 287.018 ns, mean 287.941 ns (0 allocations)

julia> @btime xnew .= convert(typeof($x), $x .- $xb);
  min 785.714 ns, mean 2.957 μs (3 allocations, 12.02 KiB)
```

ToucheSir: I was thinking specifically of https://github.com/JuliaGPU/CUDA.jl/issues/261 and how `T.(::AbstractArray{T})` allocates for non-static arrays, but the first is fixed and the second is irrelevant because of fusion.

ToucheSir (approved):

PR #57, opened 2022-02-14
Add `total(f, model)` to replace implicit `sum(f, Flux.params(model))`

mcabbott:
This proposes to add some kind of differentiable `sum(f, trainable(x))` which walks the model. I'm not certain this is the right thing yet. 

Right now this gets all trainable parameters. But perhaps a variant which takes a type `total(f, Union{Dense, Conv}, model)` might be a better explicit-parameters replacement for `modules`? Xref  https://github.com/FluxML/Flux.jl/pull/1863#discussion_r800084248

Closes https://github.com/FluxML/Functors.jl/issues/35 , probably.

Edit: since I couldn't find this, big Flux issue about explicit parameters is https://github.com/FluxML/Flux.jl/issues/1986 and snippet with a quick way to write `total` here: https://github.com/FluxML/Flux.jl/issues/2040#issuecomment-1535535892

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -9,7 +9,7 @@ export AbstractRule
 include("adjust.jl")
 
 include("destructure.jl")
-export destructure
+export destructure, total
 
 include("rules.jl")
 export Descent, Adam, Momentum, Nesterov, Rprop, RMSProp,
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -1,6 +1,6 @@
 
-using ChainRulesCore: ChainRulesCore, NoTangent, ProjectTo, unthunk
-const NoT = NoTangent()
+using ChainRulesCore: ChainRulesCore, ProjectTo, unthunk, RuleConfig, HasReverseMode, rrule_via_ad
+const NoT = ChainRulesCore.NoTangent()
 
 """
     destructure(model) -> vector, reconstructor
@@ -124,9 +124,11 @@ function (::_Tangent_biwalk)(f, x, aux)  # use with prune = NoT
   y = _trainmap(f, ch, _trainable(x), au)
   y isa Tuple{} && return NoT
   p = ProjectTo(x)
-  if p isa ProjectTo  # e.g. Array, NamedTuple
-    p(y)
-  else  # p === identity for unknown structs
+  # if p isa ProjectTo  # e.g. Array, NamedTuple
+  #   p(y)  # but for NamedTuple, this hits https://github.com/JuliaDiff/ChainRulesCore.jl/issues/538
+  if x isa Union{Number, AbstractArray}  # these don't use Tangent
+    ProjectTo(x)(unthunk(y))
+  else
     Tangent{typeof(x), typeof(y)}(y)
   end
 end
@@ -173,4 +175,65 @@ _maybewarn() = nothing
 function ChainRulesCore.rrule(::typeof(_maybewarn))
   @warn "second derivatives of destructure may not work yet, sorry!" maxlog=3
   nothing, _ -> (NoT,)
+end
+
+"""
+    total(f, model)
+
+Applies `f` to every [`trainable`](@ref), [`isnumeric`](@ref) parameter in
+the model, and returns the sum. Differentiable. Counts shared weights once. 
+
+# Examples
+```jldoctest
+julia> m = (x = [3.0, 4.0], y = (sin, [5.0]), z = (6, 7));
+
+julia> total(sum, m)
+12.0
+
+julia> total(norm, m)
+10.0
+
+julia> total(length, m) == length(destructure(m)[1])
+true
+```
+"""
+function total(f, x)
+  values = []
+  fmap(y -> push!(values, f(y)), x; exclude = isnumeric, walk = (f, z) -> foreach(f, _trainable(z)))
+  sum(values)
+end
+
+function ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(total), f, x)
+  z, backs = _total_hobbit(config, f, x)
+  total_back(dz) = (NoT, _total_grad(unthunk(dz), x, backs)...)
+  z, total_back
+end
+
+function _total_hobbit(config::RuleConfig, f, x)
+  values = []
+  backs = fmap(x; exclude = isnumeric, walk = (f, z) -> map(f, _trainable(z))) do y
+    val, back = rrule_via_ad(config, f, y)
+    push!(values, val)
+    back
+  end
+  sum(values), backs
+end
+
+function _total_grad(dz, x, backs)
+  dfs = []
+  dx = fmap(x, backs; exclude = isnumeric, walk = _Tangent_biwalk, prune = NoT) do y, b
+    df, dy = b(dz)
+    push!(dfs, df)
+    dy
+  end
+  sum(dfs), dx
+end
+
+function ChainRulesCore.rrule(::typeof(_total_grad), dz, x, backs)
+  @warn "second derivatives of total(f, x) may not work yet, sorry!" maxlog=3
+  function grad_back((df, dx))
+    df isa Zero || @error "second derivatives of total(f, x) with respect to the function are wrong!"
+    (NoT, total(dx), NoT, NoT)
+  end
+  _total_grad(dz, x, backs), grad_back
 end
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -52,11 +52,13 @@ Optimisers.trainable
 Optimisers.isnumeric
 ```
 
-Such restrictions are also obeyed by this function for flattening a model:
+Such restrictions are also obeyed by this function for flattening a model,
+and one for applying a function to every parameter:
 
 ```@docs
 Optimisers.destructure
 Optimisers.Restructure
+Optimisers.total
 ```
 
 ## Rule Definition
--- a/test/destructure.jl
+++ b/test/destructure.jl
@@ -296,3 +296,26 @@ tmp1
     y, bk = Zygote.pullback(x -> sum(destructure(x)[1]), (3, 4))
     @test bk(1.0) == (nothing,)
 end
+
+@testset "total" begin
+  @test total(sum, m1) == sum(1:3)
+  @test total(prod, m2) == prod(1:3) + prod(4:6)
+  @test total(sum, m3) == sum(1:6)
+  @test total(sum, m4) == sum(1:6)  # shared only counts once
+  @test total(sum, m6) == 6 + 4 + im
+
+  @test gradient(m -> total(sum, m), m1) == ([1,1,1],)
+  @test gradient(m -> total(sum, m), m3)[1] == (x = [1,1,1], y = nothing, z = [1,1,1])
+  @test gradient(m -> total(sum, m), m4)[1] == (x = [1,1,1], y = nothing, z = [1,1,1])
+  g6 = gradient(m -> abs2(total(sum, m)), m6)[1]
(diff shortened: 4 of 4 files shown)


src/Optimisers.jl
```diff
@@ -6,7 +6,7 @@ using LinearAlgebra
 include("interface.jl")
 
 include("destructure.jl")
-export destructure, total, total2
+export destructure, total
```
mcabbott, line 9: Clearly I forgot to delete this from the previous PR...

src/destructure.jl
```diff
@@ -107,9 +107,11 @@ function _Tangent_biwalk(f, x, aux)  # use with prune = NoT
...
   p = ProjectTo(x)
-  if p isa ProjectTo  # e.g. Array, NamedTuple
-    p(y)
-  else  # p === identity for unknown structs
+  # if p isa ProjectTo  # e.g. Array, NamedTuple
+  #   p(y)  # but for NamedTuple, this hits https://github.com/JuliaDiff/ChainRulesCore.jl/issues/538
+  if x isa Union{Number, AbstractArray}  # these don't use Tangent
+    ProjectTo(x)(unthunk(y))
+  else
     Tangent{typeof(x), typeof(y)}(y)
```
mcabbott, line 115: This is either a bug in earlier `_Tangent_biwalk`, or in ChainRulesCore.

src/destructure.jl
```diff
@@ -150,3 +152,64 @@ function ChainRulesCore.rrule(::typeof(_maybewarn))
...
+
+julia> total(length, m) == length(destructure(m)[1])
+true
+```
+"""
+function total(f, x)
+  values = []
+  fmap(y -> push!(values, f(y)), x; exclude = isnumeric, walk = (f, z) -> foreach(f, _trainable(z)))
+  sum(values)
+end
```
mcabbott, line 180: While `Any[]` doesn't seem great, this ends up about the same speed as my other idea:
```
const INIT = Base._InitialValue()

function total2(f, x; init = INIT)
  fmap(x; exclude = isnumeric, walk = (f, z) -> foreach(f, _trainable(z))) do y
    val = f(y)
    init = init===INIT ? val : (init+val)
  end
  init
end

julia> @btime total(norm, $model)  # Resnet from the docs
  min 23.863 ms, mean 23.995 ms (1541 allocations, 130.06 KiB)
730.5533f0

julia> @btime total2(norm, $model)
  min 23.834 ms, mean 23.982 ms (1538 allocations, 128.17 KiB)
730.5533f0

julia> m = (x = [3.0, 4.0], y = (sin, [5.0]), z = (6, 7));

julia> @btime total(norm, $m)
  min 1.750 μs, mean 1.846 μs (16 allocations, 752 bytes)
10.0

julia> @btime total2(norm, $m)
  min 1.675 μs, mean 1.769 μs (15 allocations, 640 bytes)
10.0
```

rejuvyesh:
Should this be more general to allow computing the norm of the gradients as well?

src/destructure.jl
```diff
@@ -150,3 +152,64 @@ function ChainRulesCore.rrule(::typeof(_maybewarn))
...
+```jldoctest
+julia> m = (x = [3.0, 4.0], y = (sin, [5.0]), z = (6, 7));
+
+julia> total(sum, m)
+12.0
+
+julia> total(norm, m)
+10.0
+
+julia> total(length, m) == length(destructure(m)[1])
```
mcabbott, line 172: This would solve https://github.com/FluxML/Flux.jl/issues/2043 (as long as trainable parameters are what you want).

Or `total(Base.summarysize, m)` for bytes, `total(_ -> 1, m)` to count arrays.

ToucheSir:
One idea for making this differentiable is to take the non-diff part (caching) out of `fmap` and instead creating a memoized callback wrapper type. Then `fmap` itself is trivially differentiable and the only remaining tricky bit is how to make that wrapper Nth times differentiable.

mcabbott:
Not sure I follow. If `fmap` were made differentiable, how would you write `total(norm, model)` using it?

ToucheSir:
`fmap` itself is not sufficient for this. My thought was that you could use it as the `map` part of a `mapreduce`, where cached leaves are replaced with some neutral element. Then the sum reduction requires no caching and can be made fully type stable. The idea behind this is to avoid the `Any[]` in https://github.com/FluxML/Optimisers.jl/pull/57/files#r806313306. Things won't be type stable under AD unfortunately, but I can see a possible path for making it so with changes to Functors internals.

mcabbott:
OK. I guess this PR's take is that since essentially nothing else about Functors.jl is type-stable, everything takes a few μs, there's not much point pushing hard here. 

Decomposing into `fmap` then `reduce` ... the `prune` keyword lets you drop duplicates, but I've forgotten why `fmapstructure` is different here:
```
julia> twice = [1,2.0];

julia> m = (twice, [3,4.0], twice, sin, 99);

julia> fmap(x -> Some(length(x)), m, prune=nothing, exclude=(x->x isa Array))
(Some(2), Some(2), nothing, sin, 99)  # need Some to know to ignore 99

julia> fmapstructure(x -> length(x), m, prune=0, exclude=(x->x isa Array))
(2, 2, 0, (), ())  # easier
```

ToucheSir:
`fmap` is the only intrinsically type unstable part of Functors. Part of that is due to the IdDict cache, but another is the fragility of a chain of mutually recursive functions under type inference that we've accrued over time in the name of backwards compatibility. https://github.com/FluxML/Functors.jl/pull/32 was in part exploring if type stable structural traversals in Functors are possible, and indeed some are (e.g. structural sum).

For this PR, my main ask is that it not cut off any paths which could bring us better type stability in the future. That doesn't seem to be the case, but I don't understand the (un)thunking well enough to say for sure. Minor comments would be the inclusion of an `init` kwarg and request for bikeshedding the name.

src/destructure.jl
```diff
@@ -166,3 +168,64 @@ function ChainRulesCore.rrule(::typeof(_maybewarn))
...
+  sum(values)
+end
+
+function ChainRulesCore.rrule(config::RuleConfig{>:HasReverseMode}, ::typeof(total), f, x)
+  z, backs = _total_hobbit(config, f, x)
+  total_back(dz) = (NoT, _total_grad(unthunk(dz), x, backs)...)
+  z, total_back
+end
+
+function _total_hobbit(config::RuleConfig, f, x)
```
ToucheSir, line 204: A brief comment about what this and `_total_grad` do would help. "hobbit" in particular is alien terminology for anyone who hasn't read a couple of specific issues on the ChainRules repo :P

mcabbott: Indeed. I rebased this but realised I have no memory of how it worked. Will revise or re-write.

ToucheSir:
Now that people are starting to use explicit params, we've seen a few instances where it would be nice to have a easy method for adding regularization terms. I believe this function should be easier to implement in a post-https://github.com/FluxML/Functors.jl/pull/43 world too.

mcabbott:
#143 and https://github.com/FluxML/Optimisers.jl/pull/57#issuecomment-1065598383 hint that the signature here should probably allow for `total(f, model, grads)` to mean roughly `sum(f(x,dx) for ...)`. That opens the thorny question of what happens to `dx` when the same `x` appears twice, are they added? Does `total(f, model, grads)` need to be differentiable?

ToucheSir:
> That opens the thorny question of what happens to `dx` when the same `x` appears twice, are they added?

Makes sense, though thinking about this stuff is always rather mind-bending.

> Does `total(f, model, grads)` need to be differentiable?

Yes, unless we can provide an alternative way for users to add a regularization term in their loss function with explicit params.

darsnack:
> That opens the thorny question of what happens to dx when the same x appears twice, are they added?

In the case of `total`, I think yes. But in the general case, it should apply the reduction operator supplied?

darsnack:
Also, I don't think adding `total(f, model, grads)` is worth holding this PR up.

mcabbott:
> But in the general case, it should apply the reduction operator supplied?

Thinking more, I think do think they are always added. 

If I do something like `mapreduce(norm, max, ([1,2], [3 4 5]))`, then the reduction operator only makes sense on the output of the mapped function -- here `max` on two scalars. It cannot be applied to arrays.

In some `fmapreduce(f, op, model, grads)`, the accumulation of distinct `dx in grads` is part of AD, it should always happen before `f` is called, using `+` (or e.g. `Zygote.accum`) not `op`.

ablaom:
@mcabbott 

I'm trying to adapt the `total` method proposed here for a different [use case](https://github.com/FluxML/MLJFlux.jl/issues/221#issuecomment-1707604760) and wondered if the following was expected behaviour:

```julia
chain = Chain(Dense(3=>5), Dense(5=>1, relu))
f(A) = sum(abs2, A)

# old way of doing things:
sum(f.(Flux.params(chain)))
# 7.247263f0

# new way:
total(f, chain)
# ERROR: MethodError: no method matching isnumeric(::Chain{Tuple{Dense{typeof(identity), Mat# rix{Float32}, Vector{Float32}}, Dense{typeof(relu), Matrix{Float32}, Vector{Float32}}}})
```

mcabbott:
No, it's not... I suspect you have Base.isnumeric, there's an unfortunate name clash? Or perhaps this has just rotted, sorry. I mean to revisit but have been busy.

ablaom:
Okay, nevermind. I was using `Base.isnumeric` instead of the local one.

ablaom:
Thanks for the lightning reply ❤️

src/destructure.jl
```diff
@@ -166,3 +168,64 @@ function ChainRulesCore.rrule(::typeof(_maybewarn))
...
+julia> total(norm, m)
+10.0
+
+julia> total(length, m) == length(destructure(m)[1])
+true
+```
+"""
+function total(f, x)
+  values = []
+  fmap(y -> push!(values, f(y)), x; exclude = isnumeric, walk = (f, z) -> foreach(f, _trainable(z)))
```
ablaom, line 194: Are we sure `total` is actually differentiable? In my use case, pasting this definition into the MLJFlux  `train!`function,  I'm hitting

```julia
ERROR: Mutating arrays is not supported -- called push!(Vector{Any}, ...)
This error occurs when you ask Zygote to differentiate operations that change
the elements of arrays in place (e.g. setting values with x .= ...)
```

with a stack trace pointing to this line.

ablaom: Sorry I don't have a MWE

darsnack: Are you checking out the PR into your environment or copy-pasting the changes manually? Cause there is a custom `rrule` for `total` right below, so Zygote shouldn't even be trying to differentiate this line.

ablaom: Right. I'm just copy-pasting and missed the `rrule` overload. Thanks.

darsnack:
Bump on merging this? We still get regularization questions frequently.

mcabbott:
Now rebased, but the tests tell me I need to remember what on earth `_Tangent_biwalk` does.

CarloLucibello:
I think that instead of introducing `total`, for more generality we should have a `trainables(model)` returning an iterable (also broadcastable and differentiable) over the training parameters. This can be used in expressions like `sum(f, trainables(model))` and much more.

ToucheSir:
Are we still running into recompilation issues using `sum(f, trainables(model))`? If not I'm fine with not pursuing this if @mcabbott is as well. Otherwise we may need to dig into whether the compilation issue is solvable or a fundamental flaw with Zygote.

CarloLucibello:
I don't think we have recompilation issues with `sum(f, trainables(model))`. Closing as it is better not increase the api surface when we have a neat way to do things.

PR #58, opened 2022-02-20
Allow zero-arrays

mcabbott:
This wants to check that zero-dimensional arrays are allowed, and preserved. 

But they aren't right now, because `subtract!` will return a number. Unless it's writing in-place. Unless we use `broadcast_preserving_zero_d`, which, edit, this PR now does:
```julia
julia> m = (a = fill(1.0), b = SArray{Tuple{}}(fill(1.0)), c = PermutedDimsArray(fill(1.0), ()));

# normal arrays

julia> m.a .- m.a
0.0

julia> Broadcast.broadcast_preserving_zero_d(-, m.a, m.a)
0-dimensional Array{Float64, 0}:
0.0
```
This doesn't work for StaticArrays, maybe that's their problem not ours:
```
julia> m.b .- m.b
Scalar{Float64}((0.0,))

julia> Broadcast.broadcast_preserving_zero_d(-, m.b, m.b)
ERROR: MethodError: no method matching similar(::Base.Broadcast.Broadcasted{StaticArrays.StaticArrayStyle{0}, Nothing, typeof(-), Tuple{Scalar{Float64}, Scalar{Float64}}}, ::Type{Scalar{Float64}}, ::Tuple{})
Closest candidates are:
  similar(::Union{Adjoint{T, S}, Transpose{T, S}} where {T, S}, ::Type{T}, ::Tuple{Vararg{Int64, N}}) where {T, N} at ~/.julia/dev/julia/usr/share/julia/stdlib/v1.8/LinearAlgebra/src/adjtrans.jl:212
  similar(::Base.ReinterpretArray, ::Type, ::Tuple{Vararg{Int64, N}} where N) at ~/.julia/dev/julia/usr/share/julia/base/reinterpretarray.jl:185
  similar(::UpperHessenberg, ::Type{T}, ::Tuple{Vararg{Int64, N}}) where {T, N} at ~/.julia/dev/julia/usr/share/julia/stdlib/v1.8/LinearAlgebra/src/hessenberg.jl:61
  ...
Stacktrace:
 [1] similar(bc::Base.Broadcast.Broadcasted{StaticArrays.StaticArrayStyle{0}, Nothing, typeof(-), Tuple{Scalar{Float64}, Scalar{Float64}}}, #unused#::Type{Scalar{Float64}})
   @ Base.Broadcast ./broadcast.jl:211
```
~~Alternatively, we could ban zero-arrays as being too weird.~~

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -21,7 +21,13 @@ function setup(rule, x; seen = Base.IdSet())
   end
 end
 
-subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
+function subtract!(x, x̄)
+  if iswriteable(x)
+    x .= x .- x̄
+  else
+    broadcast_preserving_zero_d(eltype(x), broadcasted(-, x, x̄))
+  end
+end
 
 update!(::Nothing, x, ::Zero, ::Zero...) = nothing, x
 update!(::Nothing, x, x̄s...) = nothing, x
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -2,6 +2,7 @@ module Optimisers
 
 using Functors: functor, fmap, isleaf
 using LinearAlgebra
+using Base.Broadcast: broadcast_preserving_zero_d, broadcasted
 
 include("interface.jl")
 
--- a/test/rules.jl
+++ b/test/rules.jl
@@ -230,24 +230,27 @@ end
 @testset "zero-dim arrays" begin
   empty!(LOG)
   @testset "$(name(o))" for o in RULES
-    m = (a = fill(1.0), b = SArray{Tuple{}}(fill(1.0)), c = PermutedDimsArray(fill(1.0), ()))
+    m = (; arr = fill(1.0), pda = PermutedDimsArray(fill(1.0), ()), ref = Ref(1.0))
+    # The point of PermutedDimsArray here is to test the out-of-place path, so check:
+    @test Optimisers.iswriteable(m.arr)
+    @test !Optimisers.iswriteable(m.pda)
     s = Optimisers.setup(o, m)
     for _ in 1:10^3
-      g = loggradient(o)(x -> abs2(first(x.a) + first(x.b) + first(x.c)), m)[1]
+      g = loggradient(o)(x -> abs2(first(x.arr) + first(x.pda) + first(x.ref)), m)[1]
       s, m = Optimisers.update(s, m, g)
     end
-    # The main point here is that broadcasting should not accidentally make a scalar,
-    # but `m.a` is mutated, and `m.b .+ 1` is an array, so only `m.c` is a real test.
-    @test m.a isa Array{Float64, 0}
-    @test m.b isa AbstractArray{Float64, 0}
-    @test_broken m.c isa AbstractArray{Float64, 0}
+    # Goal is to check that broadcasting does not accidentally make a scalar,
+    # but `m.arr` iscopied &  mutated, so only `m.pda` is a real test:
+    @test m.arr isa Array{Float64, 0}
+    @test m.pda isa AbstractArray{Float64, 0}
+    @test m.ref isa Ref  # because it's mutated, broadcast_preserving_zero_d would make an array
     if o isa RADAM
-      @test sum(m.a) < 0.7
-      @test_broken sum(m.a) < 0.3
+      @test sum(m.arr) < 0.7
+      @test_broken sum(m.arr) < 0.3
     else
-      @test sum(m.a) < 0.3
-      @test sum(m.b) < 0.3
-      @test sum(m.c) < 0.3
+      @test sum(m.arr) < 0.3
+      @test sum(m.pda) < 0.3
     end
+    @test_broken only(m.ref) < 0.3  # not currently regarded as trainable
   end
 end
--- a/test/rules.jl
+++ b/test/rules.jl
@@ -251,6 +251,6 @@ end
       @test sum(m.arr) < 0.3
       @test sum(m.pda) < 0.3
     end
-    @test_broken only(m.ref) < 0.3  # not currently regarded as trainable
+    @test only(m.ref) ≈ 1  # not currently regarded as trainable
   end
 end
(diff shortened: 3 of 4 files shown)


mcabbott:
Should this allow `Ref` as a mutable zero-dim container in which to store scalar parameters?

One way to do so would be make it a scalar as far as the rules are concerned, by unwrapping, something like this:
```
isnumeric(x::Ref{<:Number}) = true  # these aren't leaf according to functors
isnumeric(x::Ref{<:Integer}) = false

iswriteable(::Ref) = true  # needed not for subtract!, but for update's fmap(copy)
init(o, x::Ref) = init(o, x.x)  # treat like a scalar for what state to store

update!(ℓ::Leaf, x::Ref, ::Zero, ::Zero...) = ℓ, x
function update!(ℓ::Leaf, x::Ref, x̄s...)
  s′, x̄′ = apply!(ℓ.rule, ℓ.state, x.x, map(x̄ -> base(x̄).x, x̄s)...)
  x[] = x[] .- x̄′
  Leaf(ℓ.rule, s′), x
end
```

src/interface.jl
```diff
@@ -21,7 +21,13 @@ function setup(rule, x; seen = Base.IdSet())
   end
 end
 
-subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
+function subtract!(x, x̄)
+  if iswriteable(x)
+    x .= x .- x̄
+  else
+    broadcast_preserving_zero_d(eltype(x), broadcasted(-, x, x̄))
```
darsnack, line 28: Is this a public function? Meaning we can expect it to be stable?

darsnack (commented):
I can see how one might accidentally get a 0-dim array, but a `Ref` is a conscious choice. What's the value over using a 0-dim array or single element array? I thinking ignoring them is fine.

mcabbott:
For zero-arrays, someone will try, and I think it should either work or be an error. Seems bad if `update!` can silently replace them with numbers & then stop training, which is what falls out of broadcasting. They work in Flux now, and mutable ones will similarly work here, I think.

Maybe `Ref` ought to be a separate issue. It's the obvious mutable one-element container. But perhaps telling people to use a vector is fine. Currently ignored by Flux.

src/interface.jl
ToucheSir: I doubt this is the only use of it in the wild either: https://juliahub.com/ui/Search?q=broadcast_preserving_zero_d&type=code

darsnack: Good enough for me!

darsnack:
> For zero-arrays, someone will try, and I think it should either work or be an error. Seems bad if update! can silently replace them with numbers & then stop training, which is what falls out of broadcasting. They work in Flux now, and mutable ones will similarly work here, I think.

👍🏾 

> Maybe Ref ought to be a separate issue. It's the obvious mutable one-element container. But perhaps telling people to use a vector is fine. Currently ignored by Flux.

Certainly not contentious enough on my end that it can't be resolved now, but Flux's advice for scalar params has been a 1-element vector for quite some time. I don't see the practical reason someone would want `Ref`, so it seems safer to just not support it (by treating it non-trainable).

ToucheSir:
While looking for https://github.com/JuliaLang/julia/pull/35778#issuecomment-635576085 as background reading for this discussion, I found https://juliaarrays.github.io/StaticArrays.jl/latest/pages/api/#Scalar. Perhaps it's fine to ask users to use a type like this if they want 0-d arrays?

mcabbott:
That plays badly with Functors right now:
```julia
julia> fmap(println, (x = Fill(1.0), y = Fill(1.0)))
Fill(1.0)
(x = nothing, y = nothing)
```
And if we fix it, we can fix scalars too.

Really you should use a 1-element vector, `[1.0]`, for now. But someone will use an `Array{T,0}`, and this package should either work, or not work (and tell you), I think.

This PR votes for "work" now. `broadcast_preserving_zero_d` is a little ugly. I was concerned that more broadcasts within the rules would go wrong, by producing a scalar, but they seem to work OK.

ToucheSir:
I thought the consternation with using `broadcast_preserving_zero_d` was that certain array types in the wild don't work correctly with it? Or has that been resolved since the last top post update?

mcabbott:
Yes I don't know what to do there. These don't follow Base:
```julia
julia> using FillArrays

julia> Fill(1) .+ Fill(2)
0-dimensional Fill{Int64}, with entry equal to 3

julia> using StaticArrays

julia> s0 = SArray{Tuple{}}(fill(1.0))
Scalar{Float64}((1.0,))

julia> s0 isa AbstractArray{Float64, 0}
true

julia> s0 .+ s0
Scalar{Float64}((2.0,))
```
which results in weird behaviour or errors:
```
julia> Broadcast.broadcast_preserving_zero_d(-, Fill(1), Fill(2))
0-dimensional Array{Fill{Int64, 0, Tuple{}}, 0}:
Fill(-1)

julia> Broadcast.broadcast_preserving_zero_d(-, s0, s0)
ERROR: MethodError: no method matching similar(::Base.Broadcast.Broadcasted{StaticArrays.StaticArrayStyle{0}, Nothing, typeof(-), Tuple{Scalar{Float64}, Scalar{Float64}}}, ::Type{Scalar{Float64}}, ::Tuple{})
```
xref https://github.com/JuliaArrays/FillArrays.jl/issues/145 and nothing https://github.com/JuliaArrays/StaticArrays.jl/search?q=broadcast_preserving_zero_d

They also don't really work with Functors.jl, so perhaps they aren't usable anyway.

Maybe this PR is a bit random then, since Array actually works (by mutation) and ImmutableArrays haven't landed yet. 

And wrappers like this aren't preserved by broadcasting:
```julia
julia> using ReadOnlyArrays

julia> ReadOnlyArray(fill(0))
0-dimensional ReadOnlyArray{Int64, 0, Array{Int64, 0}}:
0

julia> ans .+ ans
0

julia> Broadcast.broadcast_preserving_zero_d(-, ReadOnlyArray(fill(0)), ReadOnlyArray(fill(0)))
0-dimensional Array{Int64, 0}:
0

julia> ReadOnlyArray([1,2]) .+ 3
2-element Vector{Int64}:
 4
 5
```

PR #60, opened 2022-03-07
Improve `WeightDecay` docstring

mcabbott:
This adds some comments borrowed from Flux in from #46, whose main change we decided wasn't needed.

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -442,20 +442,24 @@ end
 """
     WeightDecay(γ = 5f-4)
 
-Decay weights by `γ`.
+Decay weights by ``γ``, that is, add `γ .* x` to the gradient `x̄` which will be
+subtracted from `x`.
+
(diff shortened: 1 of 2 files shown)


src/rules.jl
```diff
@@ -442,20 +442,24 @@ end
...
+
+Typically composed  with other optimisers as the first transformation in an [`OptimiserChain`](@ref).
+This is equivalent to adding ``L_2`` regularization with coefficient ``γ`` to the loss.
 
 # Parameters
 - Weight decay (`γ`): Decay applied to weights during optimisation.
 """
 struct WeightDecay{T}
-  wd::T
+  gamma::T
```
ToucheSir, line 455: I guess our policy is that most fields are not considered part of the public API?

mcabbott: I guess so. And that 0.2 was so recent & we're still getting Flux to hook up to it.

darsnack (approved):

PR #66, opened 2022-04-20
Widen in `_grad!`

mcabbott:
Fixes a bug discovered in https://github.com/SciML/DiffEqFlux.jl/pull/699, that the adjoint of `re` has to allow for (some of) the leaves having a wider type than the original vector.

This approach costs a copy when this happens, but is much simpler than trying to walk the tree and infer the eltype before starting.


ToucheSir (approved):

CarloLucibello:
This supersedes #65

ChrisRackauckas:
Can this get a tag? I'd like to see if it fixes the last bits of https://github.com/SciML/DiffEqFlux.jl/pull/699 so we can merge 0.13 support.

CarloLucibello:
Tagged

PR #70, opened 2022-04-22
Fix #62

### LINKED ISSUE #62: `destructure` doesn't work correctly with certain functors
```julia
using Flux
using Functors
using Optimisers

struct Custom
    abc::Tuple
end
Functors.@functor Custom (abc,)

function (f::Custom)(x)
    x .* abc[1] .+ abc[2]
end

function Custom(;dim::Int)
    abc = (randn(Float32, dim), randn(Float32, dim))
    return Custom(abc)
end

model = Flux.Chain(Dense(4, 16), Custom(;dim=16), Dense(16, 4))
p, re = Optimisers.destructure(model)
re(p)(randn(Float32, 4, 1))
```

leads to

```julia
ERROR: LoadError: type Tuple has no field layers
Stacktrace:
  [1] getproperty
    @ ./Base.jl:42 [inlined]
  [2] functor(#unused#::Type{Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Custom, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}, c::Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:abc,), Tuple{Tuple{Int64, Int64}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}})
    @ Flux ~/.julia/packages/Flux/qAdFM/src/layers/basic.jl:44
  [3] _trainable_biwalk(f::Function, x::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Custom, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, aux::Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:abc,), Tuple{Tuple{Int64, Int64}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}})
    @ Optimisers ~/.julia/packages/Optimisers/UAVzc/src/destructure.jl:94
  [4] #fmap#30
    @ ~/.julia/packages/Functors/qBIlC/src/functor.jl:78 [inlined]
  [5] _rebuild(x::Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Custom, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, off::Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:abc,), Tuple{Tuple{Int64, Int64}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}}, flat::Vector{Float32}, len::Int64; walk::Function, kw::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Optimisers ~/.julia/packages/Optimisers/UAVzc/src/destructure.jl:83
  [6] _rebuild
    @ ~/.julia/packages/Optimisers/UAVzc/src/destructure.jl:82 [inlined]
  [7] (::Optimisers.Restructure{Chain{Tuple{Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}, Custom, Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}, NamedTuple{(:abc,), Tuple{Tuple{Int64, Int64}}}, NamedTuple{(:weight, :bias, :σ), Tuple{Int64, Int64, Tuple{}}}}})(flat::Vector{Float32})
    @ Optimisers ~/.julia/packages/Optimisers/UAVzc/src/destructure.jl:51
  [8] top-level scope
    @ ~/.julia/dev/PyCallChainRules/optimsers_destructure.jl:21
  [9] include(fname::String)
    @ Base.MainInclude ./client.jl:451
 [10] top-level scope
    @ REPL[9]:1
```

jondeuce:
This adds a couple small changes on top of [this draft PR](https://github.com/FluxML/Optimisers.jl/pull/63) in order to fix #62:

1. Wrap offset indices in a dummy struct `Offset` to fix the issue mentioned in #63 for array of arrays, where e.g. the offset structure for `x = [[1.0, 2.0]]` is now something like `o = [Offset(4)]` which is not leaflike, compared to `o = [4]` previously. Also opens the door to storing more information in this wrapper struct (original array size? eltype?), but doesn't seem to be necessary at this time
2. [`y = backing(re(y))`](https://github.com/jondeuce/Optimisers.jl/blob/927e095ae1f4742a27cc0710132ea1000ad1e20e/src/destructure.jl#L119) allows for `functor(x)` to return children which aren't its own fields: first restructure the gradient, then extract the `NamedTuple` backing for `Tangent`. It has the added benefit of adding some symmetry with `_trainable_biwalk` which naturally restructures the output of `_trainmap`, whereas `_Tangent_biwalk` previously did not

### DIFF
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -91,7 +91,7 @@ _getat(y::AbstractArray, o::Int, flat::AbstractVector) =
 
 function _trainable_biwalk(f, x, aux)
   ch, re = functor(typeof(x), x)
-  au, _ = functor(typeof(x), aux)
+  au, _ = functor(aux) 
   _trainmap(f, ch, _trainable(x), au) |> re
 end
 
@@ -103,7 +103,7 @@ end
 
 function _Tangent_biwalk(f, x, aux)  # use with prune = NoT
   ch, re = functor(typeof(x), x)
-  au, _ = functor(typeof(x), aux)
+  au, _ = functor(aux)
   y = _trainmap(f, ch, _trainable(x), au)
   y isa Tuple{} && return NoT
   p = ProjectTo(x)
@@ -126,7 +126,7 @@ ChainRulesCore.@non_differentiable _zero(x)
 function _grad!(x, dx, off, flat::AbstractVector)
   x′, _ = functor(typeof(x), x)
   dx′, _ = functor(typeof(x), base(dx))
-  off′, _ = functor(typeof(x), off)
+  off′, _ = functor(off)
   foreach((xᵢ, dxᵢ, oᵢ) -> _grad!(xᵢ, dxᵢ, oᵢ, flat), x′, dx′, off′)
   flat
 end
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -91,9 +91,12 @@ _getat(y::AbstractArray, o::Int, flat::AbstractVector) =
 
 function _trainable_biwalk(f, x, aux)
   ch, re = functor(typeof(x), x)
-  au, _ = functor(aux) 
+  au = _aux_children(aux) 
   _trainmap(f, ch, _trainable(x), au) |> re
 end
+
+_aux_children(off) = functor(off)[1]
+_aux_children(off::AbstractArray) = off  # leaflike according to Functors, but we need to see each offset
 
 function _trainmap(f, ch, tr, aux)
   map(ch, tr, aux) do c, t, a  # isnothing(t) indicates non-trainable field, safe given isnumeric(c)
@@ -103,7 +106,7 @@ end
 
 function _Tangent_biwalk(f, x, aux)  # use with prune = NoT
   ch, re = functor(typeof(x), x)
-  au, _ = functor(aux)
+  au = _aux_children(aux)
   y = _trainmap(f, ch, _trainable(x), au)
   y isa Tuple{} && return NoT
   p = ProjectTo(x)
@@ -126,7 +129,7 @@ ChainRulesCore.@non_differentiable _zero(x)
 function _grad!(x, dx, off, flat::AbstractVector)
   x′, _ = functor(typeof(x), x)
   dx′, _ = functor(typeof(x), base(dx))
-  off′, _ = functor(off)
+  off′ = _aux_children(off)
   foreach((xᵢ, dxᵢ, oᵢ) -> _grad!(xᵢ, dxᵢ, oᵢ, flat), x′, dx′, off′)
   flat
 end
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -53,16 +53,20 @@ end
 Base.show(io::IO, re::Restructure{T}) where T = print(io, "Restructure(", T.name.name, ", ..., ", re.length, ")")
 Base.length(re::Restructure) = re.length
 
+struct Offset
+  i::Int
+end
+
 # This flattens a model, and returns a web of offsets for later use:
 function _flatten(x)
-  isnumeric(x) && return vcat(_vec(x)), 0, length(x)  # trivial case
+  isnumeric(x) && return vcat(_vec(x)), Offset(0), length(x)  # trivial case
   arrays = AbstractVector[]
   len = Ref(0)
   off = fmap(x; exclude = isnumeric, walk = (f, z) -> map(f, _trainable(z))) do y
     push!(arrays, _vec(y))
     o = len[]
     len[] = o + length(y)
-    o
+    Offset(o)
   end
   reduce(vcat, arrays), off, len[]
 end
@@ -85,9 +89,9 @@ function _rebuild(x, off, flat::AbstractVector, len = length(flat); walk = _trai
   end
 end
 
-_getat(y::Number, o::Int, flat::AbstractVector) = ProjectTo(y)(flat[o + 1])
-_getat(y::AbstractArray, o::Int, flat::AbstractVector) =
-  ProjectTo(y)(reshape(flat[o .+ (1:length(y))], axes(y)))  # ProjectTo is just correcting eltypes
+_getat(y::Number, off::Offset, flat::AbstractVector) = ProjectTo(y)(flat[off.i + 1])
+_getat(y::AbstractArray, off::Offset, flat::AbstractVector) =
+  ProjectTo(y)(reshape(flat[off.i .+ (1:length(y))], axes(y)))  # ProjectTo is just correcting eltypes
 
 function _trainable_biwalk(f, x, aux)
   ch, re = functor(typeof(x), x)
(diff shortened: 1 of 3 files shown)


src/destructure.jl
```diff
@@ -85,16 +89,18 @@ function _rebuild(x, off, flat::AbstractVector, len = length(flat); walk = _trai
...
+  ProjectTo(y)(reshape(flat[off.i .+ (1:length(y))], axes(y)))  # ProjectTo is just correcting eltypes
 
 function _trainable_biwalk(f, x, aux)
   ch, re = functor(typeof(x), x)
-  au, _ = functor(typeof(x), aux)
+  au = _aux_children(aux)
   _trainmap(f, ch, _trainable(x), au) |> re
 end
 
+_aux_children(off) = functor(off)[1]
```
mcabbott, line 102: I think we can delete `_aux_children` since it no longer does anything special, and just have `au, _ = functor(aux)` in the body.

src/destructure.jl
```diff
@@ -103,13 +109,14 @@ end
...
   ch, re = functor(typeof(x), x)
-  au, _ = functor(typeof(x), aux)
+  au = _aux_children(aux)
   y = _trainmap(f, ch, _trainable(x), au)
   y isa Tuple{} && return NoT
   p = ProjectTo(x)
   if p isa ProjectTo  # e.g. Array, NamedTuple
     p(y)
   else  # p === identity for unknown structs
+    y = backing(re(y)) # extract NamedTuple backing from re(y); required if x has children which aren't its own fields
```
mcabbott, line 119: Note to self, this I need to think about. Some of this complication was working around things that are now fixed in CRC.jl, if I remember right.

mcabbott, line 119: Note to self, this I need to think about. Some of this complication was working around things that are now fixed in CRC.jl, if I remember right.

src/destructure.jl
jondeuce: Will do

src/destructure.jl
jondeuce: Yeah, admittedly this line took some trial and error and is a little bit above my pay-grade. I managed to convince myself, but perhaps there's something cleaner.

mcabbott: Ok, I think I finally understand what's going on. Sorry it took a while.

`re` constructs another Skip containing the gradient, and `backing` turns that into a NamedTuple with the same field names, which is what Tangent wants. 

The only way I can see this failing is this: If the primal type's constructor is fussy about what types it can accept, then it may not be happy to accept something which is valid as its gradient. E.g. if there is only `Skip(::AbstractLayer)`, and `re` tries to make one with a `Tangent`.

jondeuce: No worries! Yes, I struggled with that edge case too. Unfortunately I think it's quite tricky to work around. For example, suppose you have a user-defined `functor(m::MyModel) = (m.w,), w -> ...`. Then:

1. In general there's no way to reconstruct `MyModel` (or even a `NamedTuple` of fields/values) without `re`, as you do not know the corresponding field name given only `(m.w,)`, but
2. As you say, if the primal constructor isn't sufficiently generic then it won't be able to store `Tangent`/`Nothing`/etc. values in it's fields and will error before `backing` can unpack it again

Avoiding `re` would be ideal, but I think that would require `functor` to always return `NamedTuple`s on custom structs. I noticed that this [is the default in `@functor`](https://github.com/FluxML/Functors.jl/blob/0cf7942bd3f3a00bdb3179d74eaec17209b6abcd/src/functor.jl#L15), though, so maybe it's not such a painful requirement? In the mean time I can at least add a branch that would avoid `re` for structs that are `functor`ed to `NamedTuple`s.

mcabbott: In fact there's another problem I didn't spot before, what a mess:
```julia
julia> ac = TwoThirds([1.0, 2.0], [3.0], [4.0, 5.0]);  # from tests: a,c are functor-ed, and only a is trainable

julia> v2, re2 = destructure(ac)
([1.0, 2.0], Restructure(TwoThirds, ..., 2))

julia> gradient(ac) do x  # with Tangent{typeof(x), typeof(y)}(y)
             w2, _ = destructure(x)
             w2[2]^2
           end
((a = [0.0, 4.0], b = nothing, c = [4.0, 5.0]),) 

# Same, with z = backing(re(y)) :
julia> gradient(ac) do x
             w2, _ = destructure(x)
             w2[2]^2
           end
┌ Info: last case
│   x = TwoThirds([1.0, 2.0], [3.0], [4.0, 5.0])
│   y = (a = [0.0, 4.0], c = [4.0, 5.0])
└   z = NamedTuple{(:a, :b, :c), Tuple{Any, Any, Any}}(([0.0, 4.0], [3.0], [4.0, 5.0]))
((a = [0.0, 4.0], b = [3.0], c = [4.0, 5.0]),)
```

jondeuce: Oh yikes. That's a good example, hits all the pain points at once. If I'm understanding correctly, the gradient should be `((a = [0.0, 4.0], b = nothing, c = nothing),)`, right?

I think the problem is the `_trainmap` above; it populates the `nothing` values from `_trainable` (non-trainable fields) with the primal values, when they should be `NoT`. That's how the `b` and/or `c` values get back in there.

mcabbott: Yes, I think `_trainmap` needs to do something `isnothing(t) ? NoT : f(t, a)` here. That's where `c = [4.0, 5.0]` is coming from.

But `b = [3.0]` is coming from this PR's trick of calling the reconstructor made by `@functor`:
```
julia> ch, re = Functors.functor(ac)
((a = [1.0, 2.0], c = [4.0, 5.0]), var"#1#2"{TwoThirds}(TwoThirds([1.0, 2.0], [3.0], [4.0, 5.0])))

julia> re((a = [10, 20], c = nothing))
TwoThirds([10, 20], [3.0], nothing)
```

jondeuce: Gotcha. So on top of the modified `_trainmap` to fix `c`, one would still have to filter `backing(re(y))` to replace repopulated primal values which aren't `functor`-ed with `NoT` in order to fix `b`.

But, based on the output of `Tangent{typeof(x), typeof(y)}(y)`, maybe the modified `_trainmap` alone would be enough and `backing(re(y))` isn't needed after all...

jondeuce: Alright pushed something that works for both `Skip` and your `TwoThirds` example (modified `_trainmap` + filtering `backing(re(y))`). But since it uses `re` it would still fail for fussy constructors.

PR #78, opened 2022-05-20
Replace ADAM with Adam etc

mcabbott:
I don't know where this package (and Flux) got the all-capital ADAM from, but the paper calls it Adam. It also has seemingly arbitrary pairs like "ADAGrad, AdaMax". So this PR proposes to standardise on Julia-like capitalisations. 

The old ones are marked `@deprecate` which is noisy only in tests, and should still export them. 

The tests are for now untouched.

Xref https://github.com/FluxML/Flux.jl/issues/1949


CarloLucibello (approved):

mcabbott:
This broke the docs, unfortunately:

Screenshot 2022-05-24 at 22 43 21

Issue #76, opened 2022-05-13
Applying `Grads` fails


femtomc:
Simple MWE:

```julia
module GradsApplyMWE

using Optimisers
using Flux

model = Chain(Dense(50, 5))

state = Optimisers.setup(Optimisers.ADAM(), model)

data = rand(Float32, (50, 1))

grads = Flux.gradient(Flux.params(model)) do
    sum(model(data))
end

state, model = Optimisers.update(state, model, grads)

end # module
```

Throws:
```
ERROR: LoadError: type Grads has no field layers
Stacktrace:
  [1] getproperty(x::Zygote.Grads, f::Symbol)
    @ Base ./Base.jl:33
  [2] functor(#unused#::Type{Flux.Chain{Tuple{Flux.Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}}, x::Zygote.Grads)
    @ Flux ~/.julia/packages/Functors/qBIlC/src/functor.jl:23
  [3] (::Optimisers.var"#4#6"{Flux.Chain{Tuple{Flux.Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}})(x̄::Zygote.Grads)
    @ Optimisers ~/.julia/packages/Optimisers/pCISx/src/interface.jl:37
  [4] map
    @ ./tuple.jl:213 [inlined]
  [5] update!(tree::NamedTuple{(:layers,), Tuple{Tuple{NamedTuple{(:weight, :bias, :σ), Tuple{Optimisers.Leaf{Optimisers.ADAM{Float32}, Tuple{Matrix{Float32}, Matrix{Float32}, Tuple{Float32, Float32}}}, Optimisers.Leaf{Optimisers.ADAM{Float32}, Tuple{Vector{Float32}, Vector{Float32}, Tuple{Float32, Float32}}}, Nothing}}}}}, x::Flux.Chain{Tuple{Flux.Dense{typeof(identity), Matrix{Float32}, Vector{Float32}}}}, x̄s::Zygote.Grads)
    @ Optimisers ~/.julia/packages/Optimisers/pCISx/src/interface.jl:37
[...]

PR #80, opened 2022-05-27
Document the need for explicit gradients

mcabbott:
Xref https://github.com/FluxML/Optimisers.jl/issues/76

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -63,7 +63,7 @@ tree formed by the model and update the parameters using the gradients.
 Optimisers.jl does not depend on any one automatic differentiation package,
 but for now the most likely source of gradients is [Zygote.jl](https://fluxml.ai/Zygote.jl).
 Note that `update` always wants the "explicit" gradient as shown above, a tree structure, 
-rahter than the dictionary-like objects from the "implicit" style `gradient(() -> loss(...), params(model))`. 
+rather than the dictionary-like objects from the "implicit" style `gradient(() -> loss(...), params(model))`. 
 (See  [Zygote's documentation](https://fluxml.ai/Zygote.jl/dev/#Explicit-and-Implicit-Parameters-1) for more.)
 
 There is also `Optimisers.update!` which similarly returns a new model and new state,
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -84,19 +84,25 @@ identical trees of nested `NamedTuple`s.)
 
 ```julia
 
-using Lux  # https://github.com/avik-pal/Lux.jl#getting-started
+using Lux, Boltz, Zygote, Optimisers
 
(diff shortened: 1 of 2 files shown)


docs/src/index.md
```diff
@@ -58,9 +58,13 @@ st, model = Optimisers.update(st, model, m̄);
...
 is handled by [Functors.jl](https://fluxml.ai/Functors.jl), where we do a walk over the
-tree formed by the model and update the parameters using the gradients. Optimisers can
-work with different forms of gradients, but most likely use case are the gradients as
-returned by [Zygote.jl](https://fluxml.ai/Zygote.jl).
+tree formed by the model and update the parameters using the gradients.
+
+Optimisers.jl does not depend on any one automatic differentiation package,
+but for now the most likely source of gradients is [Zygote.jl](https://fluxml.ai/Zygote.jl).
+Note that `update` always wants the "explicit" gradient as shown above, a tree structure, 
+rahter than the dictionary-like objects from the "implicit" style `gradient(() -> loss(...), params(model))`.
```
darsnack, line 66:
```suggestion
rather than the dictionary-like objects from the "implicit" style `gradient(() -> loss(...), params(model))`. 
```

docs/src/index.md
```diff
@@ -58,9 +58,13 @@ st, model = Optimisers.update(st, model, m̄);
...
 Notice that a completely new instance of the model is returned. Internally, this
 is handled by [Functors.jl](https://fluxml.ai/Functors.jl), where we do a walk over the
-tree formed by the model and update the parameters using the gradients. Optimisers can
-work with different forms of gradients, but most likely use case are the gradients as
-returned by [Zygote.jl](https://fluxml.ai/Zygote.jl).
+tree formed by the model and update the parameters using the gradients.
+
+Optimisers.jl does not depend on any one automatic differentiation package,
+but for now the most likely source of gradients is [Zygote.jl](https://fluxml.ai/Zygote.jl).
+Note that `update` always wants the "explicit" gradient as shown above, a tree structure,
```
ToucheSir, line 65:
```suggestion
Note that `update` always wants the "explicit" gradient as a tree structure like shown above, 
```

mcabbott:
Latest commit bolts on a section showing Lux.jl usage. The immediate point is to try to clarify the various meanings that "explicit" seems to have around here. Maybe such a section should grow further, later?

docs/src/index.md
mcabbott: See what you think of new wording. I want to say what you *do*, up front. And then discuss what objects this returns, afterwards.

theabhirath:
Could we also document which of these (if any) have to be moved to the GPU in case of training on a GPU? That was something that was a little confusing when I first used this package 😅

darsnack (approved):

Issue #84, opened 2022-06-02
doc improvement: working with custom model types


CarloLucibello:
We should give more prominence in the docs to the usage of `Flux.@functor` and `Optimisers.trainable` to define trainable parameters of custom types.

mcabbott:
Yes. Where's the best place? 

I don't want every docstring to become a novel, but perhaps the one for `Optimisers.setup` is one good place to show that this recurses into custom structs.

The main doc page uses a Flux model as the first example of applying a rule. It could instead first construct some simpler model, and show that a newly written optimisation rule is applied to all children etc?

CarloLucibello:
this seems done

PR #85, opened 2022-06-02
Extend `setup` docstring to show custom struct

mcabbott:
For #84

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -70,26 +70,27 @@ julia> Optimisers.setup(Momentum(), m)  # same field names as m
 ```
 
 The recursion into structures uses Functors.jl, and any new `struct`s containing parameters
-need to be marked with `Functors.@functor` before use. Further refinements can, if necessary,
-be made by overloading [`trainable`](@ref).
+need to be marked with `Functors.@functor` before use.
+See [the Flux docs](https://fluxml.ai/Flux.jl/stable/models/advanced/) for more about this.
 
 ```jldoctest
-julia> struct Layer; mat; vec; end
+julia> struct Layer; mat; fun; end
 
-julia> model = (lay = Layer(rand(2, 2), rand(2)), fun = sin);
+julia> model = (lay = Layer([1 2; 3 4f0], sin), vec = [5, 6f0]);
 
 julia> Optimisers.setup(Momentum(), model)  # new struct is by default ignored
-(lay = nothing, fun = nothing)
+(lay = nothing, vec = Leaf(Momentum{Float32}(0.01, 0.9), [0.0, 0.0]))
+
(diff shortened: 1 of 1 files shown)


src/Optimisers.jl
```diff
@@ -63,8 +63,33 @@ or [`update!`](@ref).
...
-julia> Optimisers.setup(Descent(0.1f0), (x = rand(3), y = (true, false), z = tanh))
-(x = Leaf(Descent{Float32}(0.1), nothing), y = (nothing, nothing), z = nothing)
+julia> m = (x = rand(3), y = (true, false), z = tanh);
+
+julia> Optimisers.setup(Momentum(), m)  # same field names as m
+(x = Leaf(Momentum{Float32}(0.01, 0.9), [0.0, 0.0, 0.0]), y = (nothing, nothing), z = nothing)
+```
+
+The recursion into structures uses Functors.jl, and any new `struct`s containing parameters
+need to be marked with `Functors.@functor` before use. Further refinements can, if necessary,
```
ToucheSir, line 73:
```suggestion
need to be marked with `Functors.@functor` before use.
See [the Flux docs](https://fluxml.ai/Flux.jl/stable/models/advanced/#Customising-Parameter-Collection-for-a-Model) for more guidance on how to do this for layers specifically.
Further refinements can, if necessary,
```
I was going to link to `@functor` too, but looks like we need a new Flux docs build for that to render.

src/Optimisers.jl
```diff
@@ -63,8 +63,35 @@ or [`update!`](@ref).
...
+
+julia> Optimisers.setup(Momentum(), model)  # new struct is by default ignored
+(lay = nothing, fun = nothing)
+
+julia> using Functors; @functor Layer  # annotate this type as containing parameters
+
+julia> Optimisers.setup(Momentum(), model)
+(lay = (mat = Leaf(Momentum{Float32}(0.01, 0.9), [0.0 0.0; 0.0 0.0]), vec = Leaf(Momentum{Float32}(0.01, 0.9), [0.0, 0.0])), fun = nothing)
+
+julia> Optimisers.trainable(l::Layer) = (vec = l.vec,)  # if necessary, ignore some parameters
```
ToucheSir, line 91: My inclination is that we'd want a motivating example where `@functor Layer (vec,)` is not sufficient. The counter-argument would be that 99% of `@functor` use should not specify a list of properties because users may want to traverse to them with `fmap` and friends.

mcabbott: Maybe best to leave this off here?

CarloLucibello (approved):

src/Optimisers.jl
mcabbott: Done.

PR #86, opened 2022-06-03
Restrict the types of learning rates

mcabbott:
As discussed starting here https://github.com/FluxML/Optimisers.jl/pull/79#issuecomment-1146022257 

This PR weakens the constraint for gradient descent only to `Descent{T<:Number}`. That rule is so simple that perhaps it should literally do what the formula says, and apparently there are papers using this with complex η.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -22,6 +22,7 @@ Leaf(rule, state; frozen::Bool = false) = Leaf(rule, state, frozen)
 Base.:(==)(a::Leaf, b::Leaf) = children(a) == children(b)
 
 function setup(rule::AbstractRule, model)
+  checksign(rule)
   cache = IdDict()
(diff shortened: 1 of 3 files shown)


mcabbott:
Xref also https://github.com/FluxML/Flux.jl/issues/1982. We could make negative learning rates an error. Other optimiser settings like momentum have other constraints for sensible use, and should probably then be similarly constrained. 

Adding errors by hand would catch some simple typos I guess? But this won't easily help automated search to stay within bounds, for that presumably the bounds should be encoded in a way that the searcher can understand.

ToucheSir:
For hyperparameters where we can't decide between erroring, warning or doing nothing, we could make use of https://github.com/JuliaPackaging/Preferences.jl to allow users to switch between options (ideally only two).

PR #89, opened 2022-06-07
Add `adjust`

### LINKED ISSUE #88: support a nice way of changing the learning rate
It would be nice to have a easy and rule independent way of changing the learning rate like in Flux.jl.  

Right now the best way i can think of is making another state tuple while changing the first argument but then changing the learning rate is not rule independent.

@mcabbott  suggested
> Although perhaps there ought to be one, like `st = setup(rule, st)` without zero-ing the momenta. Maybe make an issue?

related: #15 (seems to be not active)

mcabbott:
Closes #88, by adding making `setup(Adam(), state)` and `setup(0.02, state)` re-build the state tree with different parameters. This would have been trivial before #30 changed `update!(o, state, x, x̄)` to `update!(tree, x, x̄)`.

Right now the new `adjust` is like `init`, acts on one thing, while `setup` & `update` act on the whole model/tree. Could avoid the name by making them callable, like `Adam()(0.02)`, which perhaps some earlier version had. 

At the moment many things are errors, such as mismatched optimisers, and rules without a field `:eta` to change. It's possible that they should instead do nothing, so that only some leaves are changed, others left alone.

Could easily be generalised to say `setup(state; eta=0.02)` allowing you to set any field of all rules. If desired.

### DIFF
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -56,9 +56,10 @@ function _adjust(ℓ::Leaf, a, ok::Ref)
 end
 
 """
-    Optimisers.adjust(rule::RuleType, η::Real) -> rule, flag
+    Optimisers.adjust(rule::RuleType, η::Real) -> rule, changed
 
 Replaces the learning rate of the optimisation rule with the given number.
+
 This method is only called by `adjust(tree, η)`, and if `RuleType` has a field
 called `:eta` and the default constructor, then the standard definition will work.
 It is only necessary to provide a method if your `struct` stores its learning rate
@@ -66,12 +67,18 @@ in a different field.
 
 # Example
 ```jldoctest
+julia> Optimisers.adjust(Adam(), 0.12345)
+(Adam{Float32}(0.12345f0, (0.9f0, 0.999f0), 1.1920929f-7), true)
+
 julia> struct DecayDescent{T} <: Optimisers.AbstractRule  # as in the documentation
          eta::T
        end
 
-julia> Optimisers.adjust(DecayDescent(0.1f0), 0.23)  # works automatically
-(DecayDescent{Float32}(0.23f0), true)
+julia> Optimisers.adjust(DecayDescent(0.1f0), 0.2345)  # works automatically
+(DecayDescent{Float32}(0.2345f0), true)
+
+julia> Optimisers.adjust(ClipNorm(), 0.345)  # does nothing, as this has no learning rate
+(ClipNorm{Float32}(10.0f0, 2.0f0, true), false)
 ```
 """
 function adjust(r::T, η::Real) where T <: AbstractRule
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -1,13 +1,12 @@
 
 """
-    Optimisers.adjust(tree, adj) -> tree
+    Optimisers.adjust(tree, η) -> tree
 
 Alters the state `tree = setup(rule, model)` to change the parameters of the
 optimisation rule, without destroying its stored state. Typically used mid-way
 through training.
-* To change just the learning rate, provide a number `adj::Real`.
-* To change all parameters, provide a new rule `adj::AbstractRule`.
-  (This will affect only leaves of the same type.)
+
+To change just the learning rate, provide a number `η::Real`.
 
 # Example
 ```jldoctest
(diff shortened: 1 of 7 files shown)


src/rules.jl
```diff
@@ -6,6 +7,24 @@
...
 @deprecate ADADelta AdaDelta
 
+# Generic adjust methods, called by setup(rule, tree) or by setup(eta, tree).
+# Some more specific ones are defined below.
+function adjust(newr::AbstractRule, r::AbstractRule)
+  typeof(newr).name.wrapper == typeof(r).name.wrapper || throw(ArgumentError(
+    "adjust(r′, r) expects the same rule with different parameters"))
+  newr
+end
+function adjust(η::Real, r::T) where T <: AbstractRule
```
mcabbott, line 17: Possibly this is backwards: `init` and `apply!` take the rule as the first argument, maybe this should follow?

ToucheSir:
Can we take this opportunity to make a getter/setter pair for the learning rate? The default implementation could look for `eta`, or it could be an empty interface method.

src/rules.jl
mcabbott: But which way around the other method goes is stranger, `adjust(r::AbstractRule, newr::AbstractRule)` clashes a bit with `setup(newr::AbstractRule, r::Leaf)`?

mcabbott:
This `adjust` does work as a setter (not sure it has the right name). What would you like a getter to return when there is no field `:eta`, just an error until defined?
```julia
julia> Optimisers.adjust(Adam(), 0.123)
Adam{Float32}(0.123f0, (0.9f0, 0.999f0), 1.1920929f-7)

julia> Adam().eta
0.001f0
```

ToucheSir:
I was thinking we might as well be super explicit and call it `learning_rate`/`learningrate`.

darsnack:
I'm worried about how overloaded `setup` becomes with this change. Can we have
- `adjust(o::AbstractRule, lr)`: interface to overload with the appropriate default for now
- `adjust(l::Leaf, lr)`: calls `adjust(l.rule, lr)`
- `adjust(tree, lr)`: walks to each leaf and calls `adjust`

Personally, I think we should generalize to any field from the get go. The above should really be `adjust(f, tree/leaf/rule)` where `f: AbstractRule -> AbstractRule` returns a new adjusted rule. Then we provide a convenient LR setter that can be passed in for `f`. This makes the mechanism open to any field while prioritizing the LR as the one people want.

mcabbott:
I wrote it this way around at first, before realising it could share the recursion with `setup`, since `tree` only ever contains `Leaf` and `nothing`. Not yet sure which I prefer. As you say this overloads `setup` to re-setup. It maintains a distinction between whole-tree functions and one-array functions; but unlike `apply!` there is seldom going to be a need to write your own method. (The one-array function could also be made to disappear by making optimisation rules callable, but not sure I like that.)

Re functionality, it would be easy to make `alter(tree; eta = 0.1)` change a named parameter. (And that argues for `alter(tree, eta::Real)` order, not the reverse.) What should `alter(tree; beta = (0,0))` do to leaves (or a tree) where nobody has `beta`, maybe it should complain?

With `alter(f::Function, tree)`, should this `f(::AbsRule)::AbsRule` be restricted to return the same outermost type of rule, to guarantee that the stored state still makes sense? If so, I wonder what you gain over `alter(tree, rule)`. At the moment there is no way to tell `setup` to (say) use a different rule for e.g. the vectors, or the larger arrays. I think we thought about allowing some `setup(f::Function, model)` at some point, but (IIRC) the question is whether `f` should just get each array, or should have some context, such as getting the field name `:bias` of the array as another argument. But such info would not be available to `alter(f, tree)`.

ToucheSir:
I think the difficulty of coming up with a general API like this is an argument for an `alter(lens, tree)`. `lens` doesn't necessarily have to be an actual lens object matching the Accessors.jl interface, but something in the spirit of that.

src/adjust.jl
```diff
@@ -0,0 +1,108 @@
...
+end
+
+"""
+    Optimisers.adjust(rule::RuleType, η::Real) -> rule, flag
+
+Replaces the learning rate of the optimisation rule with the given number.
+This method is only called by `adjust(tree, η)`, and if `RuleType` has a field
+called `:eta` and the default constructor, then the standard definition will work.
+It is only necessary to provide a method if your `struct` stores its learning rate
+in a different field.
```
ToucheSir, line 65: Perhaps we want a sentence briefly describing what `flag` means?

mcabbott: Indeed. 

This flag story is a little bit ugly. The method you may need to write for `adjust(rule::RuleType, η::Real)` doesn't work the same way as the one you call `adjust(tree, eta)` just with a trivial tree. Instead, it has this extra return. 

The only purpose is so that you get a friendly `Warning: adjust did not find any rules to act on!` if nothing matches. Maybe that's not worth doing?

src/adjust.jl
```diff
@@ -0,0 +1,108 @@
...
+#     vals = map(fs) do field
+#       get(nt, field, getfield(r, field))
+#     end
+#     T(vals...), true  # relies on having the default constructor
+#   else
+#     r, false
+#   end
+# end
+
+function adjust(oldr::AbstractRule, newr::AbstractRule)
```
ToucheSir, line 102: After a proper read-through, I feel like also calling this method `adjust` is a tad confusing. Could we just focus on learning rate adjustment for this PR and tackle whole rule replacement in a follow-up?

mcabbott: Ok, I agree. This is gone, new version is simpler...

src/adjust.jl
mcabbott: New version has no flag. This means you can silently change nothing, but that's OK. Usually you change the activation rate, and everyone has one of those.

ToucheSir (approved):

PR #93, opened 2022-06-30
Implement Rprop

MilkshakeForReal:
https://ieeexplore.ieee.org/document/298623

https://pytorch.org/docs/stable/generated/torch.optim.Rprop.html#torch.optim.Rprop

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -164,20 +164,18 @@ init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
 
 function apply!(o::Rprop, state, x, dx)
     ℓ, Γ = o.ell, o.gamma
-    g₀, η₀ = state
+    g, η = state
 
-    signs = g₀ .* dx
+    signs = sign.(g .* dx)
     signs[signs .> 0] .= ℓ[2]
     signs[signs .< 0] .= ℓ[1]
     signs[signs .== 0] .= one(eltype(signs))
 
-    η₁ =  clamp.(η₀ .* signs, Γ[1], Γ[2])
+    @.. η =  clamp(η * signs, Γ[1], Γ[2])
+    @.. g = (signs !== ℓ[1]) * dx
 
-    @.. g = (signs == ℓ[1]) * dx
-
-    dx′ = @lazy η₁ * sign(g₁)
-
-    return (g₁, η₁), dx′
+    dx′ = @lazy η * sign(g)
+    return (g, η), dx′
 end
 
 """
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -141,7 +141,7 @@ end
     Rprop(η = 1f-3, ℓ = (5f-1, 1.2f0), Γ = (1f-6, 50f0))
 
 Optimizer using the
-[Rprop](https://ieeexplore.ieee.org/document/298623) algorithm. An full-batch
+[Rprop](https://ieeexplore.ieee.org/document/298623) algorithm. A full-batch
 learning algorithm that depends only on the sign of the gradient.
 
 # Parameters
@@ -166,13 +166,10 @@ function apply!(o::Rprop, state, x, dx)
     ℓ, Γ = o.ell, o.gamma
     g, η = state
 
-    signs = broadcast(g, dx) do g, dx
-        s = sign(g * dx)
-        s > 0 ? ℓ[2] : s < 0 ? ℓ[1] : one(eltype(dx))
+    T = eltype(dx)
+    g, η = broadcast(g, η, dx) do g, η, dx
+        g * dx > 0 ? (g, min(η * ℓ[2], Γ[2])) : g * dx < 0 ? (zero(T), max(η * ℓ[1], Γ[1])) : (g, η)
     end
-
-    @.. η =  clamp(η * signs, Γ[1], Γ[2])
-    @.. g = (signs !== ℓ[1]) * dx
 
     dx′ = @lazy η * sign(g)
     return (g, η), dx′
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -166,9 +166,8 @@ function apply!(o::Rprop, state, x, dx)
     ℓ, Γ = o.ell, o.gamma
     g, η = state
 
-    T = eltype(dx)
     g, η = broadcast(g, η, dx) do g, η, dx
-        g * dx > 0 ? (g, min(η * ℓ[2], Γ[2])) : g * dx < 0 ? (zero(T), max(η * ℓ[1], Γ[1])) : (g, η)
+        g * dx > 0 ? (g, min(η * ℓ[2], Γ[2])) : g * dx < 0 ? (zero(dx), max(η * ℓ[1], Γ[1])) : (g, η)
     end
 
     dx′ = @lazy η * sign(g)
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -181,7 +181,6 @@ function apply!(o::Rprop, state, x, dx)
     return (g₁, η₁), dx′
 end
 
-
 """
     Adam(η = 1f-3, β = (9f-1, 9.99f-1), ϵ = eps(typeof(η)))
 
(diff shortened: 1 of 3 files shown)


src/rules.jl
```diff
@@ -135,6 +136,51 @@ function Base.show(io::IO, o::RMSProp)
...
+
+    signs = g₀ .* dx
+    signs[signs .> 0] .= ℓ[2]
+    signs[signs .< 0] .= ℓ[1]
+    signs[signs .== 0] .= one(eltype(signs))
+
+    η₁ =  clamp.(η₀ .* signs, Γ[1], Γ[2])
+
+    g₁ = copy(dx)
+    g₁[signs .== ℓ[1]] .= zero(eltype(g₁))
```
mcabbott, line 177: I think you can probably write this as follows:
```suggestion
    @.. g = (signs == ℓ[1]) * dx
```
First, this avoids making and then mutating `g₁`, since it might be an immutable array. (For now, tested with StaticArrays.) 

Second, less importantly, `@..` writes back into the same `g` if it is in fact mutable, otherwise binds `g` to the RHS.

MilkshakeForReal:
`Rprop` is not suitable for mini-batch training. For sure it will fail this test set
https://github.com/FluxML/Optimisers.jl/blob/master/test/rules.jl#L42

MilkshakeForReal:
As expected, changing `x = rand(10)` to `x = rand(10,20)` would pass all the tests

MilkshakeForReal:
To be specific, it is the `iRprop-` algorithm. See Algorithm 4 in http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.17.1332&rep=rep1&type=pdf

src/rules.jl
```diff
@@ -135,6 +136,47 @@ function Base.show(io::IO, o::RMSProp)
...
+
+init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
+
+function apply!(o::Rprop, state, x, dx)
+    ℓ, Γ = o.ell, o.gamma
+    g, η = state
+
+    signs = map(s-> s > 0 ? ℓ[2] :
+                    s < 0 ? ℓ[1] : one(eltype(dx)),
+                sign.(g .* dx))
```
ToucheSir, line 171:
```suggestion
    signs = map(g, dx) do g, dx
        s = sign(g * dx)
        s > 0 ? ℓ[2] : s < 0 ? ℓ[1] : one(eltype(dx))
    end
```
Manually fusing these seems about 2x faster for me on CPU and GPU, while saving on an intermediate allocation from `sign.(g .* dx)`.

test/rules.jl
```diff
@@ -39,7 +39,7 @@ end
     @test iloss(rand(10, 10), w, w′) > 1
     st = Optimisers.setup(o, w)
     for t = 1:10^5
-      x = rand(10)
+      x = rand(10,20)
```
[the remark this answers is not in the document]
MilkshakeForReal: I just wanted to trigger the test, not sure if this change still follows the motivation behind the testset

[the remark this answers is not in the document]
ToucheSir: Given that they all still pass, I think so? I thought this was required to get RProp to pass as well.

src/rules.jl
```diff
@@ -135,6 +136,48 @@ function Base.show(io::IO, o::RMSProp)
...
+
+Rprop(η = 1f-3, ℓ = (5f-1, 1.2f0), Γ = (1f-6, 50f0)) = Rprop{typeof(η)}(η, ℓ, Γ)
+
+init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
+
+function apply!(o::Rprop, state, x, dx)
+    ℓ, Γ = o.ell, o.gamma
+    g, η = state
+
+    signs = map(g, dx) do g, dx
```
ToucheSir, line 169: I'm not exactly sure where the latest test error is coming from, but try this:
```suggestion
    signs = broadcast(g, dx) do g, dx
```

src/rules.jl
```diff
@@ -135,6 +136,48 @@ function Base.show(io::IO, o::RMSProp)
...
+        s = sign(g * dx)
+        s > 0 ? ℓ[2] : s < 0 ? ℓ[1] : one(eltype(dx))
+    end
+
+    @.. η =  clamp(η * signs, Γ[1], Γ[2])
+    @.. g = (signs !== ℓ[1]) * dx
```
mcabbott, line 175: Also, why test `===` here, not `==`? Or are the types guaranteed to match?
```suggestion
    @.. g = (signs != ℓ[1]) * dx
```

src/rules.jl
mcabbott: Seems concerning that `g` and `dx` can end up different sizes. How is that possible?

src/rules.jl
MilkshakeForReal: Because we just assigned `ℓ[1]` to `signs`?

MilkshakeForReal: I just feel safer this way

src/rules.jl
```diff
@@ -135,6 +136,48 @@ function Base.show(io::IO, o::RMSProp)
...
+
+init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
+
+function apply!(o::Rprop, state, x, dx)
+    ℓ, Γ = o.ell, o.gamma
+    g, η = state
+
+    signs = map(g, dx) do g, dx
+        s = sign(g * dx)
+        s > 0 ? ℓ[2] : s < 0 ? ℓ[1] : one(eltype(dx))
```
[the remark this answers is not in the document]
MilkshakeForReal: I don't know, can the learning rate have a different type from `dx`?

src/rules.jl
mcabbott: Yes that's better. But this should not allocate anything, as it's one fused broadcast.

src/rules.jl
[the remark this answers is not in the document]
MilkshakeForReal: If yes then we should not take the type from `Rprop{T}`

test/rules.jl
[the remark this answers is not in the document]
MilkshakeForReal: Yes. Just not so sure what "independence", the name of the testset, means

MilkshakeForReal:
I don't know how to return two named tuple :(
```julia
julia> a
(data = 2, prop = 4)

julia> b
(data = 3, prop = 5)

julia> c
(data = 20, prop = 40)

julia> d,h = map((i,j,k)-> (i*j*k,zero(k)), a, b,c)
(data = (120, 0), prop = (800, 0))

julia> d
(120, 0)

julia> h
(800, 0)
```

src/rules.jl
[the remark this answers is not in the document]
mcabbott: Yes, here the model, the gradient, and the optimiser have 3 different types: 
```
julia> m = (x = Float32[1,2,3],);

julia> st = Optimisers.setup(Momentum(0.1), m)  # Float64 learning rate
(x = Leaf(Momentum{Float64}(0.1, 0.9), Float32[0.0, 0.0, 0.0]),)

julia> st, m = Optimisers.update(st, m, (x = 4:6,));  # integer gradients a bit artificial, won't occur

julia> m
(x = Float32[0.6, 1.5, 2.4],)
````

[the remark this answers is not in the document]
Comment (MilkshakeForReal) [reply src/rules.jl]: So I guess it means we should take the type from the parameters or the gradients?

File: src/rules.jl [hunk 7]
```diff
@@ -135,6 +136,49 @@ function Base.show(io::IO, o::RMSProp)
...
+    g, η = state
+
+    η = broadcast(g, η, dx) do g, η, dx
+        g * dx > 0 ? min(η * ℓ[2], Γ[2]) : g * dx < 0 ? max(η * ℓ[1], Γ[1]) : η
+    end
+
+    g = broadcast(g, dx) do g, dx
+        g * dx < 0 ? zero(dx) : dx
+    end
+
```
ToucheSir, line 176:
```suggestion
    g = broadcast(g, dx) do g, dx
        g * dx < 0 ? zero(dx) : dx
    end
```

ToucheSir (approved):
Just a tiny bit of whitespace. Otherwise changes look good to me, doctest failure seems unrelated.

ToucheSir:
Thanks!

PR #98, opened 2022-07-07
Fix the doctests and the broken CI

Saransh-cpp:
- adds colored output to GitHub Actions to diagnose failures easily
- doctests now run only once
- fixes some broken doctests
- adds the dependencies required to run doctests


mcabbott (approved):
Looks good, thanks!

PR #100, opened 2022-07-10
Transparent handling of tied weights

ToucheSir:
This makes `Leaf` a mutable type so that tied weights are represented by the same leaf instance.

Although only mutable array types are automatically detected as tied, one can also tie immutable parameters by manually creating shared `Leaf`s.

The test suite is practically the same as #42, with some slight modifications since there is no equivalent to `Tied` in this PR.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -6,24 +6,27 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
 
 abstract type AbstractRule end
 
-struct Leaf{R,S}
+mutable struct Leaf{R,S}
   rule::R
   state::S
 end
 
-function setup(rule, x; seen = Base.IdSet())
+function setup(rule, x; cache = IdDict{Any,Leaf}())
   rule isa AbstractRule || Base.depwarn("In future, all optimisation rules should be <: AbstractRule", :setup)
   if isnumeric(x)
-    x in seen && throw(ArgumentError("Optimisers.jl does not at present handle tied weights, sorry."))
-    isbits(x) || push!(seen, x)
-    return Leaf(rule, init(rule, x))
+    leaf = get(cache, x, missing)
+    ismissing(leaf) || return leaf
+    leaf = Leaf(rule, init(rule, x))
+    isbits(x) || (cache[x] = leaf)
+    return leaf
   elseif isleaf(x)
     return nothing
   else
-    return map(xᵢ -> setup(rule, xᵢ; seen), _trainable(x))
+    return map(xᵢ -> setup(rule, xᵢ; cache), _trainable(x))
   end
 end
 
+_add!(x, x̄) = iswriteable(x) ? (x .= x .+ x̄) : eltype(x).(x .+ x̄)
 subtract!(x, x̄) = iswriteable(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
 
 update!(::Nothing, x, ::Zero, ::Zero...) = nothing, x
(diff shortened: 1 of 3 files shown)


src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
...
 function update!(tree, x, x̄s...)
+  cache = IdDict{Leaf,Any}()
+  _accumulate!(cache, tree, x, x̄s...)
+  return UpdateCallback(cache, IdDict{Leaf,Any}())(tree, x, x̄s...)
+end
+
+_accumulate!(::AbstractDict{Leaf,Any}, ::Nothing, _, _...) = nothing
+_accumulate!(::AbstractDict{Leaf,Any}, ::Nothing, _, ::Zero, ::Zero...) = nothing
+_accumulate!(::AbstractDict{Leaf,Any}, ℓ::Leaf, _, ::Zero, ::Zero...) = nothing
+_accumulate!(::AbstractDict{Leaf,Any}, _, _, ::Zero, ::Zero...) = nothing
```
ToucheSir, line 51: There are a lot of overloads with high degrees of overlap across multiple functions. I couldn't think of a way to deduplicate some of them, so if anyone has ideas that would be swell.

src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
...
-  t′ = fmap(copy, tree; exclude = iswriteable)
-  x′ = fmap(copy, x; exclude = iswriteable)
-  update!(t′, x′, x̄s...)
+  # because we rely on Leaf identity for tied parameters, they require special treatment
+  cache = IdDict()
+  tree′ = fmap(tree; cache, exclude = Base.Fix2(isa, Leaf)) do ℓ
+    Leaf(ℓ.rule, fmap(copy, ℓ.state; cache, exclude = iswriteable))
+  end
+  x′ = fmap(copy, x; cache = empty!(cache), exclude = iswriteable)
+  x̄s′ = fmap(copy, x̄s; cache = empty!(cache), exclude = iswriteable)
```
ToucheSir, line 92: It turns out we were not defensively copying state or gradients before, so they could still be mutated by a call to `update`.

ToucheSir:
Doctests appear to be picking up changes on master that aren't present on this branch, is that expected? I can't tweak the test because it doesn't exist here!

src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
...
+  x̄s′ = map(x̄ -> functor(typeof(x), base(x̄))[1], x̄s)
+  x′, _ = functor(typeof(x), x)
+  foreach((stᵢ, xᵢ, x̄sᵢ...) -> _accumulate!(cache, stᵢ, xᵢ, x̄sᵢ...), tree, x′, x̄s′...)
+end
+
+# slightly cleaner way of closing over update! internal state
+struct UpdateCallback
+  acc_grads::IdDict{Leaf,Any}
+  param_cache::IdDict{Leaf,Any}
+end
```
mcabbott, line 68: The limitation might be mine but I have to say I find this struct really hard to read, compared to just closing over things which have one name.

mcabbott (commented):
Ok I tried to read this, and in the process of understanding what's going on, wrote #106. Maybe that explains my thoughts more clearly than the comments here.

src/interface.jl
ToucheSir: I think I tried that, but ran into ambiguities. This is the smallest number of methods I could come up with that didn't have ambiguities. If you can narrow that down, that would be superb.

src/interface.jl
ToucheSir: I don't understand what "just closing over things which have one name." entails here, can you elaborate? Another reason for the struct over a normal closure is self-recursion, which I use [here](https://github.com/FluxML/Optimisers.jl/pull/100/files/d90e581b71a7838b13487a28cd4c87bc9c05bd57#diff-f34fad8a7ea2affe69dada61da4337de97d1a3aab78c733223fd3bdde5b18a51R81).

src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
...
+  update!(ℓ, x, pop!(cb.acc_grads, ℓ)...)
+end
+function (cb::UpdateCallback)(tree, x, x̄s...)
   x̄s′ = map(x̄ -> functor(typeof(x), base(x̄))[1], x̄s)
   x′, re = functor(typeof(x), x)
-  xtree = map((stᵢ, xᵢ, x̄sᵢ...) -> update!(stᵢ, xᵢ, x̄sᵢ...), tree, x′, x̄s′...)
-  map(first, xtree), re(map(last, xtree))
+  xtree = map(cb, tree, x′, x̄s′...)
+  return map(first, xtree), re(map(last, xtree))
 end
```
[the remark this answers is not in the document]
ToucheSir: Yes, absolutely. I held off from doing that here in case some user was stashing old state trees and would be blindsided by the values in those leaves suddenly changing.

src/interface.jl
mcabbott: I mean things like this, which define a dict & then use it:
```
   cache = IdDict{Leaf,Any}()
   _accumulate!(cache, tree, x, x̄s...)
```
With no further names: no structs, no field names.

src/interface.jl
ToucheSir: > For copying state, can't we just say `@functor Leaf (state,)` and let `fmap` do it?

That breaks `Leaf` identity, unfortunately. `fmap` will end up untying shared parameters by creating new leaves at each location during reconstruction.

mcabbott: Won't `fmap` will preserve the Leaf identifications? That's what its cache is for.

src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
...
+    return leaf
   elseif isleaf(x)
     return nothing
   else
-    return map(xᵢ -> setup(rule, xᵢ; seen), _trainable(x))
+    return map(xᵢ -> setup(rule, xᵢ; cache), _trainable(x))
   end
 end
 
+_add!(x, x̄) = iswriteable(x) ? (x .= x .+ x̄) : eltype(x).(x .+ x̄)
```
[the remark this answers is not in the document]
ToucheSir: My worry with the lazy accumulation approach is threefold. First, it blows any chance of making this type stable out the window. Secondly, it's possible the lazy `Broadcasted` may be evaluated multiple times as it passes through a chain of rules and thus incur accumulation overhead more than once. Lastly, complicated broadcasts come with a lot of compilation latency (especially on GPU) and I'm wary of making optimizers worse than they [already are](https://github.com/JuliaGPU/GPUCompiler.jl/issues/65#issuecomment-903965872) on that front.

src/interface.jl
ToucheSir: The underscores are mostly to appease the linter and possibly improve latency(??) Perhaps `::Any` would work better, though I'm not sure that addresses your point about clarity?

src/interface.jl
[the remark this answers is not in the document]
mcabbott: We can do it eagerly to avoid this. But we cannot mutate the gradients, as they may be shared with others (e.g. from the rule for +).

src/interface.jl
ToucheSir: I have a recollection that it sometimes only preserved leaves, but re-reading the code you are correct.

src/interface.jl
mcabbott: I don't see how names can affect latency. I just mean they let your eye know what the 4th argument *means*, which `::Any` doesn't help.

src/interface.jl
ToucheSir: I recall trying this first, and deciding to bundle things into a struct after seeing a lot of long, long lines from threading the two IdDicts through multiple levels of functions. It may also have been tricky to get that working in a backwards compatible way, but it's been long enough that I don't remember the whole context.

src/interface.jl
```diff
@@ -6,47 +6,91 @@ const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
 
 abstract type AbstractRule end
 
-struct Leaf{R,S}
+mutable struct Leaf{R,S}
   rule::R
   state::S
 end
```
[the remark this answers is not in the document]
ToucheSir: See https://github.com/FluxML/Optimisers.jl/pull/100/files#r956755677.

src/interface.jl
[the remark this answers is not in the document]
mcabbott: Not sure I follow. `update!` claimed it would mutate the states if it wanted to, and would typically alter arrays. (And `update` claimed not to, but had a bug.)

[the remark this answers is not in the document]
ToucheSir: But if you had immutable arrays in your state tree before, the original state tree would be unchanged after `update!`. Perhaps we don't feel that was ever a solid guarantee (I don't), but we ought to get that point out in writing for posterity/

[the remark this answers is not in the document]
mcabbott: Sure. I think we should change the doc for `update!` to be explicit that it now guarantees to update the state tree. (I thought the old one said "inputs are trash afterwards" but in fact it is explicit only about the model.)

`update!` need not in fact return two arguments, but whether that is too confusing to change (and to differ from `update` which must) is another question.

src/interface.jl
ToucheSir: My impression was it implicitly acted as a `@nospecialize`, but looking at https://github.com/JuliaLang/julia/blob/98e1b13a7db5aa1d05b9a48085993375cf2298d0/src/method.c#L656 that may not be the case.

src/interface.jl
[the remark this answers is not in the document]
ToucheSir: Good point about the shared gradients. If this is a correctness issue, we don't have much of a choice :)

mcabbott:
Here's an evil case for shared parameters:
```julia
mutable struct MutTwo; x; y; end
Functors.@functor MutTwo

tmp = MutTwo([1.0], [2.0])
model = (a=tmp, b=tmp, c=MutTwo(tmp.x, tmp.y))
state = Optimisers.setup(Momentum(), model)

model.a === model.b
model.a !== model.c  # fields are identified, but struct is not

state.a.x === state.b.x
state.a === state.b
state.a === state.c  # unavoidable, but means we can't use Leaf ID alone?

mgrad = (a=(x=[1.], y=[10.]), b=(x=[100], y=[1000]), c=(x=[1/3], y=[1/30]))
state2, model2 = Optimisers.update(state, model, mgrad)
model2.a === model2.b
model2.a !== model2.c 
```
The state of all 3 components is `(x=Leaf(...), y=Leaf(...))`. A cache which is IdDict{Leaf,Any} can't identify the two structs. A cache which also stores higher levels of the state tree will instead identify all three structs. 

One answer here is to store tuples `(x, Leaf(...))`. Then identifying the `Leaf`s can be used as a trick to tie StaticArray parameters. But cannot tie Array parameters (which aren't already tied by `===`).

ToucheSir:
I don't think we ever guaranteed `model.a !== model.c => state.a === state.c`. `model2.a !== model2.c` after an update seems more like a bug rather than an intrinsic limitation? If `x` and `y` are `===` for all 3 components, then this should just work.

Issue #99, opened 2022-07-09
Rename or outsource `iswriteable`


ToucheSir:
Through a typo, I found that Base exports [`iswritable`](https://docs.julialang.org/en/v1/base/io-network/#Base.iswritable) (one char difference). Were we to outsource this, possible candidates include [`ChainRulesCore.is_inplaceable_destination`](https://juliadiff.org/ChainRulesCore.jl/stable/api.html#ChainRulesCore.is_inplaceable_destination) (used by `add!!`) and [`ArrayInterfaceCore.ismutable`](https://juliaarrays.github.io/ArrayInterface.jl/dev/api/#ArrayInterfaceCore.ismutable).

mcabbott:
The near-collision is unfortunate, we should at least re-name it.

The goal was to do the dead-simple minimal thing, erring on the side of safety (i.e. not writing unelss quite sure), and revisit when someone had a compelling case to do otherwise.

At least we depend on ChainRulesCore; I'm very reluctant to entangle this little package with the huge ArrayInterface. ChainRulesCore has more description of what its rules are, ArrayInterface doesn't try to explain:
```julia
julia> using ChainRulesCore, ArrayInterface, StaticArrays, ReadOnlyArrays

## CRC

julia> ChainRulesCore.is_inplaceable_destination(transpose(1:3))  # ok!
false

julia> ChainRulesCore.is_inplaceable_destination(Diagonal([1,2,3]))  # but you can only write some places, "an appropriate tangent"
true

julia> ChainRulesCore.is_inplaceable_destination(SA[1,2,3])  # ok
false

julia> ChainRulesCore.is_inplaceable_destination(ReadOnlyArray([1,2,3]))  # fooled, fails the wrong way
true

## AI

julia> ArrayInterface.ismutable(ReadOnlyArray([1,2,3]))  # likewise fooled
true

julia> ArrayInterface.ismutable(SA[1,2,3])  # isn't there literally a package registered to handle this?
true
```

ToucheSir:
[...]

PR #102, opened 2022-07-20
Rename `iswriteable`

mcabbott:
For #99


ToucheSir (approved):

CarloLucibello:
I think it is better to add a deprecation here

PR #103, opened 2022-07-27
fix type instability of `Rprop`

MilkshakeForReal:
Now the type of states is always the same as the of parameters

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -168,9 +168,8 @@ Rprop(η = 1f-3, ℓ = (5f-1, 1.2f0), Γ = (1f-6, 50f0)) = Rprop{typeof(η)}(η,
 init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
 
 function apply!(o::Rprop, state, x, dx)
-    ℓ, Γ = o.ell, o.gamma
-    ℓ = ℓ .|> eltype(x)
-    Γ = Γ .|> eltype(x)
+    ℓ = o.ell .|> eltype(x)
+    Γ = o.gamma .|> eltype(x)
     g, η = state
 
     η = broadcast(g, η, dx) do g, η, dx
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -25,7 +25,7 @@ init(o::Descent, x::AbstractArray) = nothing
 
 function apply!(o::Descent, state, x, dx)
   η = convert(float(eltype(x)), o.eta)
-  
+
   return state, @lazy dx * η  # @lazy creates a Broadcasted, will later fuse with x .= x .- dx
 end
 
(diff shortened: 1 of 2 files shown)


src/rules.jl
```diff
@@ -168,7 +168,8 @@ Rprop(η = 1f-3, ℓ = (5f-1, 1.2f0), Γ = (1f-6, 50f0)) = Rprop{typeof(η)}(η,
 init(o::Rprop, x::AbstractArray) = (zero(x), onevalue(o.eta, x))
 
 function apply!(o::Rprop, state, x, dx)
-    ℓ, Γ = o.ell, o.gamma
+    ℓ = o.ell .|> eltype(x)
+    Γ = o.gamma .|> eltype(x)
     g, η = state
```
mcabbott, line 173: This is the meat, `o.gamma` is a tuple. Would prefer `Γ = map(eltype(x), o.gamma)` but just style.

So the concern is that `Rprop(0.1)` will cause `(g, η)` to become Float64 arrays, I think. Can we test for this, ideally all of them? 

There are some tests here https://github.com/FluxML/Optimisers.jl/blob/master/test/runtests.jl#L120-L132 but they check only that the model's type isn't changed.

MilkshakeForReal: Should the returned `dx′` of `apply!` have the same element type with the input `x` or `dx`?

MilkshakeForReal: I have added some tests that should satisfy your requirement

test/runtests.jl
```diff
@@ -129,8 +129,18 @@ Optimisers.trainable(x::TwoThirds) = (a = x.a,)
...
+
+      dm = ([1.0,2.0], [3.0,4.0])
+      s1 = Optimisers.setup(Rprop(0.1), m)
+      s2, m2 = Optimisers.update(s1, m, dm)
+      @test eltype(m2[1]) == Float16   # type should not be promoted by the gradient
+      @test eltype(m2[2]) == Float32
+      @test eltype(s2[1].state[1]) == eltype(s1[1].state[1])
+      @test eltype(s2[1].state[2]) == eltype(s1[1].state[2])
+      @test eltype(s2[2].state[1]) == eltype(s1[2].state[1])
+      @test eltype(s2[2].state[2]) == eltype(s1[2].state[2])
```
mcabbott, line 141:
```suggestion
      # Rprop re-creates its state arrays, check they don't get widened:
      s5 = Optimisers.setup(Rprop(0.1), m)  # Float64 rule
      grad64 = ([1.0,2.0], SA[3.0,4.0])  # Float64 gradients
      s6, m6 = Optimisers.update(s5, m, grad64)
      @test eltype(m6[1]) == Float16
      @test eltype(m6[2]) == Float32
      @test eltype(s6[1].state[1]) == Float16
      @test eltype(s6[1].state[2]) == Float16
      @test eltype(s6[2].state[1]) == Float32
      @test eltype(s6[2].state[2]) == Float32
```

mcabbott (approved):
LGTM.

PR #104, opened 2022-08-09
Docstring for `isnumeric`

mcabbott:


docs/src/api.md
```diff
@@ -37,11 +37,14 @@ Optimisers.update!
 Optimisers.adjust(::Any, ::Real)
 ```
 
-Calling `Functors.@functor` on your model's layer types by default causes the
-optimiser to act on all suitable fields. To restrict this, define `trainable`:
+Calling `Functors.@functor` on your model's layer types by default causes
+these functions to recurse into all children, and ultimately optimise
+all `isnumeric` leaf nodes.
```
ToucheSir, line 42: Would adding a ref link work here?

Saransh-cpp:
The preview - https://fluxml.ai/Optimisers.jl/previews/PR104/api/#Optimisers.isnumeric

docs/src/api.md
Saransh-cpp: Should work!

mcabbott: I guess my feeling is that these should not be links. This sentence is introducing the docstrings immediately below, and a link that scrolls down 2cm seems more likely to be confusing than helpful.

Saransh-cpp: Oh yes, I overlooked that.

ToucheSir (approved):

Issue #96, opened 2022-07-03
Investigate using a different AD for tests


ToucheSir:
Zygote compile times make what ordinarily should be a pretty fast test suite rather sluggish. Perhaps we could borrow some functionality from CRTU: https://github.com/JuliaDiff/ChainRulesTestUtils.jl/blob/v1.8.1/src/testers.jl#L224.

mcabbott:
It would be nice to be quicker. But the tests aren't just about mathematical correctness, some are also checking that missing branches with `nothing` are handled correctly, etc. If https://github.com/JuliaDiff/Diffractor.jl/issues/66 ever happens, these should ideally be supplemented by checks that `ZeroTangent` works equally well.

ToucheSir:
Would FD.jl not generate `NoTangent` for those same branches? I've not tried it with nested structs.

mcabbott:
It seems to make dense zeros:
```julia
julia> using FiniteDifferences

julia> grad(central_fdm(5, 1), x -> sum(x.a), (a=[1,2f0], b=([3,4f0], 5.0)))
((a = Float32[0.9999863, 1.0000029], b = (Float32[0.0, 0.0], 0.0)),)
```
I wondered if Tracker might work well, but it doesn't seem to like NamedTuples. 

For comparison:
```julia
julia> Zygote.gradient(x -> sum(x.a), (a=[1,2f0], b=([3,4f0], 5.0)))
((a = Fill(1.0f0, 2), b = nothing),)

julia> Diffractor.gradient(x -> sum(x.a), (a=[1,2f0], b=([3,4f0], 5.0)))
(Tangent{NamedTuple{(:a, :b), Tuple{Vector{Float32}, Tuple{Vector{Float32}, Float64}}}}(a = InplaceableThunk(ChainRules.var"#1547#1550"{Float32, Colon}(1.0f0, Colon()), Thunk(ChainRules.var"#1548#1551"{Float32, Colon, Vector{Float32}, ProjectTo{AbstractArray, NamedTuple{(:element, :axes), Tuple{ProjectTo{Float32, NamedTuple{(), Tuple{}}}, Tuple{Base.OneTo{Int64}}}}}}(1.0f0, Colon(), Float32[1.0, 2.0], ProjectTo{AbstractArray}(element = ProjectTo{Float32}(), axes = (Base.OneTo(2),))))),),)
```

dfdx:
[...]

PR #105, opened 2022-08-19
Test with Yota, too

mcabbott:
Does not close https://github.com/FluxML/Optimisers.jl/issues/96, in fact this surely makes tests slower. But perhaps it's good to get something besides Zygote running?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,13 +13,15 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "1"
 Functors = "0.3, 0.4"
+Yota = "0.7.3"
 Zygote = "0.6.40"
 julia = "1.6"
 
 [extras]
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Yota = "cd998857-8626-517d-b929-70ad188a48f0"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [targets]
-test = ["Test", "StaticArrays", "Zygote"]
+test = ["Test", "StaticArrays", "Yota", "Zygote"]
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "1"
 Functors = "0.3, 0.4"
-Yota = "0.7.3"
+Yota = "0.8.1"
 Zygote = "0.6.40"
 julia = "1.6"
 
--- a/Project.toml
+++ b/Project.toml
@@ -13,7 +13,7 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "1"
 Functors = "0.3, 0.4"
-Yota = "0.8.1"
+Yota = "0.8.2"
 Zygote = "0.6.40"
 julia = "1.6"
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -52,7 +52,7 @@ state = Optimisers.setup(rule, model);  # initialise this optimiser's momentum e
 end;
 
 state, model = Optimisers.update(state, model, ∇model);
-@show sum(model(image));
+@show sum(model(image));  # reduced
 
 ```
 
@@ -82,14 +82,51 @@ To extract what Optimisers.jl needs, you can write `_, (_, ∇model) = Yota.grad
 or, for the Flux model above:
 
 ```julia
+using Yota
+
 loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
   sum(m(x))
 end;
 ```
 
+Unfortunately this example doesn't actually run right now. This is the error:
+```
+julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
+         sum(m(x))
+       end;
+┌ Error: Failed to compile rrule for #233(Chain(Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64, relu), Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64)),), extract details via:
+│ 	(f, args) = Yota.RRULE_VIA_AD_STATE[]
+└ @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:160
+ERROR: No deriative rule found for op %3 = getfield(%1, :x)::Array{Float32, 4} , try defining it using 
+
+	ChainRulesCore.rrule(::typeof(getfield), ::Flux.var"#233#234"{Array{Float32, 4}}, ::Symbol) = ...
+
+Stacktrace:
+  [1] error(s::String)
+    @ Base ./error.jl:35
+  [2] step_back!(tape::Umlaut.Tape{Yota.GradCtx}, y::Umlaut.Variable)
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:197
+  [3] back!(tape::Umlaut.Tape{Yota.GradCtx}; seed::Symbol)
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:238
+  [4] gradtape!(tape::Umlaut.Tape{Yota.GradCtx}; seed::Symbol)
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:249
+  [5] gradtape(f::Flux.var"#233#234"{Array{Float32, 4}}, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}}; ctx::Yota.GradCtx, seed::Symbol)
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:276
+  [6] make_rrule(f::Function, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}})
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:109
+  [7] rrule_via_ad(#unused#::Yota.YotaRuleConfig, f::Function, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}})
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:153
+...
+
+(jl_GWa2lX) pkg> st
+Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_GWa2lX/Project.toml`
+⌃ [587475ba] Flux v0.13.4
+  [cd998857] Yota v0.7.4
+```
+
 ## Usage with [Lux.jl](https://github.com/avik-pal/Lux.jl)
 
-The main design difference of Lux is that the tree of parameters is separate from
+The main design difference of Lux from Flux is that the tree of parameters is separate from
 the layer structure. It is these parameters which `setup` and `update` need to know about.
 
 Lux describes this separation of parameter storage from model description as "explicit" parameters.
@@ -104,7 +141,7 @@ using Lux, Boltz, Zygote, Optimisers
 lux_model, params, lux_state = Boltz.resnet(:resnet18) |> gpu;  # define and initialise model
 images = rand(Float32, 224, 224, 3, 4) |> gpu;  # batch of dummy data
 y, lux_state = Lux.apply(lux_model, images, params, lux_state);  # run the model
-@show sum(y)  # initial dummy loss
+@show sum(y);  # initial dummy loss
 
 rule = Optimisers.Adam()
 opt_state = Optimisers.setup(rule, params);  # optimiser state based on model parameters
@@ -112,17 +149,19 @@ opt_state = Optimisers.setup(rule, params);  # optimiser state based on model pa
 (loss, lux_state), back = Zygote.pullback(params, images) do p, x
   y, st = Lux.apply(lux_model, x, p, lux_state)
   sum(y), st  # return both the loss, and the updated lux_state
-end
-∇params, _ = back((one.(loss), nothing))  # gradient of only the loss, with respect to parameter tree
-
-@show sum(loss)
+end;
+∇params, _ = back((one.(loss), nothing));  # gradient of only the loss, with respect to parameter tree
+loss == sum(y)  # not yet changed
 
 opt_state, params = Optimisers.update!(opt_state, params, ∇params);
+
+y, lux_state = Lux.apply(lux_model, images, params, lux_state);
+@show sum(y);  # now reduced
 
 ```
 
 Besides the parameters stored in `params` and gradually optimised, any other model state
-is stored in `lux_state`, and returned by `Lux.apply`.
+is stored in `lux_state`, and updated by `Lux.apply`. (In this example, BatchNorm has state.)
 This is completely unrelated to Optimisers.jl's state, although designed in a similar spirit.
 If you are certain there is no model state, then the gradient calculation can
 be simplified to use `Zygote.gradient` instead of `Zygote.pullback`:
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -38,7 +38,7 @@ to adjust the model:
 
 ```julia
 
-using Flux, Metalhead, Optimisers
+using Flux, Metalhead, Zygote, Optimisers
 
 model = Metalhead.ResNet(18) |> gpu  # define a model to train
 image = rand(Float32, 224, 224, 3, 1) |> gpu;  # dummy data
@@ -72,6 +72,21 @@ This `∇model` is another tree structure, rather than the dictionary-like objec
 Zygote's "implicit" mode `gradient(() -> loss(...), Flux.params(model))` -- see 
 [Zygote's documentation](https://fluxml.ai/Zygote.jl/dev/#Explicit-and-Implicit-Parameters-1) for more about this difference.
 
+## Usage with [Yota.jl](https://github.com/dfdx/Yota.jl)
+
+Yota is another modern automatic differentiation package, an alternative to Zygote.
+
+Its main function is `Yota.grad`, which returns the loss as well as the gradient (like `Zygote.withgradient`)
+but also returns a gradient component for the loss function.
+To extract what Optimisers.jl needs, you can write `_, (_, ∇model) = Yota.grad(f, model, data)`
+or, for the Flux model above:
+
+```julia
+loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
+  sum(m(x))
+end;
+```
+
 ## Usage with [Lux.jl](https://github.com/avik-pal/Lux.jl)
 
 The main design difference of Lux is that the tree of parameters is separate from
@@ -79,7 +94,7 @@ the layer structure. It is these parameters which `setup` and `update` need to k
 
 Lux describes this separation of parameter storage from model description as "explicit" parameters.
 Beware that it has nothing to do with Zygote's notion of "explicit" gradients.
-(If the same model is written in Flux and Lux, `∇model` above and `∇params` below will often be
+(If the same model is written in Flux and Lux, `∇model` above and `∇params` below will be nearly
 identical trees of nested `NamedTuple`s.)
 
 ```julia
@@ -88,27 +103,37 @@ using Lux, Boltz, Zygote, Optimisers
 
 lux_model, params, lux_state = Boltz.resnet(:resnet18) |> gpu;  # define and initialise model
 images = rand(Float32, 224, 224, 3, 4) |> gpu;  # batch of dummy data
-y, _ = Lux.apply(lux_model, images, params, lux_state);  # run the model
+y, lux_state = Lux.apply(lux_model, images, params, lux_state);  # run the model
 @show sum(y)  # initial dummy loss
 
 rule = Optimisers.Adam()
 opt_state = Optimisers.setup(rule, params);  # optimiser state based on model parameters
 
-∇params, _ = gradient(params, images) do p, x  # gradient with respect to parameter tree
-  y, _ = Lux.apply(lux_model, x, p, lux_state)
-  sum(y)
-end;
+(loss, lux_state), back = Zygote.pullback(params, images) do p, x
+  y, st = Lux.apply(lux_model, x, p, lux_state)
+  sum(y), st  # return both the loss, and the updated lux_state
+end
+∇params, _ = back((one.(loss), nothing))  # gradient of only the loss, with respect to parameter tree
+
+@show sum(loss)
 
 opt_state, params = Optimisers.update!(opt_state, params, ∇params);
-
-y, _ = Lux.apply(lux_model, images, params, lux_state);
-@show sum(y)
 
 ```
 
 Besides the parameters stored in `params` and gradually optimised, any other model state
-is stored in `lux_state`. For simplicity this example does not show how to propagate the 
-updated `lux_state` to the next iteration, see Lux's documentation.
+is stored in `lux_state`, and returned by `Lux.apply`.
+This is completely unrelated to Optimisers.jl's state, although designed in a similar spirit.
+If you are certain there is no model state, then the gradient calculation can
+be simplified to use `Zygote.gradient` instead of `Zygote.pullback`:
+
+```julia
+∇params, _ = gradient(params, images) do p, x
+  y, _ = Lux.apply(lux_model, x, p, lux_state)  # discards new lux_state
+  sum(y)
+end;
+```
+
 
 ## Non-`trainable` Parameters
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -64,6 +64,14 @@ There is also [`Optimisers.update!`](@ref) which similarly returns a new model a
 but is free to mutate arrays within the old one for efficiency.
 The method of `apply!` for each rule is likewise free to mutate arrays within its state;
 they are defensively copied when this rule is used with `update`.
+(The method of `apply!` above is likewise free to mutate arrays within its state;
+they are defensively copied when this rule is used with `update`.)
+For `Adam()`, there are two momenta per parameter, thus `state` is about twice the size of `model`:
+
+```julia
+Base.summarysize(model) / 1024^2  # about 45MB
+Base.summarysize(state) / 1024^2  # about 90MB
+```
 
 Optimisers.jl does not depend on any one automatic differentiation package,
 but for now the most likely source of gradients is [Zygote.jl](https://fluxml.ai/Zygote.jl).
@@ -71,6 +79,7 @@ Note that `update` always wants the gradient from Zygote's "explicit" mode, as s
 This `∇model` is another tree structure, rather than the dictionary-like object from 
 Zygote's "implicit" mode `gradient(() -> loss(...), Flux.params(model))` -- see 
 [Zygote's documentation](https://fluxml.ai/Zygote.jl/dev/#Explicit-and-Implicit-Parameters-1) for more about this difference.
+
 
 ## Usage with [Yota.jl](https://github.com/dfdx/Yota.jl)
 
@@ -89,40 +98,6 @@ loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
 end;
 ```
 
-Unfortunately this example doesn't actually run right now. This is the error:
-```
-julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
-         sum(m(x))
-       end;
-┌ Error: Failed to compile rrule for #233(Chain(Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64, relu), Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64)),), extract details via:
-│ 	(f, args) = Yota.RRULE_VIA_AD_STATE[]
-└ @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:160
-ERROR: No deriative rule found for op %3 = getfield(%1, :x)::Array{Float32, 4} , try defining it using 
-
-	ChainRulesCore.rrule(::typeof(getfield), ::Flux.var"#233#234"{Array{Float32, 4}}, ::Symbol) = ...
-
-Stacktrace:
-  [1] error(s::String)
-    @ Base ./error.jl:35
-  [2] step_back!(tape::Umlaut.Tape{Yota.GradCtx}, y::Umlaut.Variable)
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:197
-  [3] back!(tape::Umlaut.Tape{Yota.GradCtx}; seed::Symbol)
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:238
-  [4] gradtape!(tape::Umlaut.Tape{Yota.GradCtx}; seed::Symbol)
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:249
-  [5] gradtape(f::Flux.var"#233#234"{Array{Float32, 4}}, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}}; ctx::Yota.GradCtx, seed::Symbol)
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:276
-  [6] make_rrule(f::Function, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}})
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:109
-  [7] rrule_via_ad(#unused#::Yota.YotaRuleConfig, f::Function, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}})
-    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:153
-...
-
-(jl_GWa2lX) pkg> st
-Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_GWa2lX/Project.toml`
-⌃ [587475ba] Flux v0.13.4
-  [cd998857] Yota v0.7.4
-```
 
 ## Usage with [Lux.jl](https://github.com/avik-pal/Lux.jl)
 
@@ -163,6 +138,14 @@ y, lux_state = Lux.apply(lux_model, images, params, lux_state);
 Besides the parameters stored in `params` and gradually optimised, any other model state
 is stored in `lux_state`, and updated by `Lux.apply`. (In this example, BatchNorm has state.)
 This is completely unrelated to Optimisers.jl's state, although designed in a similar spirit.
+
+```julia
+Base.summarysize(lux_model) / 1024   # just 2KB
+Base.summarysize(params) / 1024^2    # about 45MB, same as Flux model
+Base.summarysize(lux_state) / 1024   # 40KB
+Base.summarysize(opt_state) / 1024^2 # about 90MB, with Adam
+```
+
 If you are certain there is no model state, then the gradient calculation can
 be simplified to use `Zygote.gradient` instead of `Zygote.pullback`:
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -62,8 +62,6 @@ tree formed by the model and update the parameters using the gradients.
 
 There is also [`Optimisers.update!`](@ref) which similarly returns a new model and new state,
 but is free to mutate arrays within the old one for efficiency.
-The method of `apply!` for each rule is likewise free to mutate arrays within its state;
-they are defensively copied when this rule is used with `update`.
 (The method of `apply!` above is likewise free to mutate arrays within its state;
 they are defensively copied when this rule is used with `update`.)
 For `Adam()`, there are two momenta per parameter, thus `state` is about twice the size of `model`:
@@ -87,17 +85,18 @@ Yota is another modern automatic differentiation package, an alternative to Zygo
 
 Its main function is `Yota.grad`, which returns the loss as well as the gradient (like `Zygote.withgradient`)
 but also returns a gradient component for the loss function.
-To extract what Optimisers.jl needs, you can write `_, (_, ∇model) = Yota.grad(f, model, data)`
-or, for the Flux model above:
+To extract what Optimisers.jl needs, you can write (for the Flux model above):
 
 ```julia
 using Yota
 
 loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
-  sum(m(x))
+  sum(m(x)
 end;
-```
 
+# Or else, this may save computing ∇image:
+loss, (_, ∇model) = grad(m -> sum(m(image)), model);
+```
 
 ## Usage with [Lux.jl](https://github.com/avik-pal/Lux.jl)
 
--- a/test/destructure.jl
+++ b/test/destructure.jl
@@ -104,6 +104,32 @@ end
     # Zygote error in (::typeof(∂(canonicalize)))(Δ::NamedTuple{(:backing,), Tuple{NamedTuple{(:x, :y, :z)
     # Diffractor error in perform_optic_transform
   end
+  
+  @testset "using Yota" begin
+    @test_broken Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]  # Unexpected expression: $(Expr(:static_parameter, 1))
+    # These are all broken!
+    @test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], ZeroTangent())
+    @test Yota_gradient(m -> destructure(m)[1][1], m3)[1] == (x = [1,0,0], y = ZeroTangent(), z = [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][2], m4)[1] == (x = [0,1,0], y = ZeroTangent(), z = [0,0,0])
+
+    g5 = Yota_gradient(m -> destructure(m)[1][3], m5)[1]
+    @test g5.a[1].x == [0,0,1]
+    @test g5.a[2] === ZeroTangent()
+
+    g6 = Yota_gradient(m -> imag(destructure(m)[1][4]), m6)[1]
+    @test g6.a == [0,0,0]
+    @test g6.a isa Vector{Float64}
+    @test g6.b == [0+im]
+
+    g8 = Yota_gradient(m -> sum(abs2, destructure(m)[1]), m8)[1]
+    @test g8[1].x == [2,4,6]
+    @test g8[2].b.x == [8]
+    @test g8[3] == [[10.0]]
+
+    g9 = Yota_gradient(m -> sum(sqrt, destructure(m)[1]), m9)[1]
+    @test g9.c === ZeroTangent()
+  end
 end
 
 @testset "gradient of rebuild" begin
@@ -148,6 +174,36 @@ end
     end[1] ≈ [0,0,0,32,40,48]
     # Not fixed by this:
     # Zygote.@adjoint Tangent{T,B}(x::NamedTuple) where {T,B<:NamedTuple} = Tangent{T,B}(x), dx -> (dx,)
+  end
+  
+  @testset "using Yota" begin
+    re1 = destructure(m1)[2]
+    @test Yota_gradient(x -> re1(x)[1], rand(3))[1] == [1,0,0]
+    re2 = destructure(m2)[2]
+    @test Yota_gradient(x -> re2(x)[1][2], rand(6))[1] == [0,1,0,0,0,0]
+    re3 = destructure(m3)[2]
+    @test Yota_gradient(x -> re3(x).x[3], rand(6))[1] == [0,0,1,0,0,0]
+    @test Yota_gradient(x -> re3(x).z[1], rand(6))[1] == [0,0,0,1,0,0]
+
+    re4 = destructure(m4)[2]
+    @test Yota_gradient(x -> re4(x).x[1], rand(6))[1] == [1,0,0,0,0,0]
+    @test Yota_gradient(x -> re4(x).y[2], rand(6))[1] == [0,1,0,0,0,0]
+    @test Yota_gradient(rand(6)) do x
+      m = re4(x)
+      m.x[1] + 2*m.y[2] + 3*m.z[3]
+    end[1] == [1,2,0, 0,0,3]
+
+    re7 = destructure(m7)[2]
+    @test Yota_gradient(x -> re7(x).a[2][3], rand(3))[1] == [0,0,1]
+    @test Yota_gradient(x -> re7(x).b[2][2], rand(3))[1] == [0,0,0]
+    @test Yota_gradient(x -> re7(x).c[2][1], rand(3))[1] == [0,0,0]
+
+    v8, re8 = destructure(m8)
+    @test_broken Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]  # MethodError: no method matching zero(::Type{Any})
+    @test_broken Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]  # MethodError: no method matching !(::Expr)
+
+    re9 = destructure(m9)[2]
+    @test_broken Yota_gradient(x -> sum(abs2, re9(x).c[1]), 1:7)[1] == [0,0,0, 8,10,12,14]  # MethodError: no method matching zero(::Type{Array})
   end
 end
 
--- a/test/destructure.jl
+++ b/test/destructure.jl
@@ -106,16 +106,15 @@ end
   end
   
   @testset "using Yota" begin
-    @test_broken Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]  # Unexpected expression: $(Expr(:static_parameter, 1))
-    # These are all broken!
+    @test Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]
     @test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
-    @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], ZeroTangent())
-    @test Yota_gradient(m -> destructure(m)[1][1], m3)[1] == (x = [1,0,0], y = ZeroTangent(), z = [0,0,0])
-    @test Yota_gradient(m -> destructure(m)[1][2], m4)[1] == (x = [0,1,0], y = ZeroTangent(), z = [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], nothing)
+    @test Yota_gradient(m -> destructure(m)[1][1], m3)[1] == (x = [1,0,0], y = nothing, z = [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][2], m4)[1] == (x = [0,1,0], y = nothing, z = [0,0,0])
 
     g5 = Yota_gradient(m -> destructure(m)[1][3], m5)[1]
     @test g5.a[1].x == [0,0,1]
-    @test g5.a[2] === ZeroTangent()
+    @test g5.a[2] === nothing
 
     g6 = Yota_gradient(m -> imag(destructure(m)[1][4]), m6)[1]
     @test g6.a == [0,0,0]
@@ -128,7 +127,7 @@ end
     @test g8[3] == [[10.0]]
 
     g9 = Yota_gradient(m -> sum(sqrt, destructure(m)[1]), m9)[1]
-    @test g9.c === ZeroTangent()
+    @test g9.c === nothing
   end
 end
 
@@ -199,11 +198,11 @@ end
     @test Yota_gradient(x -> re7(x).c[2][1], rand(3))[1] == [0,0,0]
 
     v8, re8 = destructure(m8)
-    @test_broken Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]  # MethodError: no method matching zero(::Type{Any})
-    @test_broken Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]  # MethodError: no method matching !(::Expr)
+    @test Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]
+    @test Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]
 
     re9 = destructure(m9)[2]
-    @test_broken Yota_gradient(x -> sum(abs2, re9(x).c[1]), 1:7)[1] == [0,0,0, 8,10,12,14]  # MethodError: no method matching zero(::Type{Array})
+    @test Yota_gradient(x -> sum(abs2, re9(x).c[1]), 1:7)[1] == [0,0,0, 8,10,12,14]
   end
 end
 
--- a/test/destructure.jl
+++ b/test/destructure.jl
@@ -105,7 +105,7 @@ end
     # Diffractor error in perform_optic_transform
   end
   
-  @testset "using Yota" begin
+  VERSION < v"1.9-" && @testset "using Yota" begin
     @test Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]
     @test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
     @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], nothing)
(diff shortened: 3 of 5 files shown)


test/destructure.jl
```diff
@@ -149,6 +175,36 @@ end
...
+    end[1] == [1,2,0, 0,0,3]
+
+    re7 = destructure(m7)[2]
+    @test Yota_gradient(x -> re7(x).a[2][3], rand(3))[1] == [0,0,1]
+    @test Yota_gradient(x -> re7(x).b[2][2], rand(3))[1] == [0,0,0]
+    @test Yota_gradient(x -> re7(x).c[2][1], rand(3))[1] == [0,0,0]
+
+    v8, re8 = destructure(m8)
+    @test_broken Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]  # MethodError: no method matching zero(::Type{Any})
+    @test_broken Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]  # MethodError: no method matching !(::Expr)
```
mcabbott, line 203: Errors here:
```
using Optimisers, Functors, Yota

struct Foo; x; y; end
Functors.@functor Foo
Optimisers.trainable(x::Foo) = (x.y, x.x)

Yota_gradient(f, xs...) = Base.tail(Yota.grad(f, xs...)[2])

m1 = collect(1:3.0)
m8 = [Foo(m1, m1), (a = true, b = Foo([4.0], false), c = ()), [[5.0]]]
```
 
```
julia> Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1]
┌ Warning: trainable(x) should now return a NamedTuple with the field names, not a Tuple
└ @ Optimisers ~/.julia/packages/Optimisers/dZMII/src/interface.jl:97
ERROR: MethodError: no method matching zero(::Type{Any})

Closest candidates are:
  zero(::Type{Union{Missing, T}}) where T
   @ Base missing.jl:105
  zero(::Union{Type{P}, P}) where P<:Dates.Period
   @ Dates ~/.julia/dev/julia/usr/share/julia/stdlib/v1.9/Dates/src/periods.jl:53
  zero(::SparseArrays.AbstractSparseArray)
   @ SparseArrays ~/.julia/dev/julia/usr/share/julia/stdlib/v1.9/SparseArrays/src/SparseArrays.jl:76
  ...

Stacktrace:
  [1] zero(#unused#::Type{Any})
    @ Base ./missing.jl:106
  [2] zero(x::Vector{Any})
    @ Base ./abstractarray.jl:1182
  [3] ungetindex(x::Vector{Any}, dy::Tangent{Foo, NamedTuple{(:y,), Tuple{Vector{Float64}}}}, I::Int64)
    @ Yota ~/.julia/packages/Yota/GIFMf/src/helpers.jl:24
  [4] (::Yota.var"#getindex_pullback#44"{Vector{Any}, Tuple{Int64}})(dy::Tangent{Foo, NamedTuple{(:y,), Tuple{Vector{Float64}}}})
    @ Yota ~/.julia/packages/Yota/GIFMf/src/rulesets.jl:152

julia> Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1]
ERROR: MethodError: no method matching !(::Expr)

Closest candidates are:
  !(::ComposedFunction{typeof(!)})
   @ Base operators.jl:1074
  !(::Function)
   @ Base operators.jl:1073
  !(::Bool)
   @ Base bool.jl:35
  ...

Stacktrace:
  [1] trace!(t::Umlaut.Tracer{Yota.GradCtx}, v_fargs::Tuple{typeof(only), Umlaut.Variable})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:448
  [2] trace_call!(::Umlaut.Tracer{Yota.GradCtx}, ::Function, ::Vararg{Any})
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:76
  [3] trace_block!(t::Umlaut.Tracer{Yota.GradCtx}, ir::Core.Compiler.IRCode, bi::Int64, prev_bi::Int64, sparams::Core.SimpleVector)
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:312
  [4] trace!(t::Umlaut.Tracer{Yota.GradCtx}, v_fargs::Vector{Umlaut.Variable})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:436
  [5] trace(f::Function, args::Vector{Float64}; ctx::Yota.GradCtx, deprecated_kws::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:546
  [6] gradtape(f::var"#34#35", args::Vector{Float64}; ctx::Yota.GradCtx, seed::Int64)
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:270
  [7] grad(f::var"#34#35", args::Vector{Float64}; seed::Int64)
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:346
```

test/destructure.jl
```diff
@@ -104,6 +104,32 @@ end
     # Zygote error in (::typeof(∂(canonicalize)))(Δ::NamedTuple{(:backing,), Tuple{NamedTuple{(:x, :y, :z)
     # Diffractor error in perform_optic_transform
   end
+  
+  @testset "using Yota" begin
+    @test_broken Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]  # Unexpected expression: $(Expr(:static_parameter, 1))
```
mcabbott, line 109: Error here:
```
julia> Yota_gradient(m -> destructure(m)[1][1], m1)[1]
ERROR: Unexpected expression: $(Expr(:static_parameter, 1))
Full IRCode:

 55 1 ─ %4 = $(Expr(:static_parameter, 1))::Core.Const(Vector{Float64})
   │   %5 = $(Expr(:static_parameter, 2))::Core.Const(Int64)
   │   %1 = Core.apply_type(Optimisers.Restructure, %4, %5)::Core.Const(Optimisers.Restructure{Vector{Float64}, Int64})
   │   %2 = (%1)(_2, _3, _4)::Optimisers.Restructure{Vector{Float64}, Int64}
   └──      return %2
   
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] trace_block!(t::Umlaut.Tracer{Yota.GradCtx}, ir::Core.Compiler.IRCode, bi::Int64, prev_bi::Int64, sparams::Core.SimpleVector)
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:330
  [3] trace!(t::Umlaut.Tracer{Yota.GradCtx}, v_fargs::Tuple{UnionAll, Umlaut.Variable, Umlaut.Variable, Umlaut.Variable})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:436
  [4] trace_call!(::Umlaut.Tracer{Yota.GradCtx}, ::Type, ::Vararg{Any})
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:76
  [5] trace_block!(t::Umlaut.Tracer{Yota.GradCtx}, ir::Core.Compiler.IRCode, bi::Int64, prev_bi::Int64, sparams::Core.SimpleVector)
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:312
  [6] trace!(t::Umlaut.Tracer{Yota.GradCtx}, v_fargs::Tuple{typeof(destructure), Umlaut.Variable})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:436
  [7] trace_call!(::Umlaut.Tracer{Yota.GradCtx}, ::Function, ::Vararg{Any})
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:76
  [8] trace_block!(t::Umlaut.Tracer{Yota.GradCtx}, ir::Core.Compiler.IRCode, bi::Int64, prev_bi::Int64, sparams::Core.SimpleVector)
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:312
  [9] trace!(t::Umlaut.Tracer{Yota.GradCtx}, v_fargs::Vector{Umlaut.Variable})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:436
 [10] trace(f::Function, args::Vector{Float64}; ctx::Yota.GradCtx, deprecated_kws::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Umlaut ~/.julia/packages/Umlaut/LH23t/src/trace.jl:546
 [11] gradtape(f::var"#36#37", args::Vector{Float64}; ctx::Yota.GradCtx, seed::Int64)
    @ Yota ~/.julia/packages/Yota/GIFMf/src/grad.jl:270
```

mcabbott (commented):
Paging @dfdx about these errors.

docs/src/index.md
```diff
@@ -72,14 +72,66 @@ but is free to mutate arrays within the old one for efficiency.
...
+
+Unfortunately this example doesn't actually run right now. This is the error:
+```
+julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
+         sum(m(x))
+       end;
+┌ Error: Failed to compile rrule for #233(Chain(Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64, relu), Conv((3, 3), 64 => 64, pad=1, bias=false), BatchNorm(64)),), extract details via:
+│ 	(f, args) = Yota.RRULE_VIA_AD_STATE[]
+└ @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:160
+ERROR: No deriative rule found for op %3 = getfield(%1, :x)::Array{Float32, 4} , try defining it using
```
mcabbott, line 100: Maybe this should stay WIP for a bit.

dfdx: Thanks for pinging me! I'll be able to check out these errors during the weekend.

dfdx:
Some of the broken tests are already fixed on `main`, others need some adjustment (e.g. `ZeroTangent()` vs. `NoTangent()`), but I think I'll be able to fix them in a couple of days. 

One think that I seem to be missing is why `destructure/Restructure` needs to be differentiable. I'd expect a training loop to like this:

```julia
model = MyModel()
state = Optimisers.setup(Optimisers.Adam(), model) 
input = ...
loss = ...
for i=1:N
    grad = gradient(loss, model, input)                                                           # differentiable part
    state, model = Optimisers.update(state, model, grad)  # at every step      # non necessarily differentiable
end
```
`state` points to the trainable parameters of `MyModel()` and lets us update them, but never steps into gradient calculation. Yet, you test things like `Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1]`, so my picture of the world is obviously incomplete.

mcabbott:
Sounds good. I have no idea if the tests have ZeroTangent() vs. NoTangent() the wrong way around, fine to adjust tests to whatever is produced.

I broke Flux at some point because it turned out half the SciML universe rested on the gradient of `destructrure`, and there were exactly zero tests... it was so rudimentary I assumed it was for saving to CSV or similar use only. But I think it gets used as an interface between things which don't like nested structures (like calling some package for LBFGS) and models which do. Or obtaining a Hessian of the parameters of some Flux model.

dfdx:
I have a question regarding tests like this:
```julia
@test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
```
Currently, Yota returns `Tangent{Tuple{Vector{Float64}, Vector{Float64}}}([0.0, 1.0, 0.0], [0.0, 0.0, 0.0])` so can do e.g.:

```julia
g = Yota_gradient(m -> destructure(m)[1][2], m2)[1]
g + [0, 0, 0]     # =>     [0.0, 1.0, 0.0]
```
Is it what Optimisers.jl expects or should I better return a plain tuple as in the test?

mcabbott:
I think a Tangent is fine, this term is the gradient with respect to a Tuple. 

The test should be changed to allow for this, or perhaps the `Yota_gradient` function should convert, since its job is to make tests look the same.

dfdx:
I fixed the most hardcore issues in the tests, but after several days of investigation I can't solve 2 remaining problems:

* `ZeroTangent` vs `NoTangent`. Honestly, I still don't have a clear understanding of the difference. For example, function arguments may be generalized to callable structs, so it makes to return `ZeroTangent()` for them, yet most examples in ChainRules return `NoTangent()`. Another case is function `ChainRules.var"#fieldtype_pullback#422`, that, being applied to `ZeroTangent()`, returns `(NoTangent(), NoTangent(), NoTangent())`. So I wasn't able to adjust Yota's behavior to the tests (which reflects the behavior of Zygote, right?), but I'm open to suggestions. 
* The gradient seem to be packed and unpacked differently. For example, to account for the `Tangent{Tuple}` vs `Tuple` case above, I tried to modify `Yota_gradient` to this:

```julia
unpack(x::Tangent) = x.backing
unpack(x) = x
function Yota_gradient(f, xs...)
  g = Base.tail(Yota.grad(f, xs...)[2])
  return map(unpack, g)
end
```
It helped with some tested, but broke others. Structurally, the results seem to be correct, but I don't quite understand what needs to be adjusted - Yota, `Yota_gradient` or tests themselves. 

I'm going to proceed with testing of Yota on Flux models + Optimisers, which should uncover more inconsistencies, but if you are want to make another pass on these tests. please try Yota@0.8.0 and share your thoughts!

test/runtests.jl
```diff
@@ -13,6 +13,8 @@ struct TwoThirds a; b; c; end
 Functors.@functor TwoThirds (a, c)
 Optimisers.trainable(x::TwoThirds) = (a = x.a,)
 
+Yota_gradient(f, xs...) = Base.tail(Yota.grad(f, xs...)[2])
```
mcabbott, line 16: Maybe this is a better rough translation function:
```suggestion
Yota_gradient(f, xs...) = map(y2z, Base.tail(Yota.grad(f, xs...)[2]));
y2z(::AbstractZero) = nothing  # we don't care about different flavours
y2z(t::Tangent) = map(y2z, ChainRulesCore.backing(canonicalize(t)))
y2z(x) = x
```
The only goal is to have as few changes as possible between tests using Zygote and the same with Yota.

Project.toml
```diff
@@ -13,13 +13,15 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [compat]
 ChainRulesCore = "1"
 Functors = "0.2.8, 0.3"
+Yota = "0.7.3"
```
mcabbott, line 16:
```suggestion
Yota = "0.8.0"
```

docs/src/index.md
```diff
@@ -72,14 +72,66 @@ but is free to mutate arrays within the old one for efficiency.
...
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:109
+  [7] rrule_via_ad(#unused#::Yota.YotaRuleConfig, f::Function, args::Flux.Chain{Tuple{Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(relu), Vector{Float32}, Float32, Vector{Float32}}, Flux.Conv{2, 4, typeof(identity), Array{Float32, 4}, Bool}, Flux.BatchNorm{typeof(identity), Vector{Float32}, Float32, Vector{Float32}}}})
+    @ Yota ~/.julia/packages/Yota/GIFMf/src/cr_api.jl:153
+...
+
+(jl_GWa2lX) pkg> st
+Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_GWa2lX/Project.toml`
+⌃ [587475ba] Flux v0.13.4
+  [cd998857] Yota v0.7.4
+```
```
mcabbott, line 125:
````suggestion
julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
          sum(m(x))
         end;
ERROR: BoundsError: attempt to access Nothing at index [1]
Stacktrace:
  [1] _getfield(value::Nothing, fld::Int64)
    @ Yota ~/.julia/packages/Yota/uu3H0/src/helpers.jl:40
  [2] mkcall(::Function, ::Umlaut.Variable, ::Vararg{Any}; val::Missing, line::Nothing, kwargs::NamedTuple{(), Tuple{}}, free_kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Umlaut ~/.julia/packages/Umlaut/SvDaQ/src/tape.jl:192
  [3] mkcall
    @ ~/.julia/packages/Umlaut/SvDaQ/src/tape.jl:174 [inlined]
  [4] chainrules_transform!(tape::Umlaut.Tape{Yota.GradCtx})
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:183

...

(@v1.9) pkg> st Yota
Status `~/.julia/environments/v1.9/Project.toml`
  [cd998857] Yota v0.8.0
```
````

**Review by mcabbott — COMMENTED** [review]
Had a quick go with 0.8, and still see many errors? But will update a few things so long.

File: test/destructure.jl [hunk 7]
```diff
@@ -104,6 +104,32 @@ end
...
+    # These are all broken!
+    @test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], ZeroTangent())
+    @test Yota_gradient(m -> destructure(m)[1][1], m3)[1] == (x = [1,0,0], y = ZeroTangent(), z = [0,0,0])
+    @test Yota_gradient(m -> destructure(m)[1][2], m4)[1] == (x = [0,1,0], y = ZeroTangent(), z = [0,0,0])
+
+    g5 = Yota_gradient(m -> destructure(m)[1][3], m5)[1]
+    @test g5.a[1].x == [0,0,1]
+    @test g5.a[2] === ZeroTangent()
```
mcabbott, line 118:
```suggestion
    @test Yota_gradient(m -> destructure(m)[1][1], m1)[1] == [1,0,0]
    @test Yota_gradient(m -> destructure(m)[1][2], m2)[1] == ([0,1,0], [0,0,0])
    @test Yota_gradient(m -> destructure(m)[1][3], (m1, m1))[1] == ([0,0,1], nothing)
    @test Yota_gradient(m -> destructure(m)[1][1], m3)[1] == (x = [1,0,0], y = nothing, z = [0,0,0])
    @test Yota_gradient(m -> destructure(m)[1][2], m4)[1] == (x = [0,1,0], y = nothing, z = [0,0,0])

    g5 = Yota_gradient(m -> destructure(m)[1][3], m5)[1]
    @test g5.a[1].x == [0,0,1]
    @test g5.a[2] === nothing
```

test/destructure.jl
```diff
@@ -104,6 +104,32 @@ end
...
+    @test g6.a isa Vector{Float64}
+    @test g6.b == [0+im]
+
+    g8 = Yota_gradient(m -> sum(abs2, destructure(m)[1]), m8)[1]
+    @test g8[1].x == [2,4,6]
+    @test g8[2].b.x == [8]
+    @test g8[3] == [[10.0]]
+
+    g9 = Yota_gradient(m -> sum(sqrt, destructure(m)[1]), m9)[1]
+    @test g9.c === ZeroTangent()
```
mcabbott, line 131:
```suggestion
    @test g9.c ===nothing
```

test/destructure.jl
```diff
@@ -149,6 +175,36 @@ end
...
+    @test Yota_gradient(x -> re7(x).a[2][3], rand(3))[1] == [0,0,1]
+    @test Yota_gradient(x -> re7(x).b[2][2], rand(3))[1] == [0,0,0]
+    @test Yota_gradient(x -> re7(x).c[2][1], rand(3))[1] == [0,0,0]
+
+    v8, re8 = destructure(m8)
+    @test_broken Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]  # MethodError: no method matching zero(::Type{Any})
+    @test_broken Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]  # MethodError: no method matching !(::Expr)
+
+    re9 = destructure(m9)[2]
+    @test_broken Yota_gradient(x -> sum(abs2, re9(x).c[1]), 1:7)[1] == [0,0,0, 8,10,12,14]  # MethodError: no method matching zero(::Type{Array})
```
mcabbott, line 206:
```suggestion
    v8, re8 = destructure(m8)
    @test Yota_gradient(x -> sum(abs2, re8(x)[1].y), v8)[1] == [2,4,6,0,0]
    @test Yota_gradient(x -> only(sum(re8(x)[3]))^2, v8)[1] == [0,0,0,0,10]

    re9 = destructure(m9)[2]
    @test Yota_gradient(x -> sum(abs2, re9(x).c[1]), 1:7)[1] == [0,0,0, 8,10,12,14]
```

dfdx:
I can successfully run tests in this PR on Julia nightly with this rule added: 

```julia
function rrule(::typeof(getfield), s, f::Symbol)
  y = getproperty(s, f)
  function getproperty_pullback(dy)
      dy = unthunk(dy)
      T = typeof(s)
      nt = NamedTuple{(f,)}((dy,))
      return NoTangent(), Tangent{T}(; nt...), ZeroTangent()
  end
  return y, getproperty_pullback
end
```
Yota contains the same rule for `getproperty`, which usually is enough but doesn't work in this particular case. If the code above is an acceptable solution, I can add this rule to Yota or create a PR to ChainRules.

mcabbott:
It's possible that this package and Functors.jl should think more about whether to call `getfield` vs `getproperty`. The weird `@functor` macro https://github.com/FluxML/Functors.jl/blob/master/src/functor.jl#L11-L20 goes by `fieldnames(T)` and `getproperty` I think. 

But looking at the errors on CI, maybe it's from somewhere deeper inside, involving `Core.Box` because everything is type-unstable?
```
 Error During Test at /home/runner/work/Optimisers.jl/Optimisers.jl/test/destructure.jl:205
  Test threw exception
  Expression: (Yota_gradient((x->(sum(abs2, (re9(x)).c[1]);)), 1:7))[1] == [0, 0, 0, 8, 10, 12, 14]
  No deriative rule found for op %7 = getfield(%3, :contents)::Optimisers.Restructure{NamedTuple{(:a, :b, :c), Tuple{Vector{Float64}, Matrix{Float32}, Vector{Array}}}, NamedTuple{(:a, :b, :c), Tuple{Int64, Int64, Vector{Int64}}}} , try defining it using 
  	ChainRulesCore.rrule(::typeof(getfield), ::Core.Box, ::Symbol) = ...
``` 

That said, having a rule for `getfield` sounds fine to me. I think it should probably call ProjectTo, since this will sometimes turn e.g. `Tangent{Complex}` back into a number. And perhaps care is needed about whether closing over a symbol & a type works well, or needs `Val` to help out?
```julia
function rrule(::typeof(getfield), x::T, f::Symbol) where T
  y = getproperty(x, f)
  proj = ProjectTo(x)
  # valT = Val(T)  # perhaps more stable inside closure?
  function getfield_pullback(dy)
      nt = NamedTuple{(f,)}((unthunk(dy),))
      # not really sure whether this ought to unthunk or not, maybe ProjectTo will anyway, in which case best to be explicit?
      return NoTangent(), proj(Tangent{T}(; nt...)), ZeroTangent()
  end
  return y, getfield_pullback
end
```

mcabbott:
It's not in the tests here, but running the Metalhead example in the docs I still get this error (with or without getfield rule, 1.8 and 1.9):
```
julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
                  sum(m(x))
                   end;
ERROR: BoundsError: attempt to access Nothing at index [1]
Stacktrace:
  [1] _getfield(value::Nothing, fld::Int64)
    @ Yota ~/.julia/packages/Yota/uu3H0/src/helpers.jl:40
  [2] mkcall(::Function, ::Umlaut.Variable, ::Vararg{Any}; val::Missing, line::Nothing, kwargs::NamedTuple{(), Tuple{}}, free_kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Umlaut ~/.julia/packages/Umlaut/vGy3v/src/tape.jl:192
  [3] mkcall
    @ ~/.julia/packages/Umlaut/vGy3v/src/tape.jl:174 [inlined]
  [4] chainrules_transform!(tape::Umlaut.Tape{Yota.GradCtx})
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:183
  [5] gradtape!(tape::Umlaut.Tape{Yota.GradCtx}; seed::Int64)
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:268
  [6] gradtape(::Function, ::ResNet, ::Vararg{Any}; ctx::Yota.GradCtx, seed::Int64)
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:297
  [7] grad(::Function, ::ResNet, ::Vararg{Any}; seed::Int64)
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:367
  [8] grad(::Function, ::ResNet, ::Vararg{Any})
    @ Yota ~/.julia/packages/Yota/uu3H0/src/grad.jl:359
```

dfdx:
> But looking at the errors on CI, maybe it's from somewhere deeper inside, involving Core.Box because everything is type-unstable?

It's even curiouser! Running a random test in REPL works fine:

```julia
julia> re1 = destructure(m1)[2]
Restructure(Array, ..., 3)

julia> @test Yota_gradient(x -> re1(x)[1], rand(3))[1] == [1,0,0]
Test Passed
```
But wrap it into `@testset` and it fails!

```julia
julia> @testset "using Yota" begin
              re1 = destructure(m1)[2]
             @test Yota_gradient(x -> re1(x)[1], rand(3))[1] == [1,0,0] 
          end
          
using Yota: Error During Test at REPL[69]:3
  Test threw exception
  Expression: (Yota_gradient((x->((re1(x))[1];)), rand(3)))[1] == [1, 0, 0]
  No deriative rule found for op %3 = getfield(%1, :re1)::Optimisers.Restructure{Vector{Float64}, Int64} , try defining it using 
  
        ChainRulesCore.rrule(::typeof(getfield), ::var"#95#96"{Optimisers.Restructure{Vector{Float64}, Int64}}, ::Symbol) = ...
...
```
Perhaps, `@testset` captures the module it's running in and uses something like `getfield(captured_data, :global_var_name)`. Essentially, this is the same issue as dfdx/Yota.jl#112 .

> I think it should probably call ProjectTo [...]

Yes, it makes sense. Regarding type stability, I'm going to include your definition to Yota as is for now to keep the focus on correctness, and come back to performance later. 

> It's not in the tests here, but running the Metalhead example in the docs I still get this error (with or without getfield rule, 1.8 and 1.9):

I'm looking at it.

dfdx:
I may have spotted one of the bugs related to the failures on Metalhead example, but must make sure first. In [this piece of code](https://github.com/JuliaDiff/ChainRules.jl/blame/13ccc862899d8a3d98b09bd68edd9be8ca28197e/src/rulesets/Base/broadcast.jl#L127-L129) in generic broadcasting: 

```julia
    ys3, backs = unzip_broadcast(args...) do a...
        rrule_via_ad(cfg, f, a...)
    end
```
does `f` refer to a function being broadcasted or to the `Broadcast.broadcasted` itself? For example, in this case:

```julia
f = x -> identity(x)
args = (rand(3),)
rrule(cfg, broadcasted, f, args...)
```
which of the following is invoked:

```julia
rrule_via_ad(cfg, broadcasted, f, args...)
```
or 
```julia
rrule_via_ad(cfg, f, args...)
```
?

mcabbott:
I don't see an obvious mistake. The intention is for `rrule(cfg, broadcasted, BroadcastStyle, sqrt, [1,2,3])` to call `y, bk = rrule(cfg, sqrt, 2)` i.e. acting on the elements, no broadcasting. That gives `y=1.414`, and then the unzip gives `ys3 = [1, 1.41, 1.8]` and an array of functions. 

This `plit_bc_pullbacks(cfg, f, args...)` is never passed the `broadcasted, BroadcastStyle` parts, since it doesn't need them. So its second argument should be `sqrt`.

dfdx:
Oh, I don't think it's a mistake in the generic broadcasting, but rather in `Yota.rrule_via_ad()`! The example I'm currently testing is this:

```julia
using Flux, Yota

model = Dense(28*28, 1024, x -> identity(x))
x = rand(Float32, 28*28, 4)
grad((model, x) -> sum(model(x)), model, x)
```
which produces this nice stacktrace:

```julia
ERROR: all field arrays must have same shape
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] (::StructArrays.var"#6#7"{Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}})(ci::Vector{Function})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/structarray.jl:21
  [3] map
    @ ./tuple.jl:221 [inlined]
  [4] (StructArrays.StructArray{Tuple{Float32, Function}, 2, Tuple{Matrix{Float32}, Vector{Function}}})(c::Tuple{Matrix{Float32}, Vector{Function}})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/structarray.jl:20
  [5] (StructArrays.StructArray{Tuple{Float32, Function}})(c::Tuple{Matrix{Float32}, Vector{Function}})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/structarray.jl:97
  [6] _widenstructarray(dest::StructArrays.StructArray{Tuple{Float32, var"#25#27"}, 2, Tuple{Matrix{Float32}, Matrix{var"#25#27"}}, Int64}, i::Int64, #unused#::Type{Tuple{Float32, Function}})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:118
  [7] widen_from_type(dest::StructArrays.StructArray{Tuple{Float32, var"#25#27"}, 2, Tuple{Matrix{Float32}, Matrix{var"#25#27"}}, Int64}, i::Int64, #unused#::Type{Tuple{Float32, var"#24#26"}})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:109
  [8] widen_from_instance(dest::StructArrays.StructArray{Tuple{Float32, var"#25#27"}, 2, Tuple{Matrix{Float32}, Matrix{var"#25#27"}}, Int64}, i::Int64, el::Tuple{Float32, var"#24#26"})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:105
  [9] collect_to_structarray!(dest::StructArrays.StructArray{Tuple{Float32, var"#25#27"}, 2, Tuple{Matrix{Float32}, Matrix{var"#25#27"}}, Int64}, itr::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, ChainRules.var"#1705#1707"{YotaRuleConfig, var"#141#142"}, Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}}}}, offs::Int64, st::Tuple{CartesianIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, CartesianIndex{2}})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:77
 [10] _collect_structarray!
    @ ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:59 [inlined]
 [11] _collect_structarray(itr::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, ChainRules.var"#1705#1707"{YotaRuleConfig, var"#141#142"}, Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}}}}, elem::Tuple{Tuple{Float32, var"#25#27"}, Tuple{CartesianIndices{2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}, CartesianIndex{2}}}, ax::Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}; initializer::StructArrays.StructArrayInitializer{typeof(StructArrays.alwaysfalse), typeof(StructArrays.arrayof)})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:54
 [12] collect_structarray(itr::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, ChainRules.var"#1705#1707"{YotaRuleConfig, var"#141#142"}, Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}}}}; initializer::StructArrays.StructArrayInitializer{typeof(StructArrays.alwaysfalse), typeof(StructArrays.arrayof)})
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/collect.jl:40
 [13] StructArrays.StructArray(v::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}, ChainRules.var"#1705#1707"{YotaRuleConfig, var"#141#142"}, Tuple{Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}}}}; unwrap::typeof(StructArrays.alwaysfalse))
    @ StructArrays ~/.julia/packages/StructArrays/w2GaP/src/structarray.jl:261
 [14] StructArray
    @ ~/.julia/packages/StructArrays/w2GaP/src/structarray.jl:260 [inlined]
 [15] unzip_broadcast
    @ ~/.julia/packages/ChainRules/DUopG/src/unzipped.jl:39 [inlined]
 [16] split_bc_pullbacks(cfg::YotaRuleConfig, f::var"#141#142", args::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}})
    @ ChainRules ~/.julia/packages/ChainRules/DUopG/src/rulesets/Base/broadcast.jl:127
 [17] rrule(cfg::YotaRuleConfig, #unused#::typeof(Base.Broadcast.broadcasted), #unused#::Base.Broadcast.DefaultArrayStyle{2}, f::var"#141#142", args::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}})
    @ ChainRules ~/.julia/packages/ChainRules/DUopG/src/rulesets/Base/broadcast.jl:44
 [18] mkcall(::Function, ::YotaRuleConfig, ::Vararg{Any}; val::Missing, line::Core.LineInfoNode, kwargs::NamedTuple{(), Tuple{}}, free_kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Umlaut ~/.julia/packages/Umlaut/vGy3v/src/tape.jl:192
 [19] chainrules_transform!(tape::Tape{GradCtx})
    @ Main ~/work/Yota/src/grad.jl:184
 [20] gradtape!(tape::Tape{GradCtx}; seed::Int64)
    @ Main ~/work/Yota/src/grad.jl:271
 [21] gradtape(::Function, ::Dense{var"#141#142", Matrix{Float32}, Vector{Float32}}, ::Vararg{Any}; ctx::GradCtx, seed::Int64)
    @ Main ~/work/Yota/src/grad.jl:300
 [22] grad(::Function, ::Dense{var"#141#142", Matrix{Float32}, Vector{Float32}}, ::Vararg{Any}; seed::Int64)
    @ Main ~/work/Yota/src/grad.jl:370
 [23] grad(::Function, ::Dense{var"#141#142", Matrix{Float32}, Vector{Float32}}, ::Vararg{Any})
    @ Main ~/work/Yota/src/grad.jl:362
 [24] top-level scope
    @ REPL[26]:1
```
From the stacktrace I infer that `rrule_via_ad()` returns not what `unzip_broadcast()` expects. I made a guess that `split_bc_pullbacks()` calls `rrule_via_ad()` on `broadcasted` itself, e.g.:

```julia
julia> y, bk = rrule_via_ad(YotaRuleConfig(), broadcasted, sqrt, [1.0, 2, 3])
...
julia> y
3-element Vector{Float64}:
 1.0
 1.4142135623730951
 1.7320508075688772

julia> bk
#24 (generic function with 1 method)
```
and that `unzip_broadcast()` expects `y` and `bk` to have the same length to pack them into `StructArray`. But if you say `rrule_via_ad()` is never invoked that way, then I'm going to get a good night's sleep before the next iteration of debugging :smile:

mcabbott:
Quite the stacktrace! These lines look correct to me: The same function `f` is being passed through, and it acts on `arg` which is the result of lazy broadcasting `+`. The `broadcasted, DefaultArrayStyle` arguments are marked unused:
```julia
 [16] split_bc_pullbacks(cfg::YotaRuleConfig, f::var"#141#142", args::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}})
    @ ChainRules ~/.julia/packages/ChainRules/DUopG/src/rulesets/Base/broadcast.jl:127
 [17] rrule(cfg::YotaRuleConfig, #unused#::typeof(Base.Broadcast.broadcasted), #unused#::Base.Broadcast.DefaultArrayStyle{2}, f::var"#141#142", args::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}})
    @ ChainRules ~/.julia/packages/ChainRules/DUopG/src/rulesets/Base/broadcast.jl:44
```
To get to line 39 https://github.com/JuliaDiff/ChainRules.jl/blob/main/src/unzipped.jl#L39 the function `rrule_via_ad` must have inferred to give a Tuple which is correct. I don't see how the shape can then come out wrong, but...

dfdx:
Here's an interesting observation. If I run the same example as is:

```julia
using Flux, Yota, ChainRules


myid = x -> identity(x)
model = Dense(5, 3, myid)
x = rand(Float32, 5, 1);
val, g = grad((model, x) -> sum(model(x)), model, x)
@show val
@show g
```
I get the same stacktrace as posted above, complaining about "ERROR: all field arrays must have same shape". However, if I slightly modify `unzip_broadcast()` and just add `Broadcast.materialize(bc)`:

```julia
function unzip_broadcast(f::F, args...) where {F}
    T = Broadcast.combine_eltypes(f, args)
    if isconcretetype(T)
        T <: Tuple || throw(ArgumentError("""unzip_broadcast(f, args) only works on functions returning a tuple,
            but f = $(sprint(show, f)) returns type T = $T"""))
    end
    bc = Broadcast.instantiate(Broadcast.broadcasted(f, args...))
    bcs = Broadcast.BroadcastStyle(typeof(bc))
    if bcs isa AbstractGPUArrayStyle
        # This is a crude way to allow GPU arrays, not currently tested, TODO.
        # See also https://github.com/JuliaArrays/StructArrays.jl/issues/150
        return unzip(broadcast(f, args...))
    elseif bcs isa Broadcast.AbstractArrayStyle
        Broadcast.materialize(bc)                                             # <-- this line added
        return StructArrays.components(StructArray(bc))
    else
        return unzip(broadcast(f, args...))  # e.g. tuples
    end
    # TODO maybe this if-else can be replaced by methods of `unzip(:::Broadcast.Broadcasted)`?
end
```
The error disappears! 

The only hypothesis I have is that materialization of a broadcasted variable changes something in the global Julia state that makes it more friendly to `StructArray`, but I can't find any relevant information. 

-------
```
(ChainRules) pkg> st
Project ChainRules v1.44.5
Status `~/work/ChainRules.jl/Project.toml`
  [79e6a3ab] Adapt v3.4.0
  [d360d2e6] ChainRulesCore v1.15.3
  [34da2185] Compat v4.2.0
  [46192b85] GPUArraysCore v0.1.2
  [92d709cd] IrrationalConstants v0.1.1
  [c1ae055f] RealDot v0.1.0
  [09ab397b] StructArrays v0.6.12
  [cd998857] Yota v0.8.0 `https://github.com/dfdx/Yota.jl.git#fix-broadcast`
  [8ba89e20] Distributed
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
  [2f01184e] SparseArrays
  [10745b16] Statistics
```

mcabbott:
That is pretty odd. 

I can reproduce this, by `@eval ChainRules function unzip_broadcast(f::F, args...) where {F} ...` your definition in the repl. What's strange is that if I then `@eval` again the old code (or the old code with some printout) it still works. (The signature has not changed, the replacement code is run.) Does this mean it's some world-age problem or something?

dfdx:
Here's a hypothesis for world age problem: 

* `bc` contains a reference to a lazy function definition
* `materialize(bc)` triggers the definition and adds a new method to an existing method table
* `StructArray(bc)` w/ and w/o prior call to `materialize(bc)` thus goes different dispatch paths and hits different versions of the same function

But: 

* if I put `Base.get_world_counter()` before and after `materialize(bc)`, I see the same world number
* I don't see a function to which a new method is added, but only completely new functions that shouldn't change the dispatch path

mcabbott:
How are you adding this `materialize` line? By editing the source used for a fresh session, or by loading something while running?

dfdx:
I have a file called `_main.jl` inside `Yota/src` directory with contents similar to this:

```julia
include("core.jl")       # in its turn, core.jl includes all the files from Yota, so now Main ~ Yota

using Flux

# I think these imports are not needed anymore, but just copy pasting them 
import ChainRules: unzip_broadcast, RCR, TRI_NO, AbstractGPUArrayStyle, StructArrays
import ChainRules.StructArrays: StructArray


@eval ChainRules function unzip_broadcast(f::F, args...) where {F}
    global BC_STATE = (f, args)
    T = Broadcast.combine_eltypes(f, args)
    if isconcretetype(T)
        T <: Tuple || throw(ArgumentError("""unzip_broadcast(f, args) only works on functions returning a tuple,
            but f = $(sprint(show, f)) returns type T = $T"""))
    end
    # bc - rrule_via_ad's wrapper broadcasted to all arguments
    bc = Broadcast.instantiate(Broadcast.broadcasted(f, args...))
    bcs = Broadcast.BroadcastStyle(typeof(bc))
    if bcs isa AbstractGPUArrayStyle
        # This is a crude way to allow GPU arrays, not currently tested, TODO.
        # See also https://github.com/JuliaArrays/StructArrays.jl/issues/150
        return unzip(broadcast(f, args...))
    elseif bcs isa Broadcast.AbstractArrayStyle
        println("World age before materialize(bc): $(Base.get_world_counter())")
        # Broadcast.materialize(bc)
        println("World age after materialize(bc): $(Base.get_world_counter())")
        # global BC = bc
        return StructArrays.components(StructArray(bc))
    else
        return unzip(broadcast(f, args...))  # e.g. tuples
    end
    # TODO maybe this if-else can be replaced by methods of `unzip(:::Broadcast.Broadcasted)`?
end


function bc_test()
    myid = x -> identity(x)
    model = Dense(5, 3, myid)
    x = rand(Float32, 5, 1);
    grad((model, x) -> sum(model(x)), model, x)
end
```
Whenever I do a change, I include the whole file, thus updating all definitions from Yota + `ChainRules.unzip_broadcast`.

-------

I also noticed that the problem is fixed if I replace `rrule_via_ad()` with a dummy implementation that doesn't generate new functions:

```julia
function ChainRulesCore.rrule_via_ad(cfg::YotaRuleConfig, f, args...)
    return 1.0, dy -> (ZeroTangent(), [ZeroTangent() for _ in args]...)
end
```
In theory, I can make `rrule_via_ad()` always work as interpreter and never compile code, but it would be a high price...

mcabbott:
Focusing on these lines
```
  [4] (StructArrays.StructArray{Tuple{Float32, Function}, 2, Tuple{Matrix{Float32}, Vector{Function}}})(c::Tuple{Matrix{Float32}, Vector{Function}})

 [16] split_bc_pullbacks(cfg::Yota.YotaRuleConfig, f::var"#7#8", args::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{2}, Nothing, typeof(+), Tuple{Matrix{Float32}, Vector{Float32}}})
```
here's a smaller reproducer:
```julia
julia> using ChainRules, Yota

julia> y, bk = ChainRules.split_bc_pullbacks(Yota.YotaRuleConfig(), identity, Broadcast.broadcasted(+, [1 2; 3 4], [5, 6]));

julia> bk([7 8; 9 0])  # with identity it works fine, also sqrt
(ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), [7 8; 9 0])

julia> y, bk = ChainRules.split_bc_pullbacks(Yota.YotaRuleConfig(), x -> identity(x), Broadcast.broadcasted(+, [1 2; 3 4], [5, 6]));
ERROR: all field arrays must have same shape

(@v1.9) pkg> st Yota ChainRules
Status `~/.julia/environments/v1.9/Project.toml`
  [082447d4] ChainRules v1.44.5
  [cd998857] Yota v0.8.0
```
This seems to avoid my order-of-loading weirdness above. If I `@eval` your method with `materialize`, then it starts working. If I `@eval` again the old one without, it stops working.

mcabbott:
Some half-way steps:
```julia
julia> using ChainRules, Yota

# Easy case

julia> broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), identity, x)
       end
2×2 Matrix{Tuple{Int64, ChainRules.var"#identity_pullback#1201"}}:
 (6, identity_pullback)  (7, identity_pullback)
 (9, identity_pullback)  (10, identity_pullback)

julia> ChainRules.unzip(ans)
([6 7; 9 10], [ChainRules.var"#identity_pullback#1201"() ChainRules.var"#identity_pullback#1201"(); ChainRules.var"#identity_pullback#1201"() ChainRules.var"#identity_pullback#1201"()])

julia> broadcast(|>, [7 8; 9 0], ans[2])
2×2 Matrix{Tuple{ChainRulesCore.NoTangent, Int64}}:
 (NoTangent(), 7)  (NoTangent(), 8)
 (NoTangent(), 9)  (NoTangent(), 0)

julia> ChainRules.unzip_broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), identity, x)
       end
([6 7; 9 10], [ChainRules.var"#identity_pullback#1201"() ChainRules.var"#identity_pullback#1201"(); ChainRules.var"#identity_pullback#1201"() ChainRules.var"#identity_pullback#1201"()])

julia> broadcast(|>, [7 8; 9 0], ans[2])
2×2 Matrix{Tuple{ChainRulesCore.NoTangent, Int64}}:
 (NoTangent(), 7)  (NoTangent(), 8)
 (NoTangent(), 9)  (NoTangent(), 0)

# Now try with y -> identity(y)

julia> broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), y -> identity(y), x)
       end
2×2 Matrix{Tuple{Int64, Function}}:  ## <-- notice Function, abstract type
 (6, #21)  (7, #20)
 (9, #20)  (10, #20)

julia> ChainRules.unzip(ans)  ## notice Core.Box
([6 7; 9 10], Function[Yota.var"#21#23"(Core.Box(Yota.var"##pullback_#72#328#86"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))) Yota.var"#20#22"(Core.Box(Yota.var"##pullback_#72#328#86"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))); Yota.var"#20#22"(Core.Box(Yota.var"##pullback_#72#328#86"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))) Yota.var"#20#22"(Core.Box(Yota.var"##pullback_#72#328#86"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"())))])

julia> broadcast(|>, [7 8; 9 0], ans[2])
2×2 Matrix{Tuple{ChainRulesCore.ZeroTangent, Int64}}:
 (ZeroTangent(), 7)  (ZeroTangent(), 8)
 (ZeroTangent(), 9)  (ZeroTangent(), 0)

julia> ChainRules.unzip_broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), y -> identity(y), x)
       end
ERROR: all field arrays must have same shape

# Name the function:

julia> myid(x) = x;

julia> broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), myid, x)
       end
2×2 Matrix{Tuple{Int64, Function}}:  ## <-- looks as bad
 (6, #21)  (7, #20)
 (9, #20)  (10, #20)

julia> ChainRules.unzip_broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
         ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), myid, x)   ## now this works, with Core.Box
       end
([6 7; 9 10], Yota.var"#20#22"[Yota.var"#20#22"(Core.Box(Yota.var"##pullback_myid#334#89"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))) Yota.var"#20#22"(Core.Box(Yota.var"##pullback_myid#334#89"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))); Yota.var"#20#22"(Core.Box(Yota.var"##pullback_myid#334#89"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"()))) Yota.var"#20#22"(Core.Box(Yota.var"##pullback_myid#334#89"{ChainRules.var"#identity_pullback#1201"}(ChainRules.var"#identity_pullback#1201"())))])
```

dfdx:
Apparently, in the last example there's no error because `rrule()` for `myid` has been already compiled in the previous `broadcast()` call. Running _only_ the last statement gives the same error:

```julia
julia> ChainRules.unzip_broadcast(Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])) do x
        ChainRules.rrule_via_ad(Yota.YotaRuleConfig(), myid, x)
end
...
ERROR: all field arrays must have same shape
...
```

dfdx:
My current understanding is as follows:

* `Yota.rrule_via_ad()` generates a new `rrule()` and shifts the world age forward
* `unzip_broadcast()` doesn't immediately call `rrule_via_ad()`, but instead creates a lazy broadcasting object `bc = Broadcast.instantiate(Broadcast.broadcasted(f, args...))`, where `f` is wrapper created here:

```julia
unzip_broadcast(args...) do a...
    rrule_via_ad(cfg, f, a...)
end
```
* this broadcasted object is actually materialized only when `StructArray(bc)` is called
* when `rrule_via_ad()` itself acts on broadcasted objects, combination of double broadcasting, new function generation and something in `StructArray` constructor leads to incorrect results

Removing any of these factors solves the problem. Also, if in `unzip_broadcast()` I replace:

```julia
bc = Broadcast.instantiate(Broadcast.broadcasted(f, args...))
```
with this:
```julia
bc = broadcast(f, args...)
```
it leads to earlier evaluation and fixes the issue to. Given that a few lines later, in all 3 branches we materialize `bc` anyway:

```julia
        if bcs isa AbstractGPUArrayStyle
            # This is a crude way to allow GPU arrays, not currently tested, TODO.
            # See also https://github.com/JuliaArrays/StructArrays.jl/issues/150
            return unzip(broadcast(f, args...))
        elseif bcs isa Broadcast.AbstractArrayStyle        
            return StructArrays.components(StructArray(bc))
        else
            return unzip(broadcast(f, args...))  # e.g. tuples
        end
```
I wonder why do we need `bc` to be lazily broadcasted here at all?

mcabbott:
>  in all 3 branches we materialize bc anyway:

But one of them materialises directly two arrays, instead of allocating an array of tuples first. This path is the entire reason for this function, and for depending on StructArrays.

Cc @piever in case this weird error rings any bells. (I wonder if it's possible to hit it without AD being involved?)

mcabbott:
Digging a bit more, of course you know this, but finally I see that `rrule_via_ad` calls `eval` and `invokelatest`. Twice, for say `grad((x,y) -> sum((z -> sqrt(z)).(x)), [1,2,3], 4)` to work:
```
function ChainRulesCore.rrule_via_ad(config::YotaRuleConfig, f, args...)
    res = rrule(f, args...)   # why not pass config?
    !isnothing(res) && return res  # shortcut if there's actually a rule

    sig = map(typeof, (f, args...))
    rr = get!(GENERATED_RRULE_CACHE, sig) do
        make_rrule(f, args...)  # trace, make Expr, eval
    end

    val, pb = Base.invokelatest(rr, YOTA_RULE_CONFIG, f, args...)
    return val, dy -> Base.invokelatest(pb, dy)  # both invokelatest necc.
end
```

Is this what StructArrays has difficulty with? E.g. maybe it's using type inference to know what to allocate, and help says "(The drawback is that invokelatest is somewhat slower than calling f directly, and the type of the result cannot be inferred by the compiler.)"

piever:
Not sure if this is helpful, but here are some thoughts that could be useful.

1. One of the reasons for the mismatches could be that for StructArrays with no fields, the size of the array is hard to preserve (could be useful to test this branch with https://github.com/JuliaArrays/StructArrays.jl/pull/235, though based on the `eval` & co. issues I imagine this shouldn't be the problem).
2. StructArrays collection mechanism only uses inference when the iterator is empty.
3. There is a PR to improve broadcast for StructArrays: https://github.com/JuliaArrays/StructArrays.jl/pull/215, maybe it is helpful here (it would unify GPU and CPU implementation of `unzip_broadcast` ideally).

Though I definitely am puzzled as to why this is happening. Looks like the collection mechanism `StructArray(bc)` is failing on a lazy broadcasted object (indeed this is a different code path than `StructArray(::AbstractArray)`). An "AD-free" reproducer would definitely help narrow this down.

dfdx:
Here's a reproducible example without Yota and ChainRules:

```julia
import StructArrays
import StructArrays: StructArray

# eval a new function similar to rrule()
function make_rrule(f, args...)
    name = gensym()
    ex = :(function $name(f, args...)
        y = sum(args)
        pullback(dy) = dy + y
        return y, pullback
    end)
    return Base.eval(@__MODULE__, ex)
end

# wrap rrule-like function with required number of invokelatest()
function rrule_via_ad(f, args...)
    rr = make_rrule(f, args...)
    res = Base.invokelatest(rr, f, args...)
    y, pb_ = res
    pb = dy -> Base.invokelatest(pb_, dy)
    return y, pb
end

# original split_bc_pullbacks stripped to the bones
function split_bc_pullbacks(f::F, args::Vararg{Any,N}) where {F,N}
    wf = (a...) -> rrule_via_ad(f, a...)
    # comment/uncomment the next 2 lines to make the example fail/work
    bc = Broadcast.instantiate(Broadcast.broadcasted(wf, args...))
    # bc = broadcast(wf, args...)
    return StructArrays.components(StructArray(bc))
end

bce() = Broadcast.broadcasted(+, [1 2; 3 4], [5, 6])
split_bc_pullbacks(x -> identity(x), bce())
```
Since `make_rrule()` generates a new function on every call, this code can be run with fail in the same REPL multiple times. I think the pullback stuff can also be removed, but I want to try the aforementioned branches from StructArrays first.

mcabbott:
Stil happens on the PR's branch. You can simplify a bit further, and note that acting on a vector is OK, but higher ndims fails:
```
julia> [4,5,6] .|> split_bc_pullbacks(x -> identity(x), [1,2,3])[2]
3-element Vector{Int64}:
 5
 7
 9

julia> split_bc_pullbacks(x -> identity(x), [1 2; 3 4])[2]
ERROR: all field arrays must have same shape
```

piever:
> Trying to pick bits out of the stack trace, is this correct?
> 
> ```
> julia> mat = [(; i, f) for i in 1:3, f in (sin, sin)] |> StructArray
> 3×2 StructArray(::Matrix{Int64}, ::Matrix{typeof(sin)}) with eltype NamedTuple{(:i, :f), Tuple{Int64, typeof(sin)}}:
>  (i = 1, f = sin)  (i = 1, f = sin)
>  (i = 2, f = sin)  (i = 2, f = sin)
>  (i = 3, f = sin)  (i = 3, f = sin)
> 
> julia> StructArrays._widenstructarray(mat, 2, Tuple{Int, Function})
> 6-element StructArray(::Vector{Int64}, ::Vector{Function}) with eltype Tuple{Int64, Function}:
>     (1, sin)
>  #undef
>  #undef
>  #undef
>  #undef
>  #undef
> ```

Agh, no it isn't, well spotted! Somehow the widening mechanism was not updated to support arrays of arbitrary shape and only worked for 2D things... https://github.com/JuliaArrays/StructArrays.jl/pull/246 should hopefully fix it!

dfdx:
I can confirm JuliaArrays/StructArrays.jl#246 fixes all the issues up to my first reproducer using Flux and Yota. Thanks for the quick fix!

The Metalhead example still fails though, but that's another story, which I'm looking at now.

docs/src/index.md
```diff
@@ -72,14 +80,33 @@ This `∇model` is another tree structure, rather than the dictionary-like objec
...
+but also returns a gradient component for the loss function.
+To extract what Optimisers.jl needs, you can write `_, (_, ∇model) = Yota.grad(f, model, data)`
+or, for the Flux model above:
+
+```julia
+using Yota
+
+loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
+  sum(m(x))
+end;
```
mcabbott, line 98: I rebased this and tests pass!

This example does not, it fails with a seemingly simple error:
```
julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
          sum(m(x))
        end;


loss, (_, ∇model) = Yota.grad(m -> sum(m(image)), model)ERROR: No derivative rule found for op %454 = ntuple(%452, 4)::NTuple{4, Int64} , try defining it using 

	ChainRulesCore.rrule(::typeof(ntuple), ::Flux.var"#336#337"{4, Array{Float32, 4}}, ::Int64) = ...

Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] step_back!(tape::Umlaut.Tape{Yota.GradCtx}, y::Umlaut.Variable)
    @ Yota ~/.julia/packages/Yota/KJQ6n/src/grad.jl:219
```
That was on tagged Yota; on latest everything instead it seems to take forever, and interrupts here:
```
julia> loss, (∇function, ∇model, ∇image) = Yota.grad(model, image) do m, x
          sum(m(x))
        end;

^CERROR: InterruptException:
Stacktrace:
   [1] collect(itr::Base.Generator{Vector{Umlaut.Variable}, Yota.var"#68#72"{Umlaut.Tape{Yota.GradCtx}}})
     @ Base ./array.jl:792
   [2] todo_list(tape::Umlaut.Tape{Yota.GradCtx}, y::Umlaut.Variable)
     @ Yota ~/.julia/packages/Yota/5CVY7/src/grad.jl:113
   [3] #68
     @ ./none:0 [inlined]
   [4] iterate
     @ ./generator.jl:47 [inlined]
   [5] collect(itr::Base.Generator{Vector{Umlaut.Variable}, Yota.var"#68#72"{Umlaut.Tape{Yota.GradCtx}}})
     @ Base ./array.jl:787
   [6] todo_list(tape::Umlaut.Tape{Yota.GradCtx}, y::Umlaut.Variable)
     @ Yota ~/.julia/packages/Yota/5CVY7/src/grad.jl:113
   [7] #68
     @ ./array.jl:0 [inlined]
   [8] iterate
     @ ./generator.jl:47 [inlined]
   [9] collect_to!(dest::Vector{Vector{Umlaut.Variable}}, itr::Base.Generator{Vector{Umlaut.Variable}, Yota.var"#68#72"{Umlaut.Tape{Yota.GradCtx}}}, offs::Int64, st::Int64)
     @ Base ./array.jl:845
  [10] collect_to_with_first!(dest::Vector{Vector{Umlaut.Variable}}, v1::Vector{Umlaut.Variable}, itr::Base.Generator{Vector{Umlaut.Variable}, Yota.var"#68#72"{Umlaut.Tape{Yota.GradCtx}}}, st::Int64)
     @ Base ./array.jl:823
  [11] collect(itr::Base.Generator{Vector{Umlaut.Variable}, Yota.var"#68#72"{Umlaut.Tape{Yota.GradCtx}}})
     @ Base ./array.jl:797
--- the last 10 lines are repeated 2 more times ---

(jl_aZPcXz) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_aZPcXz/Project.toml`
  [dbeba491] Metalhead v0.8.0-DEV `https://github.com/FluxML/Metalhead.jl.git#master`
  [3bd65402] Optimisers v0.2.10 `~/.julia/dev/Optimisers`
  [09ab397b] StructArrays v0.6.13 `https://github.com/JuliaArrays/StructArrays.jl.git#master`
  [cd998857] Yota v0.8.1 `https://github.com/dfdx/Yota.jl.git#main`
```

dfdx: Hmm, I was indeed investigating incredibly long processing time, but profiler blamed type inference/abstract interpreter, so I started a long search for a better way to trace functions (e.g. see my [recent post](https://discourse.julialang.org/t/minimal-self-containing-example-of-code-transformation-using-generated-functions/88535/2) on Discourse). However, your stacktrace implies the problem may actually appear _after_ the tracing. I will try to investigate this option too closer to the end of the week.

dfdx: FYI: I opened an [issue](https://github.com/dfdx/Yota.jl/issues/131) to track this.

dfdx: Fixed. The `ResNet(18)` example now compiles and runs in 61 second (compared to 47 seconds with Zygote). Subsequent calls take ~0.4 seconds on my CPU.

mcabbott: Great, I see something similar locally, on 0.8.2

ToucheSir:
Are the failures on nightly easy to resolve?

dfdx:
It's a failure in CompilerPluginTools.jl, which apparently has not been adapted for Julia 1.9 yet. I opened JuliaCompilerPlugins/CompilerPluginTools.jl#8 to track it.

mcabbott:
Should we just skip tests on nightly, so that this can go in?

@dfdx do you know whether 1.9 works?

dfdx:
It looks like there's more work to do in CompilerPluginTools.jl to make it work on Julia 1.9, so I don't think it will happen in the nearest time. If we can skip Yota tests for Julia 1.9, it should be the most efficient solution for now.

Note that Julia nightly now points to Julia 1.10, so perhaps we need a separate entry for the 1.9.

mcabbott:
Tests with Yota are now skipped for 1.9 & up. 

Should be ready to go. Can someone approve?

ToucheSir (approved):

PR #106, opened 2022-08-28
Allow shared parameters, take III

### LINKED ISSUE #97: "Optimisers.jl does not at present handle tied weights, sorry."
Hi, and thanks for the hard work on Flux!
Do we have an idea what is causing this error?

mcabbott:
Another take on #100. Borrows the idea of making `Leaf` mutable.

~~Tries~~Tried to be simpler by pushing more of the recursion onto `fmap`:
* `setup` is just `fmapstructure` really. Its notion of sharing is thus exactly the one of Functors, one source of truth. We should fix that not to share `isbits` types, eventually.
* `update!` is just `fmap`. Much of the complication of the old walk was to reconstruct both the state and the model on the way out. But this isn't needed if Leaf is mutated. 

Tests from #100 pass with first commit. However, the shared Leaves must always match shared Arrays. It's possible that this scenario can be done even more simply, possibly without mutable Leaf. 

What #100 does is instead to take shared Leaves as the truth about parameter sharing, which some future API could set in a way not matching the model (for ImmutableArrays, etc) even though present `setup` will not. Then `update!` cannot just be `fmap`, and needs one more separate `IdDict` for the parameters. Second commit here  e84b61b bolts that on, and adds a test of it (which also pass using https://github.com/FluxML/Optimisers.jl/pull/100). But it's a bit ugly.

Edit: Third commit 0de29e1 instead just replaces the walk used for `fmap(f, tree, x)` to use `re` from its 2nd argument, while Functors still uses the cache on the 1st argument. That's tidier.

But the state tree contains the the same `()` at every non-parameter node, and Functors caches the results of these... we should fix this upstream? A possible hack for now would be to supply a special cache `IdDict{Leaf}` which cannot store anything else -- done in e17e474.

But... that's still not right. If there are mutable layer structs, then I think you cannot rely on the ID of mutable `Leaf` to tie things. So I gave up on customising `fmap` and wrote out the recursion using `(x,Leaf())` as the key for reconstruction.

---

Gradient accumulation uses an IdDict as in #100, but ~~stores a `broadcasted` adding the pieces. Which it thus requires all `apply!` methods to accept. They all do.~~ Changed to eager addition.

~~Does not at present allow for more than one derivative. But no rules use that.~~ Added. There were no tests it seems.

Fixes the bug noted in #100 that `update` could in fact mutate the state. Does this by just saying `@functor Leaf`. Added a test.

One further possibility with a mutable `Leaf` is that if can easily have a flag to mark some parameters as temporarily frozen. ~~This is implemented here (with no API to set the flag). Not sure it's what we want though. Easy to remove but perhaps if we're changing the struct we should consider other changes we might want.~~

Because `setup` does not call itself in recursion, it is fairly easy to add a warning if the model has no parameters. This was something someone complained about, I forget where.

---

Closes https://github.com/FluxML/Optimisers.jl/pull/42, closes #100, closes #97

### DIFF
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -13,15 +13,15 @@ To change just the learning rate, provide a number `η::Real`.
 julia> m = (vec = rand(Float32, 2), fun = sin);
 
 julia> st = Optimisers.setup(Nesterov(), m)  # stored momentum is initialised to zero
-(vec = Leaf(Nesterov{Float32}(0.001, 0.9), Float32[0.0, 0.0]), fun = nothing)
+(vec = Leaf(Nesterov{Float32}(0.001, 0.9), Float32[0.0, 0.0]), fun = ())
 
 julia> st, m = Optimisers.update(st, m, (vec = [16, 88], fun = nothing));  # with fake gradient
 
 julia> st
-(vec = Leaf(Nesterov{Float32}(0.001, 0.9), Float32[-0.016, -0.088]), fun = nothing)
+(vec = Leaf(Nesterov{Float32}(0.001, 0.9), Float32[-0.016, -0.088]), fun = ())
 
 julia> st = Optimisers.adjust(st, 0.123)  # change learning rate, stored momentum untouched
-(vec = Leaf(Nesterov{Float32}(0.123, 0.9), Float32[-0.016, -0.088]), fun = nothing)
+(vec = Leaf(Nesterov{Float32}(0.123, 0.9), Float32[-0.016, -0.088]), fun = ())
 ```
 
 To change other parameters, `adjust` also accepts keyword arguments matching the field
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -47,8 +47,8 @@ adjust(tree; kw...) = map(st -> adjust(st; kw...), tree)
 adjust(::Nothing, ::Real) = nothing
 adjust(::Nothing; kw...) = nothing
 
-adjust(ℓ::Leaf, eta::Real) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule, eta), ℓ.state, ℓ.frozen)
-adjust(ℓ::Leaf; kw...) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule; kw...), ℓ.state, ℓ.frozen)
+adjust(ℓ::Leaf, eta::Real) = Leaf(adjust(ℓ.rule, eta), ℓ.state)
+adjust(ℓ::Leaf; kw...) = Leaf(adjust(ℓ.rule; kw...), ℓ.state)
 
 
 """
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -47,8 +47,8 @@ adjust(tree; kw...) = map(st -> adjust(st; kw...), tree)
 adjust(::Nothing, ::Real) = nothing
 adjust(::Nothing; kw...) = nothing
 
-adjust(ℓ::Leaf, eta::Real) = Leaf(adjust(ℓ.rule, eta), ℓ.state)
-adjust(ℓ::Leaf; kw...) = Leaf(adjust(ℓ.rule; kw...), ℓ.state)
+adjust(ℓ::Leaf, eta::Real) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule, eta), ℓ.state, ℓ.frozen)
+adjust(ℓ::Leaf; kw...) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule; kw...), ℓ.state, ℓ.frozen)
 
 
 """
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -20,14 +20,26 @@ end
 Base.:(==)(a::Leaf, b::Leaf) = children(a) == children(b)
 
 function setup(rule::AbstractRule, model)
-  cnt = Ref(0)
-  # Rely on Functors to identify shared arrays, they will share a Leaf in this tree:
-  tree = fmapstructure(model, exclude = isnumeric) do x
-    cnt[] += 1
-    Leaf(rule, init(rule, x))
-  end
-  cnt[] == 0 && @warn "setup found no parameters in the given model"
+  cache = IdDict()
+  tree = _setup(rule, model; cache)
+  isempty(cache) && @warn "setup found no trainable parameters in this model"
   tree
+end
+
+# _setup is almost fmapstructure, but needs a _trainable_walk, and a cache which ignores numbers etc.
+function _setup(rule, x; cache)
+  haskey(cache, x) && return cache[x]
+  if isnumeric(x)
+    ℓ = Leaf(rule, init(rule, x))
+    if isbits(x)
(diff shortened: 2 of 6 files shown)


ToucheSir:
> A possible hack for now would be to supply a special cache `IdDict{Leaf}` which cannot store anything else.

I'd say this is less of a hack and something we should be doing more often. Either define a custom cache type, or (better), attach the cache to the callback itself by memoizing it. Then `fmap` and the rest of Functors can avoid cache management altogether.

src/adjust.jl
```diff
@@ -47,8 +47,8 @@ adjust(tree; kw...) = map(st -> adjust(st; kw...), tree)
 adjust(::Nothing, ::Real) = nothing
 adjust(::Nothing; kw...) = nothing
 
-adjust(ℓ::Leaf, eta::Real) = Leaf(adjust(ℓ.rule, eta), ℓ.state)
-adjust(ℓ::Leaf; kw...) = Leaf(adjust(ℓ.rule; kw...), ℓ.state)
+adjust(ℓ::Leaf, eta::Real) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule, eta), ℓ.state, ℓ.frozen)
+adjust(ℓ::Leaf; kw...) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule; kw...), ℓ.state, ℓ.frozen)
```
ToucheSir, line 51: Not sure these things are intrinsically tied. IMO would benefit from more discussion in a separate PR.

mcabbott: Yes. Just doodling really, and wondering if mutable Leaf makes things easier.

ToucheSir: I don't think it is, since we don't include it at all in the public documentation. Even if we did, changes to non-documented fields and adding fields is non-breaking AIUI.

mcabbott: Frozen is removed in f046185

src/interface.jl
```diff
@@ -1,57 +1,113 @@
...
-  else
-    return map(xᵢ -> setup(rule, xᵢ; seen), _trainable(x))
+@functor Leaf
+
+Base.:(==)(a::Leaf, b::Leaf) = children(a) == children(b)
+
+function setup(rule::AbstractRule, model)
+  cnt = Ref(0)
+  # Rely on Functors to identify shared arrays, they will share a Leaf in this tree:
+  tree = fmapstructure(model, exclude = isnumeric) do x
```
mcabbott, line 25: It's pretty surprising tests pass with this, as it doesn't check `trainable` at all.

src/interface.jl
```diff
@@ -1,57 +1,116 @@
...
+    x # no gradient seen
+  end
+end
+
+subtract!(x, x̄) = maywrite(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
+
+_grads!(dict::IdDict, ℓ::Leaf, x, ::Zero...) = nothing
+function _grads!(dict::IdDict, ℓ::Leaf, x, x̄s...)
+  x̄s₀ = get(dict, ℓ, map(_ -> ZeroTangent(), x̄s))
+  dict[ℓ] = map(+, x̄s, x̄s₀)  # adding Zero should be free. Lazy accumulation broadcasted(+, x̄, x̄₀) also possible.
```
ToucheSir, line 95:
```suggestion
  x̄s₀ = get(ZeroTangent, dict, ℓ)
  dict[ℓ] = x̄s + x̄s₀  # adding Zero should be free. Lazy accumulation broadcasted(+, x̄, x̄₀) also possible.
```
https://github.com/JuliaDiff/ChainRulesCore.jl/blob/v1.15.3/src/tangent_arithmetic.jl#L99-L100 tells me and testing confirms that a mapped/broadcasted add on an array of `ZeroTangent`s is not a no-op.

src/interface.jl
```diff
@@ -1,57 +1,116 @@
...
+  tree, newmodel  # note that tree is guaranteed to be updated. Also that it's not necc a tree.
 end
 
-function update(tree, x, x̄s...)
-  t′ = fmap(copy, tree; exclude = maywrite)
-  x′ = fmap(copy, x; exclude = maywrite)
-  update!(t′, x′, x̄s...)
+function _update!(tree, x; grads, params)
+  haskey(params, (tree,x)) && return params[(tree,x)]
+  isbits(tree) && return x  # means () is not cached, and also (((),),)
```
ToucheSir, line 75: This does imply we will be caching almost every level of an average Flux model (since `BitsType{NotBits, BitsTypes...}` is not a bitstype). `objectid` being not the fastest function in the world, perhaps both cache lookup and insertion should be guarded by `!ismutable(x)`.

src/interface.jl
mcabbott: Yes. Note that the map is over the tuple of higher derivatives (if they every exist) BTW. So this should be free.

src/interface.jl
mcabbott: I wondered this too. For large ImmutableArrays this may eventually need something fancier. But for now I think every `fmap` walk does the same thing.

ToucheSir: Oh I wasn't even thinking about those, but cases like https://github.com/JuliaLang/julia/issues/43542. We're unlikely to see any pathological behaviour, but I have to imagine the single comparison `ismutable` makes is more efficient than the recursive hash function `objectid` uses.

mcabbott: OK. I guess `ismutable` really is right here. For parameter arrays IIRC there was a concern that it tells you e.g. that PermutedDimsArray is immutable. But for known non-leaf types, maybe it's always right?

ToucheSir: Good point. `PermutedDimsArray` at least does implement `functor`, but you can always find an array wrapper which hasn't. Perhaps then the check should be `isleaf` instead? The `isbits` check is still useful either way.

mcabbott: I think anything `isnumeric` should have a corresponding `Leaf` and hit the `_update!(::Leaf, x; ...)` method.

This one wants only to deal with mutable non-leaf things, like my `mutable struct MutTwo` example. Which makes me think that `ismutable` is fine -- we have `Foo(MutTwo(Bar(Transpose(Array`, then the Array is leaf, and the only level at which it's worthwhile for this method to cache anything is the `MutTwo` one. If this whole stack appears twice, a fresh new `struct Foo` cannot be distinguished from the old one.

mcabbott:
Shall we do this?

I don't love it, and feel a bit bad about re-writing #100 in order to understand it... but this does add some features in the end. 

But I do think we ought to handle shared parameters, and that we want mutable `Leaf` for other reasons too. 

We can re-write the internals if https://github.com/FluxML/Functors.jl/pull/43 or something allows for a prettier version. The tests are pretty good.

Maybe `if isbits(x)` should be `if !Functors.anymutable(x)` from https://github.com/FluxML/Functors.jl/pull/39 . Or a copy of that function if it's only in Functors 0.4 & we don't want to wait.

ToucheSir:
I have no objections assuming we're not considering any behavioural changes after those Functors PRs are merged.

darsnack:
I am also okay with doing this

PR #109, opened 2022-08-30
Add a logo?

mcabbott:
This adds a logo to the docs, because they looked a bit plain, like so:

Screenshot 2022-08-29 at 20 10 16

and to the readme, like so:

Screenshot 2022-08-29 at 20 08 21

I guess this implies that the arrow-through-three-squares thing means FluxML not Flux.jl. This makes some sense in that Flux has been split into many bits... but RFC I guess?


darsnack (approved):
Seems like a good idea to me

PR #110, opened 2022-08-31
Add tests for higher order interface

darsnack:
This adds some tests that should have been added as part of #16. I actually caught a bug and had to fix it.

I have intentionally did not update the documentation. For anyone reading this PR: *higher order optimizers are still not supported.* The only purpose here is make sure we could support them in theory at the interface level. The tests added only ensure that the basic behavior based on this works. There are still no higher order rules.

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -50,7 +50,7 @@ function update(tree, x, x̄s...)
 end
 
 # default all rules to first order calls
-apply!(o, state, x, dx, dxs...) = apply!(o, state, x, dx)
+apply!(o, state, x, dx, dx2, dxs...) = apply!(o, state, x, dx)
 
 """
     isnumeric(x) -> Bool
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,6 +13,19 @@ struct TwoThirds a; b; c; end
 Functors.@functor TwoThirds (a, c)
 Optimisers.trainable(x::TwoThirds) = (a = x.a,)
 
+struct DummyHigherOrder <: AbstractRule end
+
+Optimisers.init(::DummyHigherOrder, x::AbstractArray) =
+  (ones(eltype(x), size(x)), zero(x))
+
+dummy_update_rule(st, p, dx, dx2) = @. p - (st[1] * dx + st[2] * dx2)
+function Optimisers.apply!(::DummyHigherOrder, state, x, dx, dx2)
+  a, b = state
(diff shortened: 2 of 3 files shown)


src/interface.jl
```diff
@@ -50,7 +50,7 @@ function update(tree, x, x̄s...)
 end
 
 # default all rules to first order calls
-apply!(o, state, x, dx, dxs...) = apply!(o, state, x, dx)
+apply!(o, state, x, dx, dx2, dxs...) = apply!(o, state, x, dx)
```
mcabbott, line 53: What went wrong without this?

test/runtests.jl
```diff
@@ -220,6 +233,33 @@ Optimisers.trainable(x::TwoThirds) = (a = x.a,)
...
+      stin = Optimisers.setup(o, psin)
+      stout, psout = Optimisers.update(stin, psin, dxs, dx2s)
+
+      # hardcoded rule behavior for dummy rule
+      @test psout[1] == dummy_update_rule(stin[1].state, psin[1], dxs[1], dx2s[1])
+      @test psout[2] == dummy_update_rule(stin[2].state, psin[2], dxs[2], dx2s[2])
+      @test stout[1].state[1] == stin[1].state[1] .+ 1
+      @test stout[2].state[2] == stin[2].state[2] .+ 1
+
+      # error cases for mismatched arguments
```
mcabbott, line 252:
```suggestion
      # error if only given one derivative
```

test/runtests.jl
```diff
@@ -220,6 +233,33 @@ Optimisers.trainable(x::TwoThirds) = (a = x.a,)
...
+      @test_throws MethodError Optimisers.update(stin, psin, dxs)
+
+      # first order rules compose
```
mcabbott, line 255:
```suggestion
      # first-order rules compose with second-order
```

mcabbott (approved):
This looks fine. It's more code than this:

https://github.com/FluxML/Optimisers.jl/pull/106/files#diff-3b9314a6f9f2d7eec1d0ef69fa76cfabafdbe6d0df923768f9ec32f27a249c63R19-R24

but also tests composition which is nice.

src/interface.jl
darsnack: A stack overflow when you call a higher order rule with too few derivatives. This dispatch should really only be taken when the number of higher order terms is > 0.

PR #111, opened 2022-10-13
Remove negative space from css

mcabbott:
Matches Flux after changes in https://github.com/FluxML/Flux.jl/pull/2069

Should ideally move to https://github.com/FluxML/.github perhaps?


ToucheSir (approved):

ToucheSir:
How would we use them post-move?

Issue #107, opened 2022-08-28
Frozen parameters


mcabbott:
It would be nice to be able to temporarily exclude some parameters from training. 
(Edit: I forgot that there is https://github.com/FluxML/Flux.jl/issues/1931, now folded in here.)

1. One mechanism is to alter `Leaf` to record whether is is frozen. This is what #49 does, and what #106 suggests as an aside. The former is immutable, changed by walking & re-building.  The latter makes `Leaf` mutable (for other reasons) so this can be changed in place. (Edit: implemented in https://github.com/FluxML/Optimisers.jl/pull/112, merged.)

2. Another mechanism would be to insert some `Frozen` struct into the state tree which stops further exploration. This may make it easier to freeze a whole branch. But will result in a tree with different labels to the model, some pattern like `model.layers[1].layers[1].weight` will no longer translate directly to one for the state tree.

3. A similar struct could equally be inserted into the model not the state. Or into both. Since gradient calculation never sees the state, changing the model may allow for faster gradients. Does Optimisers.jl own the `struct Frozen`, if it is to recognise it?

Maybe independently, it needs a friendly way to set & remove these labels.

4. PR #49 proposes that you give an address like `freeze(state, (:layers, 1, :enc, 3))`. It seems a bit awkward to require you to know all the field names from the root.

5. It would also be possible to work based on just one field name: `freeze(state, :enc)` acts on anything within any field called `enc` (which in practice is some `Chain(enc = ..., dec = ...)`). Likewise `freeze(state, :bias)` could affect every layer.

6. Another possibility is to allow control based on the type in the model. Then it has to walk both, `state = freeze(state, model, cond)` or perhaps `state = freeze(f, state, model)` where `f` is a `do` block which tests `x isa Dense` or whatever. Doesn't lend itself so obviously to freezing only some fields, `enc` or `bias`... unless `f` returns not a Bool but a list of fields, like `x isa Chain && return :enc`.

7. If the modification is to the model, then 6. becomes `model = freeze(f, model)`.

8. If Leaf is mutable, then instead of an address you can just pass a part of the tree: `freeze!(tree.layers[1].enc[3])`, after confirming that `model.layers[1].enc[3]` is the part you want. (Edit: implemented as #112, merged.)

There's a related API question for shared weights. At present Flux (and Functors) rely on objectid. This won't work for immutable arrays.

[...]

PR #112, opened 2022-10-13
Add `freeze!`/`thaw!`

mcabbott:
Replaces #49 with something much simpler: now that `Leaf` is mutable, you can simply pass it the appropriate limb of the state tree, rather than having some explicit address mechanism.

Addresses some of #107

### PR Checklist

- [x] Tests are added

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -138,6 +138,31 @@ Optimisers.trainable(x::Layer) = (; alpha = x.alpha)  # must be a subset of chid
 st = Optimisers.setup(DecayDescent(0.1), Layer(3))
 ```
 
+## Frozen Parameters
+
+To temporarily prevent training from affecting some parameters,
+use [freeze!](@ref Optimisers.freeze!) and `thaw!`.
+They work by mutating all `Leaf`s of the state tree, within the given limb:
+
+```julia
+using Flux, Optimisers
+
+x = randn(Float32, 28, 28, 1, 1);
+net = @autosize (size(x)...,) Chain(
+  Conv((3, 3), 1 => 3, stride=2, bias=false), Flux.flatten, Dense(_ => 2, relu),
+)
+opt = Optimisers.setup(Optimisers.Momentum(), net);
+
+net.layers[3] isa Dense  # now freeze this layer's parameters:
+Optimisers.freeze!(opt.layers[3])
+
+Optimisers.update!(opt, net, gradient(m -> sum(m(x)), net)...);
+
+opt  # bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = true)
+
+Optimisers.thaw!(opt)
+```
+
 ## Tied Parameters
 
 If the same array appears twice (or more) in the model, [Functors.jl](https://fluxml.ai/Functors.jl) should recognise this.
@@ -159,7 +184,7 @@ st.layers.enc.layers[1].weight === st.layers.dec.layers[1].weight.parent  # true
 This identification relies on `===`, and will work for ordinary `Array`s and `CuArray`s.
 It will not at present work for `reshape`d arrays, nor for immutable arrays such as those
 from StaticArrays.jl.
- 
+
  
 ## Obtaining a flat parameter vector
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -155,13 +155,14 @@ opt = Optimisers.setup(Optimisers.Momentum(), net);
 
 net.layers[3] isa Dense  # now freeze this layer's parameters:
 Optimisers.freeze!(opt.layers[3])
+opt.layers[3].bias  # confirm: Leaf(Momentum(...), [0.0, 0.0], frozen = true)
 
 Optimisers.update!(opt, net, gradient(m -> sum(m(x)), net)...);
 
-net.layers[3].bias  # stil zero, and its momentum is zero too:
-opt  # bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = true)
+net.layers[3].bias  # stil zero, and its momentum is too:
 
 Optimisers.thaw!(opt)
+opt.layers[3].bias  # Leaf(Momentum(...), [0.0, 0.0])
 ```
 
 ## Tied Parameters
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -142,7 +142,7 @@ st = Optimisers.setup(DecayDescent(0.1), Layer(3))
 
 To temporarily prevent training from affecting some parameters,
 use [freeze!](@ref Optimisers.freeze!) and `thaw!`.
-They work by mutating all `Leaf`s of the state tree, within the given limb:
+They work by mutating all `Leaf`s of the state tree, or part of it.
 
 ```julia
 using Flux, Optimisers
@@ -158,6 +158,7 @@ Optimisers.freeze!(opt.layers[3])
 
 Optimisers.update!(opt, net, gradient(m -> sum(m(x)), net)...);
 
+net.layers[3].bias  # stil zero, and its momentum is zero too:
 opt  # bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = true)
 
 Optimisers.thaw!(opt)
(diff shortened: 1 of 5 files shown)


src/adjust.jl
```diff
@@ -1,3 +1,56 @@
...
+
+julia> Optimisers.freeze!(s.x)
+
+julia> Optimisers.update!(s, m, (x = ([pi], 10pi), y = [100pi]));  # with fake gradient
+
+julia> m
+(x = ([1.0], 2.0), y = [-0.14159258336972558])
+
+julia> s  # Leaf(..., true) means frozen
+(x = (Leaf(Momentum{Float32}(0.01, 0.9), [0.0], true), ()), y = Leaf(Momentum{Float32}(0.01, 0.9), [3.14159]))
```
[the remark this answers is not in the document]
mcabbott: Like make it a keyword, you mean?

I'd like to make them colourful too... https://github.com/FluxML/Optimisers.jl/pull/53

src/adjust.jl
```diff
@@ -47,8 +107,8 @@ adjust(tree; kw...) = map(st -> adjust(st; kw...), tree)
 adjust(::Nothing, ::Real) = nothing
 adjust(::Nothing; kw...) = nothing
 
-adjust(ℓ::Leaf, eta::Real) = Leaf(adjust(ℓ.rule, eta), ℓ.state)
-adjust(ℓ::Leaf; kw...) = Leaf(adjust(ℓ.rule; kw...), ℓ.state)
+adjust(ℓ::Leaf, eta::Real) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule, eta), ℓ.state)
```
mcabbott, line 110: Notice BTW that `adjust` returns a new state tree which shares all of its momentum arrays with the old one. That's probably not a huge problem but seems not all that functional.

Maybe we should change it to `adjust!` as the main path? Mutable `Leaf` means this can, like `freeze!`, be applied to just one branch of the model if you wish.

docs/src/index.md
```diff
@@ -138,6 +138,31 @@ Optimisers.trainable(x::Layer) = (; alpha = x.alpha)  # must be a subset of chid
 st = Optimisers.setup(DecayDescent(0.1), Layer(3))
 ```
 
+## Frozen Parameters
+
+To temporarily prevent training from affecting some parameters,
+use [freeze!](@ref Optimisers.freeze!) and `thaw!`.
+They work by mutating all `Leaf`s of the state tree, within the given limb:
```
darsnack, line 145:
```suggestion
They work by mutating all `Leaf`s of the state tree, within the (sub-)tree passed in:
```

docs/src/index.md
```diff
@@ -138,6 +138,31 @@ Optimisers.trainable(x::Layer) = (; alpha = x.alpha)  # must be a subset of chid
...
+using Flux, Optimisers
+
+x = randn(Float32, 28, 28, 1, 1);
+net = @autosize (size(x)...,) Chain(
+  Conv((3, 3), 1 => 3, stride=2, bias=false), Flux.flatten, Dense(_ => 2, relu),
+)
+opt = Optimisers.setup(Optimisers.Momentum(), net);
+
+net.layers[3] isa Dense  # now freeze this layer's parameters:
+Optimisers.freeze!(opt.layers[3])
```
darsnack, line 157:
```suggestion
Optimisers.freeze!(opt.layers[3]) # pass a sub-tree in for freezing
```

src/adjust.jl
```diff
@@ -47,8 +107,8 @@ adjust(tree; kw...) = map(st -> adjust(st; kw...), tree)
...
+adjust(ℓ::Leaf; kw...) = ℓ.frozen ? ℓ : Leaf(adjust(ℓ.rule; kw...), ℓ.state)
```
darsnack, line 111: Do we want this? We don't typically thing of a tree of optimizers but a single optimizer per model. The fact that it is a tree is an implementation detail of Optimisers. Now, there isn't anything saying leaf rules should stay in sync, but if I schedule my LR and then thaw later on, it would be unexpected if those leaves are at a separate LR from the rest of the model.

darsnack: I think in practice it will depend on the code calling `adjust` whether this makes a difference or not.

mcabbott: No opinion on which way is better, really.

With this PR alone, this offers a way to adjust some but not all learning rates without re-building a tree with your bare hands). With #113 you can do that in one step. But I'm not sure anyone really needs to.

docs/src/index.md
mcabbott: You think people won't be thinking of 🌳 by the end of the sentence? I was going to say "branch" but that has another meaning... maybe I'd rather avoid naming things:
```suggestion
They work by mutating all `Leaf`s of the state tree, or part of it:
```

darsnack: Yeah your suggestion is best I think

src/adjust.jl
darsnack: I guess #113 is cleaner to me, because it decouples adjusting LRs from freezing w.r.t. training.

mcabbott: Yea maybe being orthogonal is simpler.

mcabbott: Removed in [9623ad7](https://github.com/FluxML/Optimisers.jl/pull/112/commits/9623ad79c6a42cacb3bee2d00b1d4a3a02f3ed98)

docs/src/index.md
mcabbott: Done.

mcabbott:
Shall we do this?

Would be nice to include in https://github.com/FluxML/Flux.jl/pull/2114 while sorting that out.

docs/src/index.md
```diff
@@ -138,6 +138,31 @@ Optimisers.trainable(x::Layer) = (; alpha = x.alpha)  # must be a subset of chid
...
+
+Optimisers.update!(opt, net, gradient(m -> sum(m(x)), net)...);
+
+opt  # bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = true)
```
ToucheSir, line 161: This would be nice as a doctest.

ToucheSir (commented):
Let's. One quick thought on top of @darsnack's comment suggestion.

docs/src/index.md
mcabbott: I tried & now everything is broken locally, I don't understand Documenter. 

Note that the tests at present don't load Flux at all. And that I think testing the output of this block probably means testing a bunch of Flux printing which I'm not sure this package should ever break if that changes.

docs/src/index.md
mcabbott: I think this is addressed by the line above? The present state avoids giving a name to what the part of a tree is.

docs/src/index.md
ToucheSir: Fair point. In that case, this line could be eliminated entirely and the comment moved to directly before/after the `freeze!` call. Also duplicating it after the `thaw!` call to show `frozen = false`. 
```julia
opt_state = Optimisers.setup(Optimisers.Momentum(), net);

...

Optimisers.freeze!(opt_state.layers[3])
# opt_state.layers[3].bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = true)

...

Optimisers.thaw!(opt)
# opt_state.layers[3].bias = Leaf(Momentum{Float32}(0.01, 0.9), Float32[0.0, 0.0], frozen = false)
...
```

mcabbott: Last commit does something like this

ToucheSir (approved):

docs/src/index.md
```diff
@@ -138,6 +138,33 @@ Optimisers.trainable(x::Layer) = (; alpha = x.alpha)  # must be a subset of chid
...
+
+To temporarily prevent training from affecting some parameters,
+use [freeze!](@ref Optimisers.freeze!) and `thaw!`.
+They work by mutating all `Leaf`s of the state tree, or part of it.
+
+```julia
+using Flux, Optimisers
+
+x = randn(Float32, 28, 28, 1, 1);
+net = @autosize (size(x)...,) Chain(
```
CarloLucibello, line 151: what's the point of splatting here instead of just using `size(x)`?

mcabbott: The macro at present wants to see a tuple expression, not a call. It could be made to accept anything.

PR #113, opened 2022-10-13
Add `adjust!`

mcabbott:
Now that `Leaf` is mutable, perhaps it is useful to be able to change `eta` only for some parts of the model.

PR also solves the following bug: `adjust` doesn't copy the momenta, and thus they remain shared with the previous object. 

### PR Checklist

- [x] Tests are added

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -165,7 +165,7 @@ Optimisers.thaw!(opt)
 opt.layers[3].bias  # Leaf(Momentum(...), [0.0, 0.0])
 ```
 
-## Adjusting Learning Rate
+## Adjusting Hyperparameters
 
 To change the learning rate during training, use [`adjust!`](@ref Optimisers.adjust!).
 This works much like `freeze!` by mutating the state tree, or part of it,
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -165,6 +165,26 @@ Optimisers.thaw!(opt)
 opt.layers[3].bias  # Leaf(Momentum(...), [0.0, 0.0])
 ```
 
+## Adjusting Learning Rate
+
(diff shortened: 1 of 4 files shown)


mcabbott:
We should do this, IMO. It fits with `freeze!`, and updating e.g. https://fluxml.ai/Flux.jl/dev/tutorials/2021-02-07-convnet/ to explicit parameters would like to use it.

Maybe `adjust` should in fact be marked deprecated, in favour of `adjust!`?

docs/src/index.md
```diff
@@ -165,6 +165,26 @@ Optimisers.thaw!(opt)
 opt.layers[3].bias  # Leaf(Momentum(...), [0.0, 0.0])
 ```
 
+## Adjusting Learning Rate
```
ToucheSir, line 168:
```suggestion
## Adjusting Hyperparameters
```
Or something similar. The thought being that this is more searchable.

test/runtests.jl
```diff
@@ -177,7 +177,7 @@ end
       @test eltype(s6[2].state[2]) == Float32
     end
 
-    @testset "adjusyting parameters" begin
+    @testset "adjusyting parameters, 1" begin
```
ToucheSir, line 180:
```suggestion
    @testset "adjusting parameters, out-of-place" begin
```

test/runtests.jl
```diff
@@ -221,6 +221,50 @@ end
       @test sc2.γ.state[2][1] ≈ [0.1, 0.2, 0.2]
     end
 
+    @testset "adjusyting parameters, 2!" begin  # Same tests, just in-place
```
ToucheSir, line 224:
```suggestion
    @testset "adjusting parameters, in-place" begin
```

ToucheSir (approved):

PR #116, opened 2022-11-14
Change `update!` to `update!!`

mcabbott:
This function makes a good-faith effort to mutate, but provides no guarantee. That's unlike `freeze!`, and unlike Base functions, which *guarantee* they will mutate. There's some convention that `!!` means this, e.g. from https://github.com/JuliaFolds/BangBang.jl, so this PR renames it.

Doing so opens the door to having a new `update!` in v0.3 which does guarantee to mutate, else fails with an error. 

That would be nice for Flux, where otherwise there's an awkward transition: v0.13's Flux.Optimise assumes everything mutates. If you upgrade to Optimisers.jl's `update!`, and discard what it returns, your model will work fine right now as all old models are mutable. But it will also now silently ignore and new immutable parameters. We can't check that you didn't discard the output. 

Doing this also opens the door to simplifying the interface. There is no need for `update!` to return the state tree, since this is sure to be mutated. And it's confusing to return two things & have to know the order. So perhaps it should return only the model? Or will it be too confusing compared to `update` which must return both?

The must-mutate `update!` could return `nothing`, to emphasise its difference. Code which goes `model, tree = update!(...)` will fail loudly and clearly with `nothing`. It may fail in very strange ways if this returns only the model. That's a reason to have 0 or 2 returns, never 1.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -33,7 +33,7 @@ Optimisers.OptimiserChain
 ```@docs
 Optimisers.setup
 Optimisers.update
-Optimisers.update!
+Optimisers.update!!
 Optimisers.adjust(::Any, ::Real)
 ```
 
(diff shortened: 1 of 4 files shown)


ToucheSir:
Backwards compat aside, is there a need for `update!` if we have `update!!`? Just to throw an error if immutable params are encountered, I guess?

PR #125, opened 2023-01-16
fix anonymous walk deprecation

### LINKED ISSUE #124: use structs instead of functions for walks in destructure
Starting julia with `julia --depwarn=yes` we get
```julia
julia> using Flux, Optimisers

julia> nn = Chain(
           Dense(1 => 3, tanh),
           Dense(3 => 1, tanh),
       );

julia> Optimisers.destructure(nn)
┌ Warning: Wrapping a custom walk function as an `AnonymousWalk`. Future versions will only support custom walks that explicitly subtype `AbstractWalk`.
│   caller = #fmap#27 at maps.jl:7 [inlined]
└ @ Core ~/.julia/packages/Functors/dFhrk/src/maps.jl:7
(Float32[-0.9446188, -0.1429313, -0.7826553, 0.0, 0.0, 0.0, -1.0964488, 0.8614038, -1.2199631, 0.0], Restructure(Chain, ..., 10))

julia> Optimisers.destructure(nn)
```
We should rewrite those closures into structs. 
Reported in https://github.com/FluxML/Functors.jl/issues/60.

CarloLucibello:
Fix #124

### DIFF
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -61,12 +61,16 @@ end
 Base.show(io::IO, re::Restructure{T}) where T = print(io, "Restructure(", T.name.name, ", ..., ", re.length, ")")
 Base.length(re::Restructure) = re.length
 
+
+struct _FlattenWalk <: AbstractWalk end
+(::_FlattenWalk)(recurse, x) = map(recurse, _trainable(x))
+
 # This flattens a model, and returns a web of offsets for later use:
 function _flatten(x)
   isnumeric(x) && return vcat(_vec(x)), 0, length(x)  # trivial case
   arrays = AbstractVector[]
   len = Ref(0)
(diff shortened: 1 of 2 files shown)


src/destructure.jl
```diff
@@ -61,12 +61,16 @@ end
 Base.show(io::IO, re::Restructure{T}) where T = print(io, "Restructure(", T.name.name, ", ..., ", re.length, ")")
 Base.length(re::Restructure) = re.length
 
+
+struct _FlattenWalk <: AbstractWalk end
+(::_FlattenWalk)(recurse, x) = map(recurse, _trainable(x))
```
mcabbott, line 66: I think this ought to be called `_TrainableStructWalk`, it doesn't flatten, just happens to be used there.

I also think that for this file's style, it ought to be defined just below `_flatten(x)`.

ToucheSir (approved):
Failure on nightly is due to Yota and not anything in this package.

PR #129, opened 2023-02-16
Add implementation of Lion optimiser

mashu:
Implementation of Lion optimiser which is faster than AdamW from [Symbolic Discovery of Optimization Algorithms](https://arxiv.org/abs/2302.06675)

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -218,6 +218,36 @@ function apply!(o::Adam, state, x, dx)
 end
 
 """
+    Lion(η = 0.001, β::Tuple = (0.9, 0.999))
+
+[Lion](https://arxiv.org/abs/2302.06675) optimiser.
+
+# Parameters
+- Learning rate (`η`): Magnitude by which gradients are updating the weights.
+- Decay of momentums (`β::Tuple`): Exponential decay for the first (β1) and the
+                                   second (β2) momentum estimate.
+"""
+struct Lion{T} <: AbstractRule
+  eta::T
+  beta::Tuple{T,T}
+end
+Lion(η = 1f-3, β = (9f-1, 9.99f-1)) = Lion{typeof(η)}(η, β)
+
+init(o::Lion, x::AbstractArray) = zero(x)
+
+function apply!(o::Lion, state, x, dx)
+  η, β = o.eta, o.beta
+
+  @.. state = β[2] * dx + (1-β[2]) * state
+
+  # The paper writes the update in terms of the old momentum,
+  # but easy to solve in terms of the current momentum instead:
+  dx′ = @lazy η * sign((β[2]-β[1]) * dx + β[1] * state)
+
+  return state, dx′
+end
+
+"""
     RAdam(η = 1f-3, β = (9f-1, 9.99f-1), ϵ = eps(typeof(η)))
 
 [Rectified Adam](https://arxiv.org/abs/1908.03265) optimizer.
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -14,7 +14,7 @@ export destructure
 include("rules.jl")
 export Descent, Adam, Momentum, Nesterov, Rprop, RMSProp,
        AdaGrad, AdaMax, AdaDelta, AMSGrad, NAdam, AdamW, RAdam, OAdam, AdaBelief,
-       WeightDecay, ClipGrad, ClipNorm, OptimiserChain
+       WeightDecay, ClipGrad, ClipNorm, OptimiserChain, Lion
 
 ###
 ### one-array functions
--- a/test/rules.jl
+++ b/test/rules.jl
@@ -8,7 +8,7 @@ RULES = [
   # All the rules at default settings:
   Descent(), Adam(), Momentum(), Nesterov(), Rprop(), RMSProp(),
   AdaGrad(), AdaMax(), AdaDelta(), AMSGrad(), NAdam(),
-  AdamW(), RAdam(), OAdam(), AdaBelief(),
+  AdamW(), RAdam(), OAdam(), AdaBelief(), Lion(),
   # A few chained combinations:
   OptimiserChain(WeightDecay(), Adam(0.001)),
   OptimiserChain(ClipNorm(), Adam(0.001)),


src/rules.jl
```diff
@@ -217,6 +217,35 @@ function apply!(o::Adam, state, x, dx)
...
+
+# Parameters
+- Learning rate (`η`): Amount by which gradients are discounted before updating
+                       the weights.
+- Decay of momentums (`β::Tuple`): Exponential decay for the first (β1) and the
+                                   second (β2) momentum estimate.
+"""
+struct Lion <: AbstractRule
+  eta::Float64
+  beta::Tuple{Float64,Float64}
```
chengchingwen, line 233:
```suggestion
struct Lion{T} <: AbstractRule
  eta::T
  beta::NTuple{2, T}
```

src/rules.jl
```diff
@@ -217,6 +217,35 @@ function apply!(o::Adam, state, x, dx)
...
+end
+Lion(η = 1f-3, β = (9f-1, 9.99f-1)) = Lion{typeof(η)}(η, β)
+
+sign(x) = x == 0 ? 0 : (x > 0 ? 1 : -1)
```
chengchingwen, line 237: `Base` already has a `sign` function, this isn't needed

mashu:
@chengchingwen I amended the commit so it should not reflect your suggestions.

src/rules.jl
```diff
@@ -217,6 +217,34 @@ function apply!(o::Adam, state, x, dx)
...
+end
+Lion(η = 1f-3, β = (9f-1, 9.99f-1)) = Lion{typeof(η)}(η, β)
+
+init(o::Lion, x::AbstractArray) = (zero(x), zero(x), o.beta)
+
+function apply!(o::Lion, state, x, dx)
+  η, β = o.eta, o.beta
+  mt, vt, βt = state
+  @.. ct = β[1] * mt + (1 - β[1]) * dx
+  dx′ = @lazy sign(ct)* η
```
mcabbott, line 243: More later. But isn't one of the advertised features of this thing that it needs less memory than Adam?

One idea would just be to let it eagerly create `@. β[1] * mt + (1 - β[1]) * dx > 0` which is a BitArray with the sign info, and then `dx′ = @lazy (2*bool-1) * η`.

mashu: Yes, it is half the memory and faster. But I wanted to stick to the original implementation and follow pseudo-code see the paper https://arxiv.org/pdf/2302.06675.pdf supplementary material "A Pseudocode for AdamW and Lion" section. 

Regarding your suggestion, I think it would change the original paper's implementation, since `sing()` can return three values -1, 0 and 1, but your suggestion is to treat 0 and -1 as -1, and have only -1 and 1, as far as I understand. Which is a bit different from what authors used.

mcabbott: Yes I guess the question is whether exactly `0.0` or `-0.0` occurs often enough to care, or approximately never.

If we do want to store 3 or 4 values, the next question is what's the smallest type which will do this. Are arrays of `Int8` fast? That's 1/4 the memory.

src/rules.jl
```diff
@@ -217,6 +217,34 @@ function apply!(o::Adam, state, x, dx)
...
+
+init(o::Lion, x::AbstractArray) = (zero(x), zero(x), zero(x), o.beta)
+
+function apply!(o::Lion, state, x, dx)
+  η, β = o.eta, o.beta
+  mt, vt, ct, βt = state
+  @.. ct = β[1] * mt + (1 - β[1]) * dx
+  dx′ = @lazy sign(ct)* η
+  @.. mt = β[2] * mt + (1 - β[2]) * dx
+  return (mt, vt, ct, βt .* β), dx′
```
mcabbott, line 245: Why does this update β?

src/rules.jl
```diff
@@ -217,6 +217,34 @@ function apply!(o::Adam, state, x, dx)
...
+- Decay of momentums (`β::Tuple`): Exponential decay for the first (β1) and the
+                                   second (β2) momentum estimate.
+"""
+struct Lion{T} <: AbstractRule
+  eta::T
+  beta::Tuple{T,T}
+end
+Lion(η = 1f-3, β = (9f-1, 9.99f-1)) = Lion{typeof(η)}(η, β)
+
+init(o::Lion, x::AbstractArray) = (zero(x), zero(x), zero(x), o.beta)
```
mcabbott, line 237: Why 3 copies? 

These are unpacked `mt, vt, ct, βt = state`, but it seems that `vt` is not used at all. 

One of the stated advantages is that this does not need two copies, like Adam does:

> It is more memory-efficient than Adam as it only keeps track of the momentum

mcabbott:
For easier reference the paper's description (including weight decay) is:
Screenshot 2023-02-19 at 23 26 07

Later, in the appendix, they have pseudo-code without weight decay:
Screenshot 2023-02-19 at 23 25 38

Note that this needs `m_{t-1}` but never `c_{t-1}`, which is why they advertise it as needing to store fewer arrays than Adam, i.e. like one not two copies of the parameters:
Screenshot 2023-02-19 at 23 35 13

src/rules.jl
mcabbott: Wait you can just analytically re-write the update in terms of the new momentum, not the old. 
```julia
init(o::Lion, x::AbstractArray) = zero(x)

function apply!(o::Lion, state, x, dx)
  η, (β1, β2) = o.eta, o.beta
  β2 > 1 && error("")

  @.. state = β2 * dx + (1-β2) * state

  # The paper writes the update in terms of the old momentum,
  # but easy to solve in terms of the current momentum instead:
  dx′ = @lazy η * sign((β1-β2) * dx + (1-β1) * state)

  return state, dx′
end
```

mashu: I kind of naively checked your suggestion comparing **Lion** (paper implementation after dropping not used states), **LionS** (temporary name for your shortcut) and **Adam**
```julia
using Flux, Metalhead, Zygote
using Optimisers: Lion, LionS

model = Metalhead.ResNet(18) |> gpu 
image = rand(Float32, 224, 224, 3, 1) |> gpu;

rule1 = Optimisers.Lion()
rule2 = Optimisers.LionS()
rule3 = Optimisers.Adam()
state1 = Optimisers.setup(rule1, model);
state2 = Optimisers.setup(rule2, model);
state3 = Optimisers.setup(rule3, model);

for i in 1:10
    ∇model, _ = gradient(model, image) do m, x  # calculate the gradients
        sum(m(x))
    end;

    state1, model1 = Optimisers.update(state1, model, ∇model);
    state2, model2 = Optimisers.update(state2, model, ∇model);
    state3, model3 = Optimisers.update(state3, model, ∇model);

    @show sum(model1(image));  # reduce
    @show sum(model2(image));
    @show sum(model3(image));

    model = model_1
end
````

I get the same result for both Lion, and later Adam diverges so it looks ok to me.

```
sum(model1(image)) = -36.56952f0
sum(model2(image)) = -36.56952f0
sum(model3(image)) = -36.56952f0
sum(model1(image)) = -1413.8202f0
sum(model2(image)) = -1413.8202f0
sum(model3(image)) = -1398.4893f0
sum(model1(image)) = -1380.8433f0
sum(model2(image)) = -1380.8433f0
sum(model3(image)) = -1369.8617f0
sum(model1(image)) = -1492.4427f0
sum(model2(image)) = -1492.4427f0
sum(model3(image)) = -1484.4536f0
sum(model1(image)) = -1477.8167f0
sum(model2(image)) = -1477.8167f0
sum(model3(image)) = -1472.7267f0
sum(model1(image)) = -1394.1106f0
sum(model2(image)) = -1394.1106f0
sum(model3(image)) = -1393.0034f0
sum(model1(image)) = -1266.7537f0
sum(model2(image)) = -1266.7537f0
sum(model3(image)) = -1266.7461f0
sum(model1(image)) = -1279.8573f0
sum(model2(image)) = -1279.8573f0
sum(model3(image)) = -1281.8655f0
sum(model1(image)) = -1269.1652f0
sum(model2(image)) = -1269.1652f0
sum(model3(image)) = -1272.0283f0
sum(model1(image)) = -1247.3923f0
sum(model2(image)) = -1247.3923f0
sum(model3(image)) = -1251.4849f0
```

File: src/rules.jl (hunk 4 above)
Comment (mashu) [reply src/rules.jl]: States updated to carry over the ones that are updated.

File: src/rules.jl (hunk 5 above)
Comment (mashu) [reply src/rules.jl]: I belive this also can be now resolved, as these are not used anymore.

**Comment by chengchingwen** [comment]
Looks like it's ready to go?

File: src/rules.jl [hunk 6]
```diff
@@ -217,6 +217,38 @@ function apply!(o::Adam, state, x, dx)
...
+  eta::T
+  beta::Tuple{T,T}
+end
+Lion(η = 1f-3, β = (9f-1, 9.99f-1)) = Lion{typeof(η)}(η, β)
+
+init(o::Lion, x::AbstractArray) = zero(x)
+
+function apply!(o::Lion, state, x, dx)
+  η, (β1, β2) = o.eta, o.beta
+  β2 > 1 && error("")
```
CarloLucibello, line 241:
```suggestion
```
this check should be done at construction time, if at all

mcabbott, line 241: This was a placeholder just because I dropped a factor `(1-β2)` inside the `sign`. Maybe it should say something more friendly:
```suggestion
  (β1 < 1 && β2 < 1) || throw(DomainError("Lion optimiser requires that both decay rates β be less than 1, but got β1 = $β1, β2 = $β2"))
```

src/rules.jl
```diff
@@ -217,6 +217,38 @@ function apply!(o::Adam, state, x, dx)
...
+"""
+    Lion(η = 0.001, β::Tuple = (0.9, 0.999))
+
+[Lion](https://arxiv.org/abs/2302.06675) optimiser.
+
+# Parameters
+- Learning rate (`η`): Amount by which gradients are discounted before updating
+                       the weights.
+- Decay of momentums (`β::Tuple`): Exponential decay for the first (β1) and the
+                                   second (β2) momentum estimate.
```
mcabbott, line 229: Are these accurate?

The learning rate is really just the size of the update, not the discount. Every parameter has one of `-η, 0, η` added to it.

src/rules.jl
```diff
@@ -217,6 +217,38 @@ function apply!(o::Adam, state, x, dx)
...
+
+function apply!(o::Lion, state, x, dx)
+  η, (β1, β2) = o.eta, o.beta
+  β2 > 1 && error("")
+
+  @.. state = β2 * dx + (1-β2) * state
+
+  # The paper writes the update in terms of the old momentum,
+  # but easy to solve in terms of the current momentum instead:
+  dx′ = @lazy η * sign((β1-β2) * dx + (1-β1) * state)
```
CarloLucibello, line 247: How do we end up with this?
What I get is that
```math
\beta_2 c_t = \beta_1  m_t + (\beta_2 - \beta_1) g_t 
```
so we can have
```
dx′ = @lazy η * sign((β2-β1) * dx + β1 * state)
```

mashu: I had some **after-thoughts** and tested this a bit comparing how updated model predictions differ with different optimizers.
Long story short, below are three apply! rules for different optimizer

Current **Michael suggestion** using current momentum $m_{t}$
```julia
function apply!(o::Lion, state, x, dx)
  η, β = o.eta, o.beta
  @.. state = β[2] * dx + (1-β[2]) * state
  dx′ = @lazy sign((β[1]-β[2]) * dx + (1-β[1]) * state) * η
  return state, dx′
end
```

My **initial implementation** (cleaned up) using previous state $m_{(t-1)}$ for update. Identical to pseudo-code, state is updated after the $dx'$ update line.
```julia
function apply!(o::LionPaper, state, x, dx)
  η, β = o.eta, o.beta
  dx′ = @lazy sign(β[1] * state + (1 - β[1]) * dx) * η
  @.. state = β[2] * state + (1 - β[2]) * dx
  return state, dx′
end

```
and finally **Carlo's suggestion**  using current momentum $m_{t}$
```julia
function apply!(o::LionCarlo, state, x, dx)
  η, β = o.eta, o.beta
  @.. state = β[2] * state + (1 - β[2]) * dx
  dx′ = @lazy sign((β[2]-β[1]) * dx + β[1] * state) * η
  return state, dx′
end
```

For refernece Adam's rule, after factoring out that bias correction so it matches the pseudo-code
```julia
function apply!(o::Adam, state, x, dx)
  η, β, ϵ = o.eta, o.beta, o.epsilon
  mt, vt, βt = state

  # update EMA of dx and abs2(dx)
  @.. mt = β[1] * mt + (1 - β[1]) * dx
  @.. vt = β[2] * vt + (1 - β[2]) * abs2(dx)
  # bias correction
  mt_ϵ = mt / (1 - βt[1])
  vt_ϵ = vt / (1 - βt[2])
  # update model parameters
  dx′ = @lazy mt_ϵ / (sqrt(vt_ϵ) + ϵ) * η

  return (mt, vt, βt .* β), dx′
end
```
To test these how they differ I decided to look at model output of the same model with the same weights, but just using each different optimizer. Because we are using Float32 the accuracy should be roughly to about 7th decimal before machine precision errors kick-in (value is a dummy loss being a sum and t is iteration)
comparision
Initially they seem to go hand in hand
comparision50

But it's more a question to @CarloLucibello and @mcabbott which derivation should be used, I am confused which one is correct.

mcabbott: Thanks for checking, yes I think I agree. 

What I had was this:
Screenshot 2023-02-22 at 07 58 30
but it was late, and this has 1-β & β backwards... corrected version is:

Screenshot 2023-02-22 at 08 02 27

src/rules.jl
mcabbott: The reason I put this check in was that I pulled 1/(1-β2) out of sign. That's wrong, it's 1/β2. 

IDK how worried we ought to be about crazy values. Must we match the paper when β=0, 1, Inf?

src/rules.jl
mashu: It's how it is described for every optimizer in the package so far.

mcabbott: Sure, but those are different. If the gradient is 100, they update by -100η, roughly -- it multiplies the original input. While here, the update isn't proportional to the gradient.

mashu: How about?
```
- Learning rate (`η`): Amount by which gradients are updating
                       the weights.
```

chengchingwen: The paper said that because Lion use `sign` as the gradient, the learning rate need to be smaller than others. Maybe worth mentioning.

CarloLucibello (approved):

PR #133, opened 2023-03-23
docs on freezing layers reworded

Mia-li88:
This PR reworded docs about freezing layers from https://fluxml.ai/Flux.jl/stable/models/advanced/#Freezing-Layer-Parameters to https://fluxml.ai/Optimisers.jl/dev/#Frozen-Parameters

### PR Checklist

- [ ] Tests are added


mcabbott:
Thanks for taking this on.

But I don't think this is the right approach. This package shouldn't really explain how Flux used to work. Such an explanation only be seen by someone who already knows to look here for the new way.

PR #134, opened 2023-03-23
fix: rm Functors@0.3

ven-k:
`AbstractWalk` was introduced in Functors@0.4.0 (https://github.com/FluxML/Functors.jl/blob/v0.4.0/src/walks.jl)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Optimisers"
 uuid = "3bd65402-5787-11e9-1adc-39752487f4e2"
 authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
-version = "0.2.16"
+version = "0.2.17"
 
 [deps]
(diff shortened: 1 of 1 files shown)


mcabbott (approved):
Should have been in #125 really. 

Is it worth yanking 0.2.16 over this? Maybe I'm asking, did you hit this in the wild because some package is holding you back to Functors@0.3, and hence 0.2.16 will break many people's installation? Or is this just tidying up?

ven-k:
Turing.jl caps Bijectors@0.10; which in turn caps Functors@0.3.
(See: https://github.com/TuringLang/Turing.jl/pull/1946)

ATM I've pinned Optimisers@0.2.15

mcabbott:
Ok. Then I think https://github.com/JuliaRegistries/General/pull/80197 ought to be merged first, and then we tag this version.

PR #137, opened 2023-04-05
Rule for gradient accumulation

### LINKED ISSUE #130: Interface for gradient accumulation
### Motivation and description

Gradient accumulation is an important technique when the model does not fit in the memory with large batch size.

### Possible Implementation

Suggested by @ToucheSir, 

> defining an `AbstractRule` for accumulation which just negates the gradient. Then when when update! is called, https://github.com/FluxML/Optimisers.jl/blob/master/src/interface.jl#L98 will cancel out the negative. To make this work, run setup as usual on the model using this grad accum rule, then use that tree to update!(accum_tree, accum_grads, new_grads)  instead. Once you've finished accumulating, just run update!(opt_state_tree, model, accum_grads) as you would without accumulation.

CarloLucibello:
Fix #130 

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -3,18 +3,7 @@ using ChainRulesCore: canonicalize, backing, Tangent, AbstractZero, ZeroTangent
 base(dx::Tangent) = backing(canonicalize(dx))
 base(dx) = dx
 
-"""
-  NoUpdate()
-
-A dummy gradient that will not be applied to the parameters.
-Can be returned by `apply!` rules to indicate that no update should be done.
-
-When returned by a rule in a [`OptimiserChain`](@ref), it will also prevent the 
-following rules from being called.
-"""
-struct NoUpdate end
-
-const Zero = Union{Nothing, AbstractZero, NoUpdate}  # Union{Zygote, Diffractor, Optimisers}
+const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
 
 abstract type AbstractRule end
 
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -1,8 +1,20 @@
-
 using ChainRulesCore: canonicalize, backing, Tangent, AbstractZero, ZeroTangent
+
 base(dx::Tangent) = backing(canonicalize(dx))
 base(dx) = dx
-const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
+
+"""
+  NoUpdate()
+
+A dummy gradient that will not be applied to the parameters.
+Can be returned by `apply!` rules to indicate that no update should be done.
+
+When returned by a rule in a [`OptimiserChain`](@ref), it will also prevent the 
+following rules from being called.
+"""
+struct NoUpdate end
+
+const Zero = Union{Nothing, AbstractZero, NoUpdate}  # Union{Zygote, Diffractor, Optimisers}
 
 abstract type AbstractRule end
 
@@ -96,7 +108,7 @@ function _update!(ℓ::Leaf, x; grads, params)
 end
 
 subtract!(x, x̄) = maywrite(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
-subtract!(x, x̄::Nothing) = x
+subtract!(x, x̄::Zero) = x
 
 _grads!(dict::IdDict, ℓ::Leaf, x, ::Zero...) = nothing
 function _grads!(dict::IdDict, ℓ::Leaf, x, x̄s...)
@@ -223,3 +235,4 @@ Broadcast.materialize(x::Lazy) = Broadcast.instantiate(x.bc)
 
 onevalue(λ::T, x::AbstractArray{T}) where T = map(_ -> λ, x)
 onevalue(λ, x::AbstractArray{T}) where T = onevalue(convert(float(T), λ), x)
+
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -96,6 +96,7 @@ function _update!(ℓ::Leaf, x; grads, params)
 end
 
 subtract!(x, x̄) = maywrite(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
+subtract!(x, x̄::Nothing) = x
 
 _grads!(dict::IdDict, ℓ::Leaf, x, ::Zero...) = nothing
 function _grads!(dict::IdDict, ℓ::Leaf, x, x̄s...)
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -630,7 +630,11 @@ With an empty sequence, `OptimiserChain()` is the identity,
 so `update!` will subtract the full gradient from the parameters.
 This is equivalent to `Descent(1)`.
 
+When `NoUpdate()` is returned by the `apply!` call for one of the optimisers, 
+it will prevent the following optimisers from being called.
+
 # Example
+
 ```jldoctest
 julia> o = OptimiserChain(ClipGrad(1.0), Descent(0.1));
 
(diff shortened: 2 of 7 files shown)


src/rules.jl
```diff
@@ -667,3 +667,39 @@ end
...
+function init(o::AccumGrad, x)
+  return (zero(x), 0, init(o.opt, x))
+end
+
+function apply!(o::AccumGrad, state, x, dx)
+  accum_dx, counter, opt_state = state
+  counter += 1
+  @.. accum_dx = accum_dx + dx
+  if counter == o.n
+    opt_state, dx′ = apply!(o.opt, opt_state, x, accum_dx ./ o.n)
```
CarloLucibello, line 698: A first version was
```suggestion
    @.. accum_dx = accum_dx / o.n
    opt_state, dx′ = apply!(o.opt, opt_state, x, accum_dx)
    @.. accum_dx = zero(accum_dx)
```
but the zeroing in the last line kills the `@lazy` gradient computed by `o.opt`, so it looks like we need to allocate a new array `accum_dx ./ o.n` but it would be nice to avoid that.

test/rules.jl
```diff
@@ -244,3 +244,27 @@ VERSION < v"1.9-" && @testset "using Yota" begin
...
+  x0 = rand(5)
+  x = copy(x0)
+  lr = 0.01
+  tree = Optimisers.setup(AccumGrad(Descent(lr), 3), x)
+
+  g1 = rand(5)
+  tree, x1 = Optimisers.update(tree, x, g1)
+  @test x1 ≈ x
+  @test x1 ≈ x0 
+  @test_broken x1 === x
```
CarloLucibello, line 258: Why this is not passing? Can we avoid making a copy?

darsnack: `update` always copies first. `update!` will not copy if the underlying storage is mutable.

src/rules.jl
```diff
@@ -667,3 +667,39 @@ end
...
+  
+  function AccumGrad(opt::R, n::Int) where R<:AbstractRule
+    n > 0 || throw(ArgumentError("AccumGrad must accumulate at least one gradient"))
+    return new{R}(opt, n)  
+  end
+end
+
+function init(o::AccumGrad, x)
+  return (zero(x), 0, init(o.opt, x))
+end
```
darsnack, line 691: This will need a custom `adjust` method

src/interface.jl
```diff
@@ -96,6 +96,7 @@ function _update!(ℓ::Leaf, x; grads, params)
 end
 
 subtract!(x, x̄) = maywrite(x) ? (x .= x .- x̄) : eltype(x).(x .- x̄)
+subtract!(x, x̄::Nothing) = x
```
mcabbott, line 99:
```suggestion
subtract!(x, x̄::Zero) = x
```

src/rules.jl
mcabbott: Maybe you can zero it at the start when count==0, rather than at the end?

test/rules.jl
```diff
@@ -244,3 +244,27 @@ VERSION < v"1.9-" && @testset "using Yota" begin
     @test loss(w, w′) < 0.001
   end
 end
+
+@testset "AccumGrad" begin
+  x0 = rand(5)
+  x = copy(x0)
+  lr = 0.01
+  tree = Optimisers.setup(AccumGrad(Descent(lr), 3), x)
```
mcabbott, line 252: A torture test would use something with a large momentum, so that it fails if the rule is applied 3 times, rather that once for the total gradient. But not sure that's necessary.

test/rules.jl
```diff
@@ -244,3 +244,27 @@ VERSION < v"1.9-" && @testset "using Yota" begin
...
+  @test x2 ≈ x
+  @test x2 ≈ x0 
+  g3 = rand(5)
+  tree, x3 = Optimisers.update(tree, x2, g3)
+  @test x3 ≈ x0 .- lr .* (g1 .+ g2 .+ g3) ./ 3
+  g4 = rand(5)
+  
+  tree, x4 = Optimisers.update(tree, x3, g4)
+  @test x4 ≈ x3
+end
```
mcabbott, line 270: In rules.jl, adding some cases to `RULES = [` will run tests of this thing with StaticArrays, complex numbers, etc.

src/rules.jl
```diff
@@ -667,3 +667,39 @@ end
...
+"""
+  AccumGrad(opt::AbstractRule, n::Int)
+
+A wrapper for an optimisation rule which accumulates gradients over `n` steps before
+applying the rule. This is useful for training with batch sizes larger than the
+available memory.
+"""
+struct AccumGrad{R<:AbstractRule} <: AbstractRule
+  opt::R
+  n::Int
```
mcabbott, line 681: The big question is whether the design should be 
```
AccumGrad(OptimiserChain(ClipGrad(), AdamW()), 3)
```
or
```
OptimiserChain(AccumGrad(3), ClipGrad(), AdamW())
```

I haven't thought too hard yet, but was the latter considered somewhere? `AccumGrad(3) => ClipGrad() => AdamW()`

CarloLucibello: > I haven't thought too hard yet, but was the latter considered somewhere?

briefly in #130, but now that I've digged a bit into it I like `OptimiserChain(AccumGrad(3), ClipGrad(), AdamW())`
more, it's nicer to read, more decoupled, and one make sense also of different permutations of the arguments.

We could just tweak the `OptimiserChain` to stop the propagation when it hits `nothing` signal. 
Or if we think that propagating `nothing` could also make sense, we could define a `NoUpdate` type.

src/interface.jl
```diff
@@ -1,8 +1,20 @@
...
+
+A dummy gradient that will not be applied to the parameters.
+Can be returned by `apply!` rules to indicate that no update should be done.
+
+When returned by a rule in a [`OptimiserChain`](@ref), it will also prevent the 
+following rules from being called.
+"""
+struct NoUpdate end
+
+const Zero = Union{Nothing, AbstractZero, NoUpdate}  # Union{Zygote, Diffractor, Optimisers}
```
mcabbott, line 17:
```suggestion
const Zero = Union{Nothing, AbstractZero}  # Union{Zygote, Diffractor}
```

src/rules.jl
```diff
@@ -654,11 +658,15 @@ init(o::OptimiserChain, x::AbstractArray) = map(opt -> init(opt, x), o.opts)
...
   foldl(tuple.(o.opts, states); init = ((), dx)) do (states′, dx′), (opt, state)
-    state′, dx′ = apply!(opt, state, x, dx′, dxs...)
+    state′, dx′ = _apply_in_chain!(opt, state, x, dx′, dxs...)
     return (states′..., state′), dx′
   end
 end
 
+_apply_in_chain!(opt, state, x, dx′, dxs...) = apply!(opt, state, x, dx′, dxs...)
+_apply_in_chain!(opt, state, x, dx′::NoUpdate, dxs...) = state, NoUpdate()
+
```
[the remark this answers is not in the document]
CarloLucibello: I agree that using a `Zero` as marker is consistent with what currently do, so maybe we should go with it in this PR. 
But what we do seems semantically wrong. Isn't `nothing` supposed to be a placeholder for `zero(x)`. Shouldn't
these two produce the same effects:
```julia
Optimisers.update!(st, x, nothing)
Optimisers.update!(st, x, zero(x))
```
?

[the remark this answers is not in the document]
mcabbott: I suspect this is more an accident than a design, but I'm not sure it's an awful one. 

If you are doing ordinary AD and happen to get an array of zeros on some batch, probably you do want that to update the momenta etc. 

But you won't get `nothing` just because of the data in that batch. Instead, you'll get it because you are e.g. doing transfer learning, or the generator & discriminator on even/odd steps, or something like that. You will get `nothing`  not for one array, but for a whole part of the model. And it seems like you probably don't want to update the momenta for the part of the model not being trained, but instead just ignore them completely.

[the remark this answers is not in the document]
CarloLucibello: I think your examples should correspond to the opt_tree only having part of the model or to using different trees for discriminator and generator. Anyways, it's better to continue this discussion somewhere else, I'll open an issue.

mcabbott (approved):

src/rules.jl
```diff
@@ -667,3 +672,60 @@ end
 
 adjust(ℓ::OptimiserChain, eta::Real) = OptimiserChain(map(opt -> adjust(opt, eta), ℓ.opts)...)
 adjust(ℓ::OptimiserChain; kw...) = OptimiserChain(map(opt -> adjust(opt; kw...), ℓ.opts)...)
+
+
+"""
+  AccumGrad(n::Int)
```
darsnack, line 678: Minor typo for rendering
```suggestion
    AccumGrad(n::Int)
```

PR #148, opened 2023-06-21
Add a variant of Adam called "PAdam"

4SAnalyticsnModelling:
Add a new variant of Adam optimizer called "PAdam"


CarloLucibello:
@4SAnalyticsnModelling can you add a test for this? See how it is done for the other optimisers

4SAnalyticsnModelling:
Will do, as soon as I can.

PR #149, opened 2023-07-12
Add a new optimizer PAdam

4SAnalyticsnModelling:
Hi @mcabbott, I am recreating this pull request by merging all proposed inter-dependent changes in different files, as you suggested. Thanks for the instructions which greatly helped :)

### DIFF
--- a/test/rules.jl
+++ b/test/rules.jl
@@ -8,7 +8,7 @@ RULES = [
   # All the rules at default settings:
   Descent(), Adam(), Momentum(), Nesterov(), Rprop(), RMSProp(),
   AdaGrad(), AdaMax(), AdaDelta(), AMSGrad(), NAdam(),
-  AdamW(), RAdam(), OAdam(), AdaBelief(), Lion(),
+  AdamW(), RAdam(), OAdam(), AdaBelief(), PAdam(), Lion(),
   # A few chained combinations:
   OptimiserChain(WeightDecay(), Adam(0.001)),
   OptimiserChain(ClipNorm(), Adam(0.001)),
@@ -181,7 +181,7 @@ end
   empty!(LOG)
   @testset "$(name(opt))" for opt in [
               # The Flux PR had 1e-2 for all. But AdaDelta(ρ) needs ρ≈0.9 not small. And it helps to make ε not too small too:
-              Adam(1e-2), RMSProp(1e-2), RAdam(1e-2), OAdam(1e-2), AdaGrad(1e-2), AdaDelta(0.9, 1e-5), NAdam(1e-2), AdaBelief(1e-2),
+              Adam(1e-2), RMSProp(1e-2), RAdam(1e-2), OAdam(1e-2), AdaGrad(1e-2), AdaDelta(0.9, 1e-5), NAdam(1e-2), AdaBelief(1e-2), PAdam(1e-2)
               # These weren't in Flux PR:
               Descent(1e-2), Momentum(1e-2), Nesterov(1e-2), AdamW(1e-2), 
               ]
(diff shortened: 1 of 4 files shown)


test/rules.jl
```diff
@@ -181,7 +181,7 @@ end
   empty!(LOG)
   @testset "$(name(opt))" for opt in [
               # The Flux PR had 1e-2 for all. But AdaDelta(ρ) needs ρ≈0.9 not small. And it helps to make ε not too small too:
-              Adam(1e-2), RMSProp(1e-2), RAdam(1e-2), OAdam(1e-2), AdaGrad(1e-2), AdaDelta(0.9, 1e-5), NAdam(1e-2), AdaBelief(1e-2),
+              Adam(1e-2), RMSProp(1e-2), RAdam(1e-2), OAdam(1e-2), AdaGrad(1e-2), AdaDelta(0.9, 1e-5), NAdam(1e-2), AdaBelief(1e-2), PAdam(1e-2),
```
mcabbott, line 184: I believe this one is causing the test error, since `max` won't work on complex numbers.

It's fine to have some rules which only work on real numbers, in which case this one can just be removed from this testset's list of cases.

4SAnalyticsnModelling: Agreed. Removed.

4SAnalyticsnModelling: @mcabbott how about getting some custom "max" function in so that both Real and Complex numbers can be handled? I am thinking about something like this

```julia
function max_(x, y)
    x_ = occursin("Complex", string(eltype(x))) ? abs(x) : x
    y_ = occursin("Complex", string(eltype(y))) ? abs(y) : y
    return x_ > y_ ? x : y
end
```

4SAnalyticsnModelling: @mcabbott how about slotting in a custom ```max_``` function instead of ```max``` function to accommodate both real and complex numbers? I am thinking about something like this
```julia
function max_(x :: T, y :: T) where T
    if eltype(x) == Complex{BigFloat} || eltype(x) == Complex{Float16} || eltype(x) == Complex{Float32} ||
    eltype(x) == Complex{Float64} || eltype(x) == Complex{BigInt} || eltype(x) == Complex{Int8} || 
    eltype(x) == Complex{Int16} || eltype(x) == Complex{Int32} || eltype(x) == Complex{Int32} ||
    eltype(x) == Complex{Int64} || eltype(x) == Complex{Int128} || eltype(y) == Complex{BigFloat} || 
    eltype(y) == Complex{Float16} || eltype(y) == Complex{Float32} || eltype(y) == Complex{Float64} || 
    eltype(y) == Complex{BigInt} || eltype(y) == Complex{Int8} || eltype(y) == Complex{Int16} || 
    eltype(y) == Complex{Int32} || eltype(y) == Complex{Int32} || eltype(y) == Complex{Int64} || 
    eltype(y) == Complex{Int128} 
        x_ = abs(x)
        y_ = abs(y)
    else
        x_ = x
        y_ = y
    end
    return x_ > y_ ? x : y
end
```
no performance or allocation issue with this custom function as compared to the base function

mcabbott: I haven't tried to understand why the rule has max there for real numbers. There might be a nearby rule for complex numbers, and going by `abs(x)` would not be a crazy variant, but I don't know if it's really natural or not, maybe something else should happen?

Unless the paper specifically works something out for complex numbers, I think leaving it an error is the least surprising thing.

4SAnalyticsnModelling: The rule has ```max``` there to have lower thresholds for the values of ```v̂```. So I thought the best way we could compare the two complex numbers was to take absolute values of those.

But leaving it as an error for complex numbers works fine for now. The paper doesn't explicitly says anything about the complex numbers.

PR #151, opened 2023-07-26
Use `eltype(x)` everywhere, ignore `typeof(η)`

mcabbott:
First commit should fix #150, by working all in Float32. ~~Chooses to do this by always following the momentum from `setup`, not the `x`, but perhaps that's messy?~~ changed to `eltype(x)` now.

Since the type of `η` is now ignored, perhaps we should just remove all the type parameters? Then e.g. `Adam(0) == Adam(0.0)`, fixes #119, closes #120. Also closes #86 (redundant). 

~~Not all rules are changed yet, RFC?~~

PR #152, opened 2023-07-26
Rule for mixed precision training

CarloLucibello:
Implements a fundamental strategy for training big models. 
In this implementation, the high precision weights are part of the optimiser's state. 

The optimiser introduced here should be coupled to a loss scaling strategy (not in this PR, and probably not ever in this repo) to obtain robust mixed precision training.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -858,8 +858,7 @@ function apply!(o::AccumGrad, state, x, dx)
 end
 
 """
-    MixedPrecision(opt)
-    MixedPrecision{T}(opt)
+    MixedPrecision([T = Float32,] opt)
 
 An optimiser that wraps another optimiser `opt` in order to perform mixed precision
 training [1]. 
@@ -879,14 +878,14 @@ then `x` is updated with the value of `xT`.
 # Examples
 
 ```julia
-x = rand(Float16, 2) # a trainable parameter in low precision
+x = rand(Float16, 2) # A trainable parameter in low precision
 
-opt = MixedPrecision(Adam(1e-3)) 
-opt_state = Optimisers.setup(opt, x) # the state contains a copy of x in Float32 precision
+opt = MixedPrecision(Adam(1e-3)) # Equivalent to MixedPrecision(Float32, Adam(1e-3))
+opt_state = Optimisers.setup(opt, x) # The state contains a copy of x in Float32 precision
 
-g = rand(Float16, 2) # a gradient in low precision
+g = rand(Float16, 2) # A gradient in low precision
 
-# accumulation is performed in high precision,
+# Accumulation is performed in high precision,
 # then also the low precision x is synced
 Optimisers.update!(opt_state, x, g)  
 ```
@@ -898,7 +897,7 @@ end
 @functor MixedPrecision
 
 MixedPrecision(opt::AbstractRule) = MixedPrecision{Float32, typeof(opt)}(opt)
-MixedPrecision{T}(opt::AbstractRule) where T = MixedPrecision{T, typeof(opt)}(opt)
+MixedPrecision(T::Type, opt::AbstractRule) = MixedPrecision{T, typeof(opt)}(opt)
 
 function init(o::MixedPrecision{T}, x::AbstractArray) where T
   xT = T.(x)
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -899,17 +899,20 @@ end
 MixedPrecision(opt::AbstractRule) = MixedPrecision{Float32, typeof(opt)}(opt)
 MixedPrecision{T}(opt::AbstractRule) where T = MixedPrecision{T, typeof(opt)}(opt)
 
-to_precision(::Type{T}, x::AbstractArray) where T = convert(AbstractArray{T}, x)
-
-function init(o::MixedPrecision{T}, x) where T
-  xT = to_precision(T, x)
+function init(o::MixedPrecision{T}, x::AbstractArray) where T
+  xT = T.(x)
   return (xT, init(o.opt, xT))
 end
 
 function apply!(o::MixedPrecision{T}, state, x, dx) where T
   xT, st = state
   st′, dx′ = apply!(o.opt, st, xT, dx)
-  subtract!(xT, dx′)
-  @. x = xT
-  return (xT, st′), nothing
+  xT = subtract!(xT, dx′)
+  if maywrite(x)
+    x .= xT
+    dx′ = nothing
+  else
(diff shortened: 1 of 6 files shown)


darsnack:
This will need a custom `adjust` too

CarloLucibello:
Good to go?

darsnack (approved):
Good for me

src/rules.jl
```diff
@@ -752,3 +752,68 @@ function apply!(o::AccumGrad, state, x, dx)
...
+  return (xT, init(o.opt, xT))
+end
+
+function apply!(o::MixedPrecision{T}, state, x, dx) where T
+  xT, st = state
+  st′, dx′ = apply!(o.opt, st, xT, dx)
+  xT = subtract!(xT, dx′)
+  if maywrite(x)
+    x .= xT
+    dx′ = nothing
```
darsnack, line 811:
```suggestion
    dx′ = Zero()
```

darsnack (commented):
One more minor thing

mcabbott:
Would like to ask for some time to read this closely. 

Haven't understood why it doesn't use OptimiserChain. Not sure that exposing `MixedPrecision{Float32}` as the official way to specify the type is so nice.

darsnack:
It’s unfortunate that the API here doesn’t use `OptimiserChain` like `AccumGrad`, but I think it is unavoidable. We need to invoke the inner optimizer’s `apply!` with higher precision which we could do by promoting `x` and `dx` as the first thing in the chain (along with #151). But then we still need `subtract!` to happen at higher precision and the result synced back to the high precision copy of `x` in `MixedPrecision`‘s state. So doing it would require a big change to `update`/`subtract!`.

It makes me think that `OptimiserChain` is not the best API in general. Yes, it makes writing the rule easier. But it doesn’t work well for cases like this, and it has a potential foot gun where you can place the rule in the wrong spot in a chain. Wrapping seems more intuitive as an API (“whatever this rule that I’m wrapping does, do that in precision `T`” vs. grokking how the update transforms in a chain).

CarloLucibello:
>  Not sure that exposing MixedPrecision{Float32} as the official way to specify the type is so nice.

Can be `MixedPrecision(Float32, opt)` instead

src/rules.jl
CarloLucibello: `nothing` is correct here, Zero is abstract: `const Zero = Union{Nothing, AbstractZero}`

mcabbott:
I have questions but haven't had time to read up more. 

First, this setup starts with the low-precision model, and temporarily stores a higher-precision copy for the purpose of accumulation without overflow. Is this standard? There's no very easy way to get back the high-precision model.

Somehow I thought the default was the reverse, to start and end with the high-precision one, and treat the low-precision model as a temporary step for cheaper gradients. This could not, I think, be implemented as a rule here.

Second, I thought it was conventional to scale the loss used for low-precision steps, and then scale the gradients. There's no sign of that here. Am I mis-remembering? Scaling the gradients could be done by composing with Descent but if it's standard then maybe it should be easy.

src/rules.jl
```diff
@@ -752,3 +752,66 @@ function apply!(o::AccumGrad, state, x, dx)
...
+  else
+    dx′ = x .- eltype(x).(xT)
+  end
+  return (xT, st′), dx′
+end
+
+adjust(o::MixedPrecision, eta::Real) = MixedPrecision(adjust(o.opt, eta))
+adjust(o::MixedPrecision; kw...) = MixedPrecision(adjust(o.opt; kw...))
```
mcabbott, line 817: Don't these forget `T`?

src/rules.jl
```diff
@@ -752,3 +752,66 @@ function apply!(o::AccumGrad, state, x, dx)
...
+
+g = rand(Float16, 2) # A gradient in low precision
+
+# Accumulation is performed in high precision,
+# then also the low precision x is synced
+Optimisers.update!(opt_state, x, g)  
+```
+"""
+struct MixedPrecision{T<:Number, O<:AbstractRule} <: AbstractRule
+  opt::O
```
mcabbott, line 790: Maybe this should be rule?
```suggestion
  rule::O
```

src/rules.jl
```diff
@@ -752,3 +752,66 @@ function apply!(o::AccumGrad, state, x, dx)
...
+call it `xT`, as well as the internal state of `opt` also at precision `T`.
+If `T` is not specified, it defaults to `Float32`.
+
+Call `g` the gradient of `x`. Both `g` and `x` are typically in a precision lower than `T`
+(e.g. `Float16`).
+
+In the `update!(opt_state, x, g)` call, `opt` is used to update `xT` instead of `x`, 
+then `x` is updated with the value of `xT`. 
+
+[1] Micikevicius et al. '17, "Mixed Precision Training", https://arxiv.org/abs/1710.03740 .
```
mcabbott, line 772:
```suggestion
# Reference

[1] Micikevicius et al., "Mixed Precision Training", ICLR 2018, https://arxiv.org/abs/1710.03740 .
```

mcabbott:
Just to sketch another possibility, Flux could instead wrap pairs of low+high precision copies of the same model:

```julia
bimodel = MixedPrec(Float16, model32)  # makes a Float16 copy, stores a scale for loss.
bimodel(x)  # calls variant matching eltype(x)? Default Float16?

opt_state = setup(Adam(), bimodel)  # special method?
gs = gradient(loss, bimodel)  # this could be a special method which scales the loss?
update!(opt_state, bimodel, gs[1]) # this knows about the scale, updates both halves.
```

CarloLucibello:
> Just to sketch another possibility, Flux could instead wrap pairs of low+high precision copies of the same model:

This is a solution to keep in mind and maybe worth pursuing in Flux. 
In any case, it is good to have a more basic block here in Optimisers.jl

src/rules.jl
```diff
@@ -752,3 +752,66 @@ function apply!(o::AccumGrad, state, x, dx)
...
+
+function apply!(o::MixedPrecision{T}, state, x, dx) where T
+  xT, st = state
+  st′, dx′ = apply!(o.opt, st, xT, dx)
+  xT = subtract!(xT, dx′)
+  if maywrite(x)
+    x .= xT
+    dx′ = nothing
+  else
+    dx′ = x .- eltype(x).(xT)
```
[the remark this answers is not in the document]
CarloLucibello: > On this path, should the subtraction happen in high or low precision, does it matter?

I don't know.
I've changed this to `dx′ = eltype(x).(x .- xT)`, which could help retain some more precision.

PR #160, opened 2023-09-12
Add all-keyword constructors, much like `@kwdef`

mcabbott:
Allows this:
```julia
julia> AdamW(lambda=333e-3)
OptimiserChain(Adam(0.001, (0.9, 0.999), 1.0e-8), WeightDecay(0.333))
```
Not entirely sure it's worth the complexity & extra documentation.

(Note that `adjust!` already exposes field names, so in a sense they are already API.)

### Checklist

- [x] Tests are added

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -8,12 +8,13 @@
 
 """
     Descent(η = 1f-1)
+    Descent(; eta)
 
 Classic gradient descent optimiser with learning rate `η`.
 For each parameter `p` and its gradient `dp`, this runs `p -= η*dp`.
 
 # Parameters
-- Learning rate (`η`): Amount by which gradients are discounted before updating
+- Learning rate (`η == eta`): Amount by which gradients are discounted before updating
                        the weights.
 """
 struct Descent{T} <: AbstractRule
@@ -37,13 +38,14 @@ end
 
 """
     Momentum(η = 0.01, ρ = 0.9)
+    Momentum(; [eta, rho])
 
 Gradient descent optimizer with learning rate `η` and momentum `ρ`.
 
 # Parameters
-- Learning rate (`η`): Amount by which gradients are discounted before updating
+- Learning rate (`η == eta`): Amount by which gradients are discounted before updating
                        the weights.
-- Momentum (`ρ`): Controls the acceleration of gradient descent in the
+- Momentum (`ρ == rho`): Controls the acceleration of gradient descent in the
                   prominent direction, in effect dampening oscillations.
 """
 @def struct Momentum <: AbstractRule
@@ -89,6 +91,7 @@ end
 
 """
     RMSProp(η = 0.001, ρ = 0.9, ϵ = 1e-8; centred = false)
+    RMSProp([eta, rho, epsilon, centred])
 
 Optimizer using the
 [RMSProp](https://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf)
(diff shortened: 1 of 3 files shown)


CarloLucibello:
I would really like to see this through. For instance, I hate having to copy and paste the default values of beta1 and beta2 from the docstring of AdamW and write  `AdamW(1e-3, (0.9, 0.999), 1e-4)` instead of `AdamW(eta=1e-3, decay=1e-4)`

mcabbott:
Sorry I missed your comment here. But `AdamW` precisely my motivating case too. Maybe we should do it?

ToucheSir:
No objections from me, though I would put in pin in un-greeking some of the common shared hyperparam names for the next major release.

src/rules.jl
```diff
@@ -99,11 +102,11 @@ generally don't need tuning.
...
+- Learning rate (`η == eta`): Amount by which gradients are discounted before updating
                        the weights.
-- Momentum (`ρ`): Controls the acceleration of gradient descent in the
+- Momentum (`ρ == rho`): Controls the acceleration of gradient descent in the
                   prominent direction, in effect dampening oscillations.
-- Machine epsilon (`ϵ`): Constant to prevent division by zero
+- Machine epsilon (`ϵ == epsilon`): Constant to prevent division by zero
                          (no need to change default)
 - Keyword `centred` (or `centered`): Indicates whether to use centred variant
                                      of the algorithm.
```
mcabbott, line 112: How should these be documented? Latest commit is one suggestion.

src/rules.jl
```diff
@@ -488,22 +492,27 @@ end
...
 # Parameters
-- Learning rate (`η`): Amount by which gradients are discounted before updating
+- Learning rate (`η == eta`): Amount by which gradients are discounted before updating
                        the weights.
-- Decay of momentums (`β::Tuple`): Exponential decay for the first (β1) and the
+- Decay of momentums (`β::Tuple == beta`): Exponential decay for the first (β1) and the
                                    second (β2) momentum estimate.
-- Weight decay (`γ`): Decay applied to weights during optimisation.
-- Machine epsilon (`ϵ`): Constant to prevent division by zero
+- Weight decay (`γ == gamma`): Decay applied to weights during optimisation.
```
mcabbott, line 506: Rea AdamW, is `gamma` the letter we want?

mcabbott (commented):
Maybe we should just do this. 

Docstrings only mention it for a few. Perhaps we can revisit later how best to document this.

src/rules.jl
ToucheSir: Maybe keep the ascii version in the part of the docstring that lists the actual parameter name, then use a latex-ified greek characters in the description of what the parameter does if applicable?

mcabbott: Not sure I follow. 

Goal of the current style is partly to be usable even if you don't know the names of greek letters. This was true before, as long as you can identify where the same scribble shows up.

src/rules.jl
```diff
@@ -115,10 +118,11 @@ struct RMSProp <: AbstractRule
   centred::Bool
 end
 
-function RMSProp(η = 0.001, ρ = 0.9, ϵ = 1e-8; centred::Bool = false, centered::Bool = false)
+function RMSProp(η, ρ = 0.9, ϵ = 1e-8; centred::Bool = false, centered::Bool = false)
```
CarloLucibello, line 121:
```suggestion
function RMSProp(η = 0.001, ρ = 0.9, ϵ = 1e-8; centred::Bool = false, centered::Bool = false)
```

src/rules.jl
```diff
@@ -488,22 +492,27 @@ end
...
+- Decay of momentums (`β::Tuple == beta`): Exponential decay for the first (β1) and the
                                    second (β2) momentum estimate.
-- Weight decay (`λ`): Controls the strength of ``L_2`` regularisation.
-- Machine epsilon (`ϵ`): Constant to prevent division by zero
+- Weight decay (`λ == lambda`): Controls the strength of ``L_2`` regularisation.
+- Machine epsilon (`ϵ == epsilon`): Constant to prevent division by zero
                          (no need to change default)
 """
-AdamW(η = 0.001, β = (0.9, 0.999), λ = 0, ϵ = 1e-8) =
+AdamW(η, β = (0.9, 0.999), λ = 0.0, ϵ = 1e-8) =
```
CarloLucibello, line 510:
```suggestion
AdamW(η=0.001, β = (0.9, 0.999), λ = 0.0, ϵ = 1e-8) =
```

src/rules.jl
mcabbott: The method `RMSProp(; eta = 0.001, rho = 0.9, epsilon = 1e-8, kw...)` just below provides the case `RMSProp()`.

CarloLucibello (approved):

src/rules.jl
ToucheSir: Something like `Machine epsilon (`epsilon`): Constant $$\eps$$ to prevent division by zero...`. I wanted to write it in as a suggestion but the comment wasn't on the right line for that.

PR #165, opened 2023-11-02
Add in-place `destructure!`

### LINKED ISSUE #146: Restructure makes a copy
In some situations, you have to restructure a lot if you use Flux, for instance if you want to run your batches as seperate solves in DiffEqFlux using an EnsembleProblem. You have to use something like a ComponentArray to pass the parameters through the solver and to let the adjoint methods do their work in differentiating the solve. But restructuring using a ComponentArray is unreasonably(?) slow in Flux. Switching to Lux eliminates those problems, but it seems like something that could be implemented better in Flux or ComponentArrays.

Example code:
```
using Flux, Profile, PProf, Random, ComponentArrays
layer_size = 256
layer = Flux.Dense(layer_size => layer_size)
params_, re = Flux.destructure(layer)
params = ComponentArray(params_)
function eval(steps, input)
	vec = input
	for i in 1:steps
		vec = re(params)(vec)
	end
end
Profile.clear()
@profile eval(100000, rand(Float32, layer_size))
pprof(; web=true)
```

Here, we spend 10% of the time in sgemv matrix multiplication, another 10% in the rest of the Dense call and about 75% in the Restructure. This gets worse if the networks are smaller. As far as I can read the flame graph, the restructure seems to spend a lot of time in the GC:

0d19ec3a732aaf409e85b06ebb4ebe6392198e20_2_1044x1000

Could there be a way to mitigate this specific problem? In particular if you use the same parameters. I think this would make some example code a lot faster too.

I also opened a discourse about this because I'm not sure if it's an issue with Flux specifically: https://discourse.julialang.org/t/flux-restructure-for-componentarrays-jl-unreasonably-slow/97849/3

mcabbott:
This adds a variant of `destructure` with minimal changes such that it writes back into the original model, instead of creating a copy. This may close #146, cc @glatteis

Marked draft as it seems surprisingly slow -- why?
```julia
julia> model = Chain(Dense(28^2 => 1024, relu), Dense(1024 => 10));

julia> params, re = destructure(model);  # old

julia> params, re! = destructure!(model);  # new

julia> @btime $re($params);  # This is the reconstruction cost
  min 229.334 μs, mean 374.473 μs (70 allocations, 3.11 MiB)

julia> @btime copy($params);  # ... and it's mostly allocation, same mean:
  min 219.417 μs, mean 367.168 μs (3 allocations, 3.11 MiB)

julia> @btime $re!($params);  # this avoids the allocations, but is quite slow.
  min 432.917 μs, mean 472.293 μs (58 allocations, 2.02 KiB)
```

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -13,7 +13,7 @@ include("utils.jl")
 include("adjust.jl")
 
 include("destructure.jl")
-export destructure
+export destructure, destructure!
 
 include("trainables.jl")
 export trainables
--- a/src/destructure.jl
+++ b/src/destructure.jl
@@ -9,6 +9,8 @@ Copies all [`trainable`](@ref Optimisers.trainable), [`isnumeric`](@ref Optimise
 to a vector, and returns also a function which reverses this transformation.
 Differentiable.
 
+See also [`destructure!`](@ref).
+
 # Example
 ```jldoctest
 julia> v, re = destructure((x=[1.0, 2.0], y=(sin, [3.0 + 4.0im])))
@@ -30,6 +32,36 @@ function destructure(x)
   flat, off, len = _flatten(x)
   flat, Restructure(x, off, len)
 end
+
+"""
+    destructure!(model) -> vector, reconstructor
+
+This is a variant of [`destructure`](@ref), whose reconstruction function mutates the model.
+Requires that all trainable parameters in the model be mutable arrays!
+
+# Example
+```jldoctest
+julia> m = (x=[1.0, 2.0], y=(sin, Float32[3.0 4.0], cos))
+
+julia> v, re! = destructure!(m)
+([1.0, 2.0, 3.0, 4.0], Restructure!(NamedTuple, ..., 4))
+
+julia> m === re!([3, 5, 7, 9])  # mutates the original m, and returns it
+true
+
+julia> m
+(x = [3.0, 5.0], y = (sin, Float32[7.0 9.0], cos))
+```
+"""
+function destructure!(x)
+  flat, off, len = _flatten(x)
+  flat, Restructure!(x, off, len)
+end
+
+# function destructure!(flat::AbstractVector, x)
+#   flat, off, len = _flatten!(flat, x)
+#   flat, Restructure!(x, off, len)
+# end
 
 """
     Restructure(Model, ..., length)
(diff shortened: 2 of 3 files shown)


glatteis:
Thanks you! Weird that it is so slow.

ToucheSir:
Could the in-place version be tripping some aliasing heuristic and hitting a slow path? I guess a profile would be illuminating.

kishore-nori:
Thank you initiating and implementing this idea, I think this is a great idea and would be very useful, I was trying this out because I am interested in-place copy of parameters into a model from a flat vector. From my comparisons, I suspect that one of the reason for the slowness of the in-place version is cache issues involving `copyto!` (simultaneously handling two big memories) which doesn't occur in the usual `restructure`, this can be observed when we have a relatively smaller model.

Additionally the following version (just utilising the fact that `copyto!` doesn't need reshape, as memory based) is faster in my tests,

```julia
function _rebuild_alt!(x, off, flat::AbstractVector, len = length(flat); walk = _Trainable_biwalk(), kw...)
  len == length(flat) || throw(DimensionMismatch("Rebuild expected a vector of length $len, got $(length(flat))"))
  fmap(x, off; exclude = isnumeric, walk, kw...) do y, o
    vecy = vec(y)
    copyto!(y, _getat_alt(vecy, o, flat, view))
  end
  x
end

_getat_alt(y::AbstractVector, o::Int, flat::AbstractVector, get=getindex) =
ProjectTo(y)(get(flat, o .+ (1:length(y))))
``` 
and its get better than the usual `restructure` when the model get smaller, which probably directs at cache issues for `copyto!`

```julia
using Flux, Optimisers, Zygote, BenchmarkTools

N = 1024
model = Chain(Dense(28^2 => N, relu), Dense(N => 10));

params,re = destructure(model)
params!,re! = destructure!(model)
params_alt!,re_alt! = destructure_alt!(model) # using above alternatives

@btime $re($params)
@btime $re!($params)
@btime $re_alt!($params)

  106.964 μs (44 allocations: 3.11 MiB)

  250.546 μs (35 allocations: 1.53 KiB)

  156.664 μs (39 allocations: 1.69 KiB)
```

kishore-nori:
Nice, that's good to know, and how do I make `@btime` output both `min` and `mean` like you have? (sorry for going off-topic).

And is the PR good to go?

mcabbott:
Mean is from https://github.com/JuliaCI/BenchmarkTools.jl/pull/258, which I should eventually re-write to `@btime / @btimes` or `@bmin / @btime` or something.

I see I did write some tests of this, it could all use one more look over. There's a commented out `destructure!(flat::AbstractVector, x)` which would allow the other direction not to allocate a whole copy too... not sure if that's almost ready or another whole project.

kishore-nori:
> Mean is from https://github.com/JuliaCI/BenchmarkTools.jl/pull/258

Interesting! That would be very useful! I ll see if I can take some pirate code out of it to start using locally :) 

I didn't take a look at `destructure!(flat::AbstractVector, x)` earlier, but looking at it now, I think it seems good, just that the length compatibility check needs to be done post all the `copyto!` to `flat` is done or in the last iteration. Can we do it before using `Flux.State` (earlier I was using `Flux.Params` for these purposes; but these parse over all the layers to make it happen, so there is a redundancy :/). Another thing is the type checking for `flat`, is it required here? Anyways, I think for most purposes `destructure!` is used just once (I can't think of cases other than where there is pruning involved), so even if  `destructure!(flat::AbstractVector, x)` is not ready, the rest should be useful in its own right, I think. But is there a reason on why it was commented out? I can try looking into it..

mcabbott:
I commented the method out just to focus on getting one thing working first. I believe it still needs tests, but otherwise this is nearly done.

Maybe I should check that my scary warning is true. I think something like this will return zero gradient:
```
v, re! = destructure!(model)
gradient(v) do w
  _ = re!(w)  # mutates model, in a way Zygote does not see
  sum(abs2, model(x))
end
```

kishore-nori:
That's great! 

Yes, you are right it returns `(nothing,)`, `_` breaks the connection in the chain of `rrule`s.

FerreolS:
This  `destructure!` seems promising for the small size problem (which is one of my use case).
Is there any reason why it is still a draft? Can I do something?

PR #166, opened 2024-02-02
Don't load Yota at all

mcabbott:


ToucheSir (approved):

mcabbott:
Paging @dfdx -- what's the status of Yota on latest Julia? Should we hold out for a fix here?

dfdx:
It can be fixed (see dfdx/Umlaut.jl/55 for the main required change), but I'm not sure Optimisers should depend on Yota at the moment. Yota is on hold until better times. Specifically, until more wide ecosystem for deep learning is available, so quite indefinite time. Till then, Optimisers may benefit more from a shorter list of more actively developed packages. 

Yet, if you believe Yota will bring some unique benefits, please ping me and I will gladly update the package.

mcabbott:
Certainly no pressure to work harder for the sake of this package! Just thought I shouldn't cut this without checking in.

We all had hopes that several AD packages might be viable at the same time, but remain stuck on Zygote...

PR #173, opened 2024-04-03
add trainables_with_path

CarloLucibello:
continuation of #171, to be merged after that.


CarloLucibello:
In https://github.com/FluxML/Optimisers.jl/pull/171#issuecomment-2037210879 @darsnack said

> Also, while okay for Functors, I'm not a fan of duplicated *_with_path functions. Like @ToucheSir suggested, if we want two variants, then a keyword flag seems like a better API.

If we are not concerned with type instability and performance issues I agree that a flag would be better.

CarloLucibello:
I don't understand how to switch this PR to target master, I'll open a new one

PR #174, opened 2024-04-05
add `path` option to `trainables`

CarloLucibello:
As discussed in #173 and #171


CarloLucibello:
ready to go

CarloLucibello:
I'd like a quick approval here since its blocking other work.

Issue #179, opened 2024-10-03
Optimiser state a not moving to GPU


vpuri3:
```julia
julia> using CUDA, Optimisers

julia> opt_st = Optimisers.setup(Optimisers.Adam(), rand(2))
Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), ([0.0, 0.0], [0.0, 0.0], (0.9, 0.999)))

julia> cu(opt_st).state[1]
2-element Vector{Float64}:
 0.0
 0.0

julia> cu(opt_st.state)[1]
2-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.0
 0.0
```

```juila
(NeuralROMs) pkg> st CUDA
Project NeuralROMs v0.0.1
Status `~/.julia/dev/NeuralROMs.jl/Project.toml`
  [052768ef] CUDA v5.5.2

(NeuralROMs) pkg> st Optimisers
Project NeuralROMs v0.0.1
Status `~/.julia/dev/NeuralROMs.jl/Project.toml`
  [3bd65402] Optimisers v0.3.3
```

vpuri3:
can be fixed with
```julia
julia> Adapt.@adapt_structure Optimisers.Leaf

julia> cu(opt_st)
Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), (Float32[0.0, 0.0], Float32[0.0, 0.0], (0.9, 0.999)))

julia> cu(opt_st).state
(Float32[0.0, 0.0], Float32[0.0, 0.0], (0.9, 0.999))

julia> cu(opt_st).state[1]
2-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.0
 0.0
```

CarloLucibello:
Since Leaf is a functor, it will move to GPU when using Flux.gpu or MLDataDevices.jl.

Also, if the input to setup is on GPU (which tipically is what you want), the state will be on GPU as well.

So maybe there is no need for Adapt?

vpuri3:
I routinely save the optimizer state to checkpoint files during training. So I need to move em back to the GPU when I restart training.

ToucheSir:
Yes, but that's already doable with `Flux.gpu(opt_st)` or `gdev = MLDataDevices.gpu_device(); gdev(opt_st)`. Is there a reason you're not able to use either of those libraries and _must_ use `CUDA.cu` directly?

vpuri3:
[...]

PR #180, opened 2024-10-03
Add `Adapt.adapt_structure` method for `Optimisers.Leaf`

vpuri3:
Fix https://github.com/FluxML/Optimisers.jl/issues/179

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
 version = "0.3.3"
 
 [deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,17 @@ authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
 version = "0.3.3"
 
 [deps]
-Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 Functors = "d9f16b24-f501-4c13-a1f2-28368ffc5196"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
+
+[weakdeps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+
+[extensions]
+OptimisersAdaptExt = ["Adapt"]
 
 [compat]
 Adapt = "4"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Optimisers"
 uuid = "3bd65402-5787-11e9-1adc-39752487f4e2"
+version = "0.4.2"
 authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
-version = "0.3.3"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
@@ -12,17 +12,20 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
 [weakdeps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
+EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 
 [extensions]
 OptimisersAdaptExt = ["Adapt"]
+OptimisersEnzymeCoreExt = "EnzymeCore"
 
 [compat]
 Adapt = "4"
 ChainRulesCore = "1"
-Functors = "0.4.9"
+EnzymeCore = "0.8.5"
+Functors = "0.4.9, 0.5"
 Statistics = "1"
 Zygote = "0.6.40"
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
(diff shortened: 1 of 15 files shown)


Project.toml
```diff
@@ -4,6 +4,7 @@ authors = ["Mike J Innes <mike.j.innes@gmail.com>"]
 version = "0.3.3"
 
 [deps]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
```
CarloLucibello, line 7: Compact bounds are needed

mcabbott:
I think this won't preserve identity? That's the big difference between Functors and Adapt

vpuri3:
@mcabbott can you explain? I don't know what identity is.


@CarloLucibello addressing your comment on the issue
> Since Leaf is a functor, it will move to GPU when using Flux.gpu or MLDataDevices.jl.

That is true, the MWE works with MLDataDevices. However, we still need Adapt functionality. Consider the case when Leaf is stored as part of a struct. Then using `MLDataDevice.gpu_device` doesn't move the state to the GPU even if we have Adapt.adapt_structure defined for the object.

```julia
using Optimisers, CUDA, LuxCUDA, MLDataDevices, Adapt

struct TrainState{Tp, To}
  p::Tp
  opt_st::To
end

Adapt.@adapt_structure TrainState

p = rand(2)
opt_st = Optimisers.setup(Optimisers.Adam(), p)
ts = TrainState(p, opt_st)
device = gpu_device()
device(ts).opt_st.state[1]
```

mcabbott:
Functors keeps an IdDict so that if the same array appears twice, this property is preserved by fmap. Optimisers.jl follows that too, and will use (and expect, IIRC) the same Leaf in such cases. So I don't see an easy way to male `cu` work for this.

vpuri3:
Would something like this solve the problem?

```julia
function Adapt.adapt_storage(to, leaf::Leaf)
    return fmap(x -> Adapt.adapt_storage(to, x), leaf)
end
```

ToucheSir:
It would not, because the IdDict needs to be shared between `Leaf`s. This is why `Flux.gpu` is a standalone function right now, FWIW.

vpuri3:
> It would not, because the IdDict needs to be shared between `Leaf`s. This is why `Flux.gpu` is a standalone function right now, FWIW.

Maybe it's possible to grab the IdDict and bind it to the new Leaf object? Where is it defined?

BTW the fix in this PR `Adapt.@adapt_storage Optimisers.Leaf` is working fine in my training runs.

ToucheSir:
> BTW the fix in this PR `Adapt.@adapt_storage Optimisers.Leaf` is working fine in my training runs.

That's because your model doesn't have any shared/"tied" parameters. e.g. `model.layer1.W === model.layer2.W`. Which is fine, but libraries like Optimisers have to support all use cases.

> Maybe it's possible to grab the IdDict and bind it to the new Leaf object? Where is it defined?

It's created at the top level in `Functors.fmap` and threaded down through the state tree. I'm not sure what it means to "grab the IdDict" in the context of overriding `adapt_structure`. `fmap` only ever calls `adapt`, and `adapt` doesn't take a cache argument.

vpuri3:
@ToucheSir, thanks for explaining. I added a warning to the adapt_structure pointing the user to Flux.gpu, and moved it all to an extension. Now `cu(opt_st)` won't silently not do the data transfer.

The behavior is as follows:
```julia
julia> using Optimisers, CUDA, LuxCUDA

julia> opt_st = Optimisers.setup(Optimisers.Adam(), zeros(2))
Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), ([0.0, 0.0], [0.0, 0.0], (0.9, 0.999)))

julia> cu(opt_st)
┌ Warning: `Optimisers.Leaf` object does not support device transfer via
│ `Adapt.jl`. Avoid this by calling `Flux.gpu/cpu` or
│ `MLDataDevices.cpu_device()/gpu_device()` on the optimiser state object.
│ See below GitHub issue for more details.
│ https://github.com/FluxML/Optimisers.jl/issues/179 
└ @ OptimisersAdaptExt ~/.julia/dev/Optimisers.jl/ext/OptimisersAdaptExt.jl:7
Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), (Float32[0.0, 0.0], Float32[0.0, 0.0], (0.9, 0.999)))

julia> cu(opt_st).state[1] |> typeof
CuArray{Float32, 1, CUDA.DeviceMemory}

julia> using MLDataDevices

julia> gpu_device()(opt_st)
Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), (Float32[0.0, 0.0], Float32[0.0, 0.0], (0.9, 0.999)))

julia> gpu_device()(opt_st).state[1]
2-element CuArray{Float32, 1, CUDA.DeviceMemory}:
 0.0
 0.0
```

ext/OptimisersAdaptExt.jl
```diff
@@ -0,0 +1,18 @@
+module OptimisersAdaptExt
+
+import Adapt
+import Optimisers: Leaf
+
+function Adapt.adapt_structure(to, leaf::Leaf)
+  @warn """`Optimisers.Leaf` object does not support device transfer via
+  `Adapt.jl`. Avoid this by calling `Flux.gpu/cpu` or
+  `MLDataDevices.cpu_device()/gpu_device()` on the optimiser state object.
+  """ maxlog=1
```
mcabbott, line 10: Do we want this only once? It's a potential correctness bug, not a performance issue. 
```suggestion
  """
```

vpuri3:
Thanks @ToucheSir , I resolved the merge conflicts.

ext/OptimisersAdaptExt.jl
ToucheSir: Is your worry that the message will get lost in the shuffle somehow? My thought was that people may have a valid use for this, and as long as they know what they're getting into the library doesn't have to remind them on every call.

Another practical concern would be what happens when someone tries to call `cu(large state tree)`. Not setting `maxlog` would mean other logging is drowned out because this warning would trigger for every `Leaf`.

ext/OptimisersAdaptExt.jl
```diff
@@ -0,0 +1,19 @@
...
+
+import Adapt
+import Optimisers: Leaf
+
+function Adapt.adapt_structure(to, leaf::Leaf)
+  @warn """`Optimisers.Leaf` object does not support device transfer via
+  `Adapt.jl`. This could lead to incorrect gradient updates. Avoid this by
+  calling `Flux.gpu/cpu` or `MLDataDevices.cpu_device()/gpu_device()` on the
+  optimiser state object.
+  """ maxlog=1
```
vpuri3, line 11: @mcabbott i edited the warning to say that this is a correctness issue.

ext/OptimisersAdaptExt.jl
mcabbott: I agree that printing 100 times on a big model is too much. Ideal would be once, on every invocation, IMO... but that's hard to make happen.

It's not the world's biggest correctness bug to ignore shared parameters, so maybe we should live with it. Maybe the message should say that's what the problem is?

PR #187, opened 2024-11-05
make docstrings consistent

CarloLucibello:


src/rules.jl
```diff
@@ -17,15 +17,14 @@ For each parameter `p` and its gradient `dp`, this runs `p -= η*dp`.
...
+@def struct Descent <: AbstractRule
+  eta = 0.1
 end
-Descent(; eta = 1f-1) = Descent(eta)
 
 init(o::Descent, x::AbstractArray) = nothing
 
 function apply!(o::Descent, state, x, dx)
-  η = convert(float(eltype(x)), o.eta)
+  η = ofeltype(x, o.eta)
```
mcabbott, line 27: When all the types got simplified, this ones was left allowing any T because someone had some use involving complex decay. Maybe I'd vote to leave it alone?

src/rules.jl
```diff
@@ -637,6 +650,7 @@ end
 
 """
     ClipNorm(ω = 10, p = 2; throw = true)
+    ClipNorm(; omega, p, throw)
```
mcabbott, line 653: I don't think there is such a method:
```suggestion
```

PR #188, opened 2024-11-05
Add the option couple to AdamW and set the default to match pytorch

CarloLucibello:
See https://github.com/FluxML/Flux.jl/issues/2433 for the details

Close https://github.com/FluxML/Flux.jl/issues/2433

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -518,7 +518,7 @@ Implemented as an [`OptimiserChain`](@ref) of [`Adam`](@ref) and [`WeightDecay`]
     The previous rule, which is closer to the original paper, can be obtained by setting `AdamW(..., couple=false)`.
     See [this issue](https://github.com/FluxML/Flux.jl/issues/2433) for more details.
 """
-struct AdamW <: AbstractRule
+struct AdamW{T1,T2,T3,T4} <: AbstractRule
   eta::T1
   beta::T2
   epsilon::T3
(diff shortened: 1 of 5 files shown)


src/rules.jl
```diff
@@ -506,12 +506,54 @@ Implemented as an [`OptimiserChain`](@ref) of [`Adam`](@ref) and [`WeightDecay`]
...
+"""
+struct AdamW{T1,T2,T3,T4} <: AbstractRule
+  eta::T1
+  beta::T2
+  epsilon::T3
+  lambda::T4
+  couple::Bool
+end
+
+function AdamW(η, β = (0.9, 0.999), λ = 0.0, ϵ = 1e-8; couple::Bool = true)
```
ToucheSir, line 529:
```suggestion
function AdamW(η, β = (0.9, 0.999), λ = 0.0, ϵ = 1e-8; couple::Union{Nothing,Bool} = nothing)
```
Could we do something like this and then add a warning in the constructor if `couple` isn't set?

ToucheSir (approved):

mcabbott:
I'm happy to change the default but I wonder if `couple` is too obscure a name... if you read someone's code could you guess what it means?

One extreme would be to have `AdamWpaper` and `AdamWtorch`, then neither gets the neutral name.

CarloLucibello:
I would leave this PR as it is. 99% of users will write `AdamW(lr)` and get what pytorch does and will be happy. We want `couple` to just not appear in most codebases and not create mental burden to most users. It will be available and clearly discoverable through the docstring for the few people aware of the paper/pytorch discussion.

PR #190, opened 2024-11-06
fix epsilon for Float16

### LINKED ISSUE #167: Adam optimizer can produce NaNs with Float16 due to small epsilon
Dear All,

I have found a bug in Adam's optimiser when used with Float16. 
An MWE woud look like this
```julia
using Optimisers

x = Float16[0.579, -0.729, 0.5493]
δx = Float16[-0.001497, 0.0001875, -0.013176]

os = Optimisers.setup(Optimisers.Adam(Float16(1e-4)), x);
os, x = Optimisers.update(os, x, δx)
```
where after update `x = Float16[Inf, -Inf, 0.5493]`. This happens with `Optimisers 0.3.2`. I believe the problem is caused on this line https://github.com/FluxML/Optimisers.jl/blob/1908a1cd599f656b15304a9722328bf9b2eed360/src/rules.jl#L221 because default `ϵ` is in Float64 and it is too small.
When I initialize Adam with different epsilon as `Optimisers.Adam(1e-4, (0.9, 0.999),  eps(Float16))`, it works well.
A possible solution would be to make sure that the epsilon has at least some value with respect to the given type of Float. For example  change `T(o.epsilon)` to `max(T(o.epsilon), eps(T))` on line 216.

I can prepare a pull-request with this test.

### LINKED ISSUE #189: mark as public any non-exported but documented interface

CarloLucibello:
Fix #167 adding `ϵ = max(ϵ, eps(T(0)))`. 

Also fix #189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ Manifest.toml
 .vscode/
 docs/build/
 .DS_Store
+/test.jl
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -4,6 +4,7 @@ using Functors: functor, fmap, fmap_with_path,
                 KeyPath, haskeypath, getkeypath,
                 isleaf, @functor, fmapstructure, children, AbstractWalk
 using LinearAlgebra
+using Compat: @compat
 
 include("interface.jl")
 export AbstractRule
@@ -24,6 +25,8 @@ export Descent, Adam, Momentum, Nesterov, Rprop, RMSProp,
        AdaGrad, AdaMax, AdaDelta, AMSGrad, NAdam, AdamW, RAdam, OAdam, AdaBelief,
        WeightDecay, SignDecay, ClipGrad, ClipNorm, OptimiserChain, Lion,
        AccumGrad
+
+@compat public apply!, init, setup, update, update!
 
 ###
 ### one-array functions
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -131,6 +131,7 @@ init(o::RMSProp, x::AbstractArray) = (zero(x), o.centred ? zero(x) : false)
 
 function apply!(o::RMSProp, state, x::AbstractArray{T}, dx) where T
   η, ρ, ϵ = T(o.eta), T(o.rho), T(o.epsilon)
+  ϵ = max(ϵ, eps(T(0)))
   quad, lin = state
 
   @.. quad = ρ * quad + (1 - ρ) * abs2(dx)
@@ -217,6 +218,7 @@ init(o::Adam, x::AbstractArray{T}) where T = (zero(x), zero(x), T.(o.beta))
 
 function apply!(o::Adam, state, x::AbstractArray{T}, dx) where T
   η, β, ϵ = T(o.eta), T.(o.beta), T(o.epsilon)
+  ϵ = max(ϵ, eps(T(0)))
   mt, vt, βt = state
 
   @.. mt = β[1] * mt + (1 - β[1]) * dx
@@ -280,6 +282,7 @@ init(o::RAdam, x::AbstractArray{T}) where T = (zero(x), zero(x), T.(o.beta), 1)
 
 function apply!(o::RAdam, state, x::AbstractArray{T}, dx) where T
   η, β, ϵ = T(o.eta), T.(o.beta), T(o.epsilon)
+  ϵ = max(ϵ, eps(T(0)))
   ρ∞ = 2/(1-β[2]) - 1 |> real
 
   mt, vt, βt, t = state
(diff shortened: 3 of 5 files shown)


src/rules.jl
```diff
@@ -131,6 +131,7 @@ init(o::RMSProp, x::AbstractArray) = (zero(x), o.centred ? zero(x) : false)
 
 function apply!(o::RMSProp, state, x::AbstractArray{T}, dx) where T
   η, ρ, ϵ = T(o.eta), T(o.rho), T(o.epsilon)
+  ϵ = max(ϵ, eps(T(0)))
```
mcabbott, line 134: I don't think this does the right thing:
```
julia> eps(Float32)  # roughly what we hard-code now
1.1920929f-7

julia> eps(0f0)  # what this PR uses
1.0f-45
```
I think a better pattern is this:
```suggestion
  η, ρ, ϵ = T(o.eta), T(o.rho), max(T(o.epsilon), eps(T))
```

CarloLucibello: I chose `eps(T(0))` because i didn't want to change the current behavior `eps=1f-8` for `Float32` but do the minimal change needed for Float16
```julia
julia> eps(Float16)
Float16(0.000977)

julia> eps(Float16(0))
Float16(6.0e-8)
```

src/Optimisers.jl
```diff
@@ -25,6 +26,8 @@ export Descent, Adam, Momentum, Nesterov, Rprop, RMSProp,
        WeightDecay, SignDecay, ClipGrad, ClipNorm, OptimiserChain, Lion,
        AccumGrad
 
+@compat public apply!, init, setup, update, update!
```
ToucheSir, line 29:
```suggestion
VERSION >= v"1.11.0-DEV.469" && eval(Meta.parse("public apply!, init, setup, update, update!"))
```
Per https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041/2, we can avoid a dependency.

.gitignore
```diff
@@ -2,3 +2,4 @@ Manifest.toml
 .vscode/
 docs/build/
 .DS_Store
+/test.jl
```
ToucheSir, line 5: Not sure where these came from, can they be reverted?

src/rules.jl
mcabbott: This doesn't seem big enough to avoid the issue in https://github.com/FluxML/Optimisers.jl/issues/167 : 
```julia
julia> using Optimisers

julia> x = Float16[0.579, -0.729, 0.5493]
3-element Vector{Float16}:
  0.579
 -0.729
  0.5493

julia> δx = Float16[-0.001497, 0.0001875, -0.013176]
3-element Vector{Float16}:
 -0.001497
  0.0001875
 -0.013176

julia> os = Optimisers.setup(Optimisers.Adam(Float16(1e-4)), x);

julia> os, x = Optimisers.update(os, x, δx)
(Leaf(Adam(0.000100017, (0.9, 0.999), 1.0e-8), (Float16[-0.0001498, 1.88e-5, -0.001319], Float16[0.0, 0.0, 2.0e-7], (0.81, 0.998))), Float16[Inf, -Inf, 0.5493])

julia> os = Optimisers.setup(Optimisers.Adam(eta=1e-4, epsilon=eps(Float16(0))), x);  # this PR

julia> os, x = Optimisers.update(os, x, δx)
(Leaf(Adam(0.0001, (0.9, 0.999), 5.96046e-8), (Float16[-0.0001498, 1.88e-5, -0.001319], Float16[0.0, 0.0, 2.0e-7], (0.81, 0.998))), Float16[Inf, -Inf, 0.5522])

julia> os = Optimisers.setup(Optimisers.Adam(eta=1e-4, epsilon=eps(Float16)), x);  # as suggested in #167

julia> os, x = Optimisers.update(os, x, δx)
(Leaf(Adam(0.0001, (0.9, 0.999), 0.000976562), (Float16[-0.0001498, 1.88e-5, -0.001319], Float16[0.0, 0.0, 2.0e-7], (0.81, 0.998))), Float16[Inf, -Inf, 0.5522])
```
... in fact neither are big enough. 

So, what problem do we want this to solve?

mcabbott: Ok. How broadly is 6e-8 big enough for this purpose? It seems a little random... maybe I'd be happier to hard-code an obviously chosen number, which is what `1e-8` is... like `_eps(::Type{Float16}, e) = max(1e-7, e)`?

.gitignore
CarloLucibello: just for my convenience. I'd like to keep it, it doesn't hurt anyone

src/rules.jl
CarloLucibello: done

PR #192, opened 2024-11-08
Add `Duplicated` methods

mcabbott:
This moves part of https://github.com/FluxML/Flux.jl/pull/2471 to an extension here. 

Bumps Julia support to 1.10.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -536,25 +536,25 @@ end
     end
 
     @testset "Enzyme Duplicated" begin
-        x_dx = Duplicated(Float16[1,2,3], Float16[1,0,-4])
-        st = Optimisers.setup(Momentum(1/9), x_dx)  # acts only on x not on dx
-        @test st isa Optimisers.Leaf
-        @test nothing === Optimisers.update!(st, x_dx)  # mutates both arguments
-        @test x_dx.val ≈ Float16[0.8887, 2.0, 3.445]
+      x_dx = Duplicated(Float16[1,2,3], Float16[1,0,-4])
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -534,6 +534,28 @@ end
...
+      model = (x=shared, y=shared)
+      grad = deepcopy(model) # Enzyme produces something like this, grad.x === grad.y, already accumulated.
+      dup = Duplicated(model, model)
+      st2 = Optimisers.setup(Descent(0.1), model)
+      Optimisers.update!(st2, dup)
+      @test model.x ≈ [0.9]
+      shared .= 1
+      Optimisers.update!(st2, model, grad)
+      model.x ≈ [0.8]  # This is wrong, but don't make it a test.
+      # Ideally, perhaps the 3-arg update! could notice that grad.x===grad.y, and not accumulate the gradient in this case?
```
ToucheSir, line 555: I get what you mean, but do we promise to handle shared gradient arrays? Might be easier to add a warning in the docs. Maybe a broken test outside of the Enzyme test suite if we really care about this.

ToucheSir (approved):

test/runtests.jl
mcabbott: At present (after this PR) I think the usual path with Zygote and this new path with Enzyme should always agree.

Whether the normal path can auto-detect this (as in the comment) I'm not really sure. It's possible to make Zygote return two components which are `===`, and it's conceivable that this could happen with shared parameters. 

Nevertheless, detecting this would probably be better than not. In real use, e.g. you use the same layer twice, Zygote is certainly going to return two new arrays for shared parameters.

PR #196, opened 2024-12-10
Add Apollo optimizer (https://arxiv.org/pdf/2412.05270) - WIP

murrellb:
This adds a draft of the low-rank Apollo optimizer that was preprinted yesterday (https://arxiv.org/pdf/2412.05270). Looks like it has some very nice properties, especially with the low memory footprint.

This works on the one case I've tested, but should be considered a WIP/draft until the codebase from the preprint is available, especially since there was a little guesswork from the manuscript alone. But I figured I'd open a PR in case anyone else wants to play with it, and if there is interest in merging I'll spend a bit more effort to check that it matches the "official" implementation when that arrives.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -599,6 +599,54 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
   return (mt, st, βt .* β), dx′
 end
 
+
+"""
+    Apollo(opt, r, u)
+
+Apollo optimizer from Zhu et al. (https://arxiv.org/pdf/2412.05270). Tracks moments in a low-rank subspace, aiming for Adam-like behavior with minimal additional memory usage.
+`opt` is an AdamW optimizer, `r` is the random projection rank (smaller for lower memory use), and `u` is the random projection update interval.
+"""
+struct Apollo{T1} <: AbstractRule
+    opt::T1
+    r::Int #Subspace rank
+    u::Int #Subspace update frequency (T in paper)
+end
+
+#Use the base init and apply for 1D arrays
+init(o::Apollo, x::AbstractArray{T,1}) where T = init(o.opt, x)
+apply!(o::Apollo, state, x::AbstractArray{T,1}, dx) where T = apply!(o.opt, state, x, dx)
+
+function init(o::Apollo, x::AbstractArray{T,2}) where T
+    rank = min(o.r, ceil(Int, size(x,2) / 2))
+    P = randn(T, rank, size(x,1)) .* T(1/rank)
+    ((similar(x, rank, size(x,2)) .= 0, similar(x, rank, size(x,2)) .= 0, o.opt.beta), 0, P)
+end
+
+function apply!(o::Apollo, state, x::AbstractArray{T,2}, dx) where T
+    (mt, vt, βt), t, P = state
+    η = T(o.opt.eta)
+    λ = T(o.opt.lambda)
+    β = T.(o.opt.beta)
+    ϵ = T(o.opt.epsilon)
+    if mod(t, o.u) == 100 
+        rank = min(o.r, ceil(Int, size(x,2) / 2))
+        @show rank, typeof(rank)
+        P = randn(T, rank, size(x,1)) .* T(1/rank)
+    end
+    R = P * dx
+    Optimisers.@.. mt = β[1] * mt + (1 - β[1]) * R
+    Optimisers.@.. vt = β[2] * vt + (1 - β[2]) * abs2(R)
+    Rhat = @. mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)
+    s = [Optimisers.norm(view(Rhat, :, j)) / Optimisers.norm(view(R, :, j)) for j in 1:size(R,2)]
+    S = Diagonal(s)
+    dx′′ = η * dx * S + λ * x
+    return ((mt, vt, βt .* β), t+1, P), dx′′
+end
+
+
+
+
+
 """
     WeightDecay(λ = 5e-4)
     WeightDecay(; [lambda])


src/rules.jl
```diff
@@ -599,6 +599,54 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+    Apollo(opt, r, u)
+
+Apollo optimizer from Zhu et al. (https://arxiv.org/pdf/2412.05270). Tracks moments in a low-rank subspace, aiming for Adam-like behavior with minimal additional memory usage.
+`opt` is an AdamW optimizer, `r` is the random projection rank (smaller for lower memory use), and `u` is the random projection update interval.
+"""
+struct Apollo{T1} <: AbstractRule
+    opt::T1
+    r::Int #Subspace rank
+    u::Int #Subspace update frequency (T in paper)
+end
```
mcabbott, line 613: If this must be AdamW, then you could do this, which supplies defaults:
```suggestion
@def struct Apollo{T1} <: AbstractRule
    opt = AdamW()
    r = 10 # Subspace rank
    u = 10 # Subspace update frequency (T in paper)
end
```

src/rules.jl
```diff
@@ -599,6 +599,54 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+
+#Use the base init and apply for 1D arrays
+init(o::Apollo, x::AbstractArray{T,1}) where T = init(o.opt, x)
+apply!(o::Apollo, state, x::AbstractArray{T,1}, dx) where T = apply!(o.opt, state, x, dx)
+
+function init(o::Apollo, x::AbstractArray{T,2}) where T
```
mcabbott, line 619: For arrays of >2D (e.g. weight of Conv), should there be methods to reshape to matrix & reshape back?

mcabbott (commented):
Sounds good.

If you edit this line  https://github.com/FluxML/Optimisers.jl/blob/master/test/rules.jl#L11  then it CI will check that it does converge on some sample problems.

src/rules.jl
murrellb: Yes, that is on my list. There is also this pesky assertion about the dimension ordering that means that some matrices will have to be transposed:

image

but I'm not sure a lazy transpose as W comes in and goes out will be optimal - might have to write a new path for those. I suspect we (as in "humanity") don't know if this actually helps, so I'm go make sure this can be controlled by a flag.

murrellb: Both of these are included now.

src/rules.jl
murrellb: I wanted the user to be able to pass in either an Int or a function for rank (the latter where they can scale the rank based on the dim), so I've written some custom constructors with defaults instead of this approach.

murrellb:
I've now added an attempt at a "gradient norm growth limiter" because this paper used this in conjunction with Apollo. I think they apply this over the whole model, but here it will apply per tensor. This should still do the trick though, and it seems like it might be more generally useful for controlling loss spikes.

Apollo seems to error on some "gradient type" tests, where something is passed in that I didn't expect and don't understand. I think it doesn't even like calling `size` on these inputs? Haven't looked too closely, but maybe this is a familiar issue to you folks?

Other than that, I actually think that maybe we don't need to wait for the ref implementation to merge? Would Optimisers.jl consider an "experimental" tier of optmizers (eg. "ExperimentalApollo"), where they might not match the reference implementation and where their behavior may change in the future? There are many optimizers coming out, and this might be a decent strategy to stay cutting-edge for ones that are not yet battle-tested?

murrellb:
This is now working on GPU. In combination with https://github.com/FluxML/Zygote.jl/pull/1541 I can now train a 7 billion parameter transformer on a single 48gb A6000 GPU:

image

src/rules.jl
```diff
@@ -599,6 +599,136 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
   return (mt, st, βt .* β), dx′
 end
 
+
+"""
+    GradNormGrowthLimiter(γ = 1.1; m = 1e-3, ϵ = 1e-8, throw = true, paramscale_min = true)
+
+Gradient norm growth limiter from Chen et al. (https://arxiv.org/pdf/2410.01623) and used with Apollo in Zhu et al. (https://arxiv.org/pdf/2412.05270).
```
CarloLucibello, line 606:
```suggestion
Gradient norm growth limiter from [Chen et al.](https://arxiv.org/abs/2410.01623) and used with Apollo in [Zhu et al.](https://arxiv.org/abs/2412.05270).
```

src/rules.jl
```diff
@@ -599,6 +599,136 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+With Optimisers.jl this will apply per-tensor, which may not be the same as the implementations in these papers. It still seems to help, but the ideal settings may vary.
+This also introduces `m` a hard minimum on the gradient norm, and never rescales grads below this, preventing a tensor from getting "trapped" near zero.
+This can be a fixed min, or scaled by the number of parameters in the tensor (with `paramscale_min = true`).
```
CarloLucibello, line 609: explain the role of gamma?

src/rules.jl
```diff
@@ -599,6 +599,136 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
   return (mt, st, βt .* β), dx′
 end
 
+
+"""
+    GradNormGrowthLimiter(γ = 1.1; m = 1e-3, ϵ = 1e-8, throw = true, paramscale_min = true)
```
CarloLucibello, line 604: should the default value for `m` correspond to the original paper (i.e. m=0 i suppose)?

murrellb:
Thanks @CarloLucibello - I just woke up and noticed some issues with this PR, so I was about to convert it to draft! I need to sort out `adjust` for Apollo. I'll work through your GradNormGrowthLimiter comments too.

src/rules.jl
murrellb: `m=0` makes sense when this is applied to the entire model, but could be fatal when applied tensor-wise. I think it is better to have non-footgun defaults, and make it clearer that this isn't a faithful reproduction?

src/rules.jl
murrellb: Done in the new version.

src/rules.jl
murrellb: I've kept a non-zero default, but I've tweaked the docs to clarify that this method isn't quite the same as in those papers. (I also switched the "scaling m by the number of parameters" to using `sqrt`).

zhuhanqing:
Hi all, I am one of the main authors of the APOLLO paper, Hanqing Zhu. 
Thanks so much for bringing our APOLLO to FluxML. I am so sorry our code releases are still blocked by the internal review process. If you have any questions or want to know implementation details, please let us know, and we are more than happy to assist you with this integration and implementation process!

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+
+Gradient norm growth limiter. Inspired by [Chen et al.](https://arxiv.org/abs/2410.01623) and used with Apollo in [Zhu et al.](https://arxiv.org/abs/2412.05270), but
+with Optimisers.jl this will apply per-tensor instead of per-model, and as a result the defaults are different. `γ` controls the maximum that the gradient norm can grow
+from one step to the next. This implementation also introduces `m` a hard minimum on the gradient norm threshold, and never rescales grads below this, preventing a tensor
+from getting "trapped" near zero. This can be a fixed min, or scaled by the square root of the number of parameters in the tensor (with `paramscale_min = true`).
```
mcabbott, line 609: Can this explain what it does do, mathematically, before explaining that it's different to some paper?

> `γ` controls the maximum that the gradient norm can grow from one step to the next.

I don't know what this means without reading the code. Can you write like `if norm(dx, 2) > γ * norm(dx_prev, 2)` to explain the condition, and explain exactly what happens if this is violated?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+    opt::T1
+    eta::T2
+    r::T3 #Maps non-first dims to rank
+    u::T4 #Subspace update frequency (T in paper)
+    sort_dims::T5 #Whether to swap the dims of x and dx when the second dim is smaller than the first
+end
+
+
+Apollo() = Apollo(AdamW(0.001), 0.001, dim -> ceil(Int, sqrt(dim)), 100, true)
+Apollo(η::Real, rank::Int; u = 100, sort_dims = true) = Apollo(AdamW(η), η, dim -> max(dim, rank), u, sort_dims)
```
mcabbott, line 672: Are you sure you want `max`?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+    Apollo(opt::AdamW, rank::Int; u = 100, sort_dims = false)
+    Apollo(opt::AdamW; rank_function::Function = dim -> ceil(Int, sqrt(dim)), u = 100, sort_dims = false)
+
+Apollo optimizer from Zhu et al. (https://arxiv.org/pdf/2412.05270). Tracks moments in a low-rank subspace, aiming for Adam-like behavior with minimal additional memory usage.
+First argument can be an AdamW optimizer, or a learning rate (which will use the default AdamW optimizer with that learning rate). Second argument can be a rank, or a function
+to compute the rank from the second dimension (or the product of all dims > 1) of the weight matrix (or tensor).
+"""
+struct Apollo{T1, T2, T3, T4, T5} <: AbstractRule
+    opt::T1
+    eta::T2
```
mcabbott, line 664: Why store opt and eta?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+    r::T3 #Maps non-first dims to rank
+    u::T4 #Subspace update frequency (T in paper)
+    sort_dims::T5 #Whether to swap the dims of x and dx when the second dim is smaller than the first
```
mcabbott, line 667: These have fixed types, right?
```suggestion
    u::Int # Subspace update frequency (T in paper)
    sort_dims::Bool # Whether to swap the dims of x and dx when the second dim is smaller than the first
```

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+end
+
+
+function apply!(o::Apollo, state, x::AbstractArray{T}, dx) where T
+  swapped = false
+  original_size = size(x)
+  x = reshape(x, size(x,1), nonfirstdims(x))
+  
+  dx = Broadcast.materialize(dx) #This is to stop the "gradient type" @lazy test from failing due to reshape.
+  dx = reshape(dx, size(x,1), nonfirstdims(x))
```
mcabbott, line 700: Do you need to materialize in matrix case?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+      P .*= T(sqrt(1/rank))
+  end
+  R = P * dx
+  @.. mt = β[1] * mt + (1 - β[1]) * R
+  @.. vt = β[2] * vt + (1 - β[2]) * abs2(R)
+  Rhat = @. mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)
+  s = sqrt.(sum(abs2.(Rhat), dims=1))[:] ./ (sqrt.(sum(abs2.(R), dims=1))[:] .+ ϵ)
+  dx′′ = η * (dx .* reshape(s, 1, :)) + λ * x
+  if swapped
+      dx′′ = dx′′'
```
mcabbott, line 727:
```suggestion
      dx′′ = transpose(dx′′)
```

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+      rank = o.r(second_dim)
+      randn!(P)
+      P .*= T(sqrt(1/rank))
+  end
+  R = P * dx
+  @.. mt = β[1] * mt + (1 - β[1]) * R
+  @.. vt = β[2] * vt + (1 - β[2]) * abs2(R)
+  Rhat = @. mt / (1 - βt[1]) / (sqrt(vt / (1 - βt[2])) + ϵ)
+  s = sqrt.(sum(abs2.(Rhat), dims=1))[:] ./ (sqrt.(sum(abs2.(R), dims=1))[:] .+ ϵ)
+  dx′′ = η * (dx .* reshape(s, 1, :)) + λ * x
```
mcabbott, line 725: These lines allocate a lot. 

Rhat isn't used?

For the rest maybe it can be something like
```
sum1R2 = sum(abs2, R; dims=1)  # it's already the right shape, no need for [:] & reshape(s, 1, :)?
s = @. sqrt(sum1R2) / sqrt(Rhat + ϵ)
dx′′ = @lazy η * (dx * s) + λ * x  # one fused broadcast
```

mcabbott, line 725: These lines allocate a lot. 

Rhat isn't used?

For the rest maybe it can be something like
```
sum1R2 = sum(abs2, R; dims=1)  # it's already the right shape, no need for [:] & reshape(s, 1, :)?
s = @. sqrt(sum1R2) / sqrt(Rhat + ϵ)
dx′′ = @lazy η * (dx * s) + λ * x  # one fused broadcast
```

src/rules.jl
mcabbott: Also, this sort of branching introduces type instability. IDK if we care but perhaps worth some thought. Maybe there's a nicer way to just store everything transposed?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+struct Apollo{T1, T2, T3, T4, T5} <: AbstractRule
+    opt::T1
+    eta::T2
+    r::T3 #Maps non-first dims to rank
+    u::T4 #Subspace update frequency (T in paper)
+    sort_dims::T5 #Whether to swap the dims of x and dx when the second dim is smaller than the first
+end
+
+
+Apollo() = Apollo(AdamW(0.001), 0.001, dim -> ceil(Int, sqrt(dim)), 100, true)
```
mcabbott, line 671: Can't this method just be created by giving a default to eta in the next one?

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+"""
+struct GradNormGrowthLimiter <: AbstractRule
+    γ::Float64
+    m::Float64 #Min grad norm, to stop a tensor getting stuck near zero
+    ϵ::Float64
+    throw::Bool
+    paramscale_min::Bool
+end
+
+GradNormGrowthLimiter(γ = 1.1; m = 1e-3, ϵ = 1e-8, throw = true, paramscale_min = true) = GradNormGrowthLimiter(γ, m, ϵ, throw, paramscale_min)
```
mcabbott, line 619: We don't have greek-letter keyword options, nor field names -- the API should never ask the user to type these. They are used only in documentation / as local variables. Probably the first 3 should be positional.

Bikeshedding names bit, to avoid overly long things, the constructor could be:
```suggestion
NormGrowLimit(γ = 1.1, m = 1e-3, ε = 1e-8; throw = true, scale = true) = NormGrowLimit(γ, m, ε, throw, scale)
```

src/rules.jl
```diff
@@ -599,6 +599,140 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+
+Gradient norm growth limiter. Inspired by [Chen et al.](https://arxiv.org/abs/2410.01623) and used with Apollo in [Zhu et al.](https://arxiv.org/abs/2412.05270), but
+with Optimisers.jl this will apply per-tensor instead of per-model, and as a result the defaults are different. `γ` controls the maximum that the gradient norm can grow
+from one step to the next. This implementation also introduces `m` a hard minimum on the gradient norm threshold, and never rescales grads below this, preventing a tensor
+from getting "trapped" near zero. This can be a fixed min, or scaled by the square root of the number of parameters in the tensor (with `paramscale_min = true`).
+"""
+struct GradNormGrowthLimiter <: AbstractRule
+    γ::Float64
+    m::Float64 #Min grad norm, to stop a tensor getting stuck near zero
+    ϵ::Float64
```
mcabbott, line 614: We don't allow unicode field names, suggest:
```suggestion
    gamma::Float64
    mu::Float64 # Min grad norm, to stop a tensor getting stuck near zero
    epsilon::Float64
```

src/rules.jl
murrellb: Thanks for catching this.

src/rules.jl
murrellb: I originally just stored `opt`, but then getting `adjust` working seemed tricky (likely a skill issue on my part though). Options were to include all the other AdamW params directly in this struct, or have an AdamW that only applies to the low-rank moments (which doesn't use eta, so its eta is redundant), and a separate eta that gets tweaked by adjust. The latter seemed better because then you can just wrap an existing AdamW in this.

src/rules.jl
murrellb: Yup.

src/rules.jl
murrellb: For everything except the whatever comes in during the "gradient type" test you don't need materialize. I wasn't 100% sure exactly what is coming in during those tests, so wasn't sure how to separate them from regular matrix/tensors. What do you suggest here?

src/rules.jl
murrellb: Maybe an optimization we can figure out later if it becomes an issue?

src/rules.jl
mcabbott: Storing an AdamW seems fine, surely we can make `adjust` just work through onto the inner struct.

src/rules.jl
murrellb: Done, but changed the variable names to avoid eg. gamma.

src/rules.jl
murrellb: I went with NormGrowthCap here.

src/rules.jl
murrellb: This is fixed via a different route.

src/rules.jl
murrellb: I got something like this working, but the @lazy breaks things, so omitted for now.

src/rules.jl
murrellb: I've made adjust work on the inner Adam now, so have dropped the additional eta.

src/rules.jl
murrellb: Done.

murrellb:
FYI I've made this: https://github.com/MurrellGroup/CannotWaitForTheseOptimisers.jl which can house new optimisers until you folks are happy to have them included in Optimisers.jl (mostly because I really just need to be able to use them conveniently for now).

PR #198, opened 2024-12-11
fix AdamW

### LINKED ISSUE #197: AdamW: epsilon and lambda swapped?
image

Weird, given that the docs imply epsilon is non-zero:

image

But looking at the code, the field order in the struct vs the constructors seems to swap them?

image

CarloLucibello:
fix #197

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -531,14 +531,14 @@ Implemented as an [`OptimiserChain`](@ref) of [`Adam`](@ref) and [`WeightDecay`]
 struct AdamW{T1,T2,T3,T4} <: AbstractRule
   eta::T1
   beta::T2
-  epsilon::T3
   lambda::T4
+  epsilon::T3
   couple::Bool
 end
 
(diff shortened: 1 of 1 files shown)


ToucheSir (approved):

src/rules.jl
```diff
@@ -538,7 +538,7 @@ end
 
 function AdamW(η, β = (0.9, 0.999), λ = 0.0, ϵ = 1e-8; couple::Bool = true)
```
mcabbott, line 539: One method puts epsilon 3rd, and one puts it 4th. Shouldn't these orders always agree?

mcabbott: (That was supposed to highlight the struct definition, too.)

Also, why does this allow any types, not Float64 like other rules?

CarloLucibello: I would allow for at least `AdamW(eta; couple=true)` so 
> Or we could elaborate the macro to always allow n positional and N-n keywords for the remaining arguments.

seems a viable option that preserves all current construction modalities, right?

mcabbott: Yes.

For now at least fix the types?

CarloLucibello: Ok

PR #199, opened 2024-12-12
restrict AdamW's fields to Float64 

CarloLucibello:
As per comment https://github.com/FluxML/Optimisers.jl/pull/198#discussion_r1880405919,
since we do it for any other rule (explicitly or through `@def`) except for `Descent`.

Is there any reason why `Descent` is left out @mcabbott ?


mcabbott:
> reason why Descent is left out

Yes, someone complained that they had a use case for Descent with complex eta. Maybe discussed in the original PR.

mcabbott (approved):

Issue #201, opened 2024-12-15
Weight decay (L2) and sign decay (L1) at the same time?


murrellb:
### Motivation and description

In [other contexts](https://en.wikipedia.org/wiki/Elastic_net_regularization), combining L1 and L2 regularization can be reasonable. In Optimisers, they have the same parameter name, which, if I understand correctly, will mean that `adjust` will change both?

murrellb:
Similarly, what if I wanted to use SignDecay with AdamW, so I set AdamW's lambda to 0. Would trying to adjust the SignDecay lambda cause AdamW's lambda to then be non-zero?

mcabbott:
Yes it will change both:
```julia
julia> os = Optimisers.setup(OptimiserChain(SignDecay(0.1), AdamW(lambda=0.1)), (; x=[12 34.]))
(x = Leaf(OptimiserChain(SignDecay(0.1), AdamW(0.001, (0.9, 0.999), 0.1, 1.0e-8, true)), (nothing, ([0.0 0.0], [0.0 0.0], (0.9, 0.999)))),)

julia> Optimisers.adjust!(os, lambda=0.3)

julia> os
(x = Leaf(OptimiserChain(SignDecay(0.3), AdamW(0.001, (0.9, 0.999), 0.3, 1.0e-8, true)), (nothing, ([0.0 0.0], [0.0 0.0], (0.9, 0.999)))),)
```
That's a reason to give them all different names, e.g. `lambda1` for SignDecay and `lambda2` for WeightDecay.

But it's a breaking change, as people may be using `adjust` already. Unfortunately can't now be combined with the change for AdamW.

murrellb:
How about adding an adjust method that lets you specify the type. Like `adjust!(state, (SignDecay, lambda = 0.05))` and only adjusts optimisers that match the type?

mcabbott:
All things are possible but I don't think we should add further complexity to `adjust!`'s interface. 

The less-breaking way is to change all 3, and then overload the implementation of `adjust!` for just these types to still accept lambda. Although if anyone has saved a state tree, then that will still be broken.

mcabbott:
Note that because of an earlier re-naming (maybe from what Flux called things, maybe to match the AdamW convention, when AdamW made a chain with WeightDecay) you can in fact change L1 and L2 parameters independently:
```julia
[...]

PR #202, opened 2024-12-19
Controlling which optimizer rule you're adjusting.

murrellb:
This PR tries to address the issue in https://github.com/FluxML/Optimisers.jl/issues/201 where, if two Rules have the same parameter name, you can't use the `adjust` interface to control them independently.

This is intended to be non-breaking, so the previous behavior is undisturbed:
image
but you can now additionally specify the type you wish to change:
image

Implementation-wise, this added quite a few `adjust` and `adjust!` methods. I'm not 100% sure all of them are needed, so maybe we can prune a few back but I wasn't quite sure of the edge cases I'm not aware of, so I tried to cover everything.

I this PR's version, analogous to above, if you just pass in a float you'll change eta, but only in the optimizer you intend:
image
This is maybe less critical to have, so we could cut back all of the non-keyword methods that specify a type.

Doing all this introduces a bit of core package developer complexity, but it is all generic so you shouldn't need to touch it when adding a new optimizer, and the usage seems pretty intuitive from a user perspective, so I don't think it adds interface complexity. To be honest, the current behavior seems like a footgun, and as a user with an OptimizerChain, I'd only ever use `adjust`/`adjust!` while specifying the type, because otherwise I'd have to check the parameter list of all the OptimizerChain components just to be sure I'm not doing something I don't intend.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -144,3 +144,33 @@ function _adjust(r::T, nt::NamedTuple) where T <: AbstractRule
   end
   T(vals...)  # relies on having the default constructor
 end
+
+###
+#adjust with type control
+###
+
+adjust!(ℓ::Leaf, oT::Type, eta::Real) = (ℓ.rule = adjust(ℓ.rule, oT, eta); nothing)
+adjust!(ℓ::Leaf, oT::Type; kw...) = (ℓ.rule = adjust(ℓ.rule, oT; kw...); nothing)
+
+adjust(ℓ::Leaf, oT::Type, eta::Real) = Leaf(adjust(ℓ.rule, oT, eta), ℓ.state, ℓ.frozen)
+adjust(ℓ::Leaf, oT::Type; kw...) = Leaf(adjust(ℓ.rule, oT; kw...), ℓ.state, ℓ.frozen)
+
+adjust!(tree, oT::Type, eta::Real) = foreach(st -> adjust!(st, oT, eta), tree)
+adjust!(tree, oT::Type; kw...) = foreach(st -> adjust!(st, oT; kw...), tree)
+
+adjust(r::AbstractRule, oT::Type, eta::Real) = ifelse(isa(r, oT), adjust(r, eta), r)
+adjust(r::AbstractRule, oT::Type; kw...) = ifelse(isa(r, oT), adjust(r; kw...), r)
+
+adjust!(r::AbstractRule, oT::Type, eta::Real) = ifelse(isa(r, oT), adjust!(r, eta), r)
+adjust!(r::AbstractRule, oT::Type; kw...) = ifelse(isa(r, oT), adjust!(r; kw...), r)
+
+function adjust(tree, oT::Type, eta::Real)
+  t′ = fmap(copy, tree; exclude = maywrite)
+  adjust!(t′, oT, eta)
+  t′
+end
+function adjust(tree, oT::Type; kw...)
+  t′ = fmap(copy, tree; exclude = maywrite)
+  adjust!(t′, oT; kw...)
+  t′
+end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -799,6 +799,9 @@ end
 adjust(ℓ::OptimiserChain, eta::Real) = OptimiserChain(map(opt -> adjust(opt, eta), ℓ.opts)...)
 adjust(ℓ::OptimiserChain; kw...) = OptimiserChain(map(opt -> adjust(opt; kw...), ℓ.opts)...)
 
+adjust(ℓ::OptimiserChain, oT::Type, eta::Real) = OptimiserChain(map(opt -> adjust(opt, oT, eta), ℓ.opts)...)
+adjust(ℓ::OptimiserChain, oT::Type; kw...) = OptimiserChain(map(opt -> adjust(opt, oT; kw...), ℓ.opts)...)
+
 
 """
     AccumGrad(n::Int)


mcabbott:
I can't easily copy-paste from screenshots...

1. The spelling proposed here is `Optimisers.adjust!(os, SignDecay, lambda=0.3)`, passing just the type.

Alternatives are:

2. `Optimisers.adjust!(os, SignDecay(0.3))`, passing a whole new rule.

3. Changing e.g. SignDecay  to use `lambda1` and WeightDecay to use `lambda2` to avoid the clash.

4. Something more exotic, like `adjust!(fun::Function, os, lambda=0.3)` where this `fun` must return a Bool and gets... the rule? 

Are there other options?

I suppose 3 doesn't help if you wish to adjust some learning rates and not others, we aren't going to rename them like `eta37`.

The advantage of 2 is you need never know the field names. The disadvantage is that `adjust!(os, AdamW(lambda=0.3))` will reset `eta` too, you have to still have your original choices around.

murrellb:
This should be easy, convenient, and visible (so that the user is aware of the possible footgun). That rules out 5.

> `Optimisers.adjust!(os, SignDecay(0.3))`, passing a whole new rule.

Will this replace a whole OptimiserChain, or just the rule that matches the type? The fact that I'm asking means that the user has to memorize/understand an extra convention to use this.

> Something more exotic, like adjust!(fun::Function, os, lambda=0.3) where this fun must return a Bool and gets... the rule?

I'm not sure what I'd do with this that you can't do with the current PR, but with (slightly) more difficulty.

Another option is allowing you to label an optimizer component, with a default that matches its type. So you can go:
```julia
os = Optimisers.setup(OptimiserChain(SignDecay(0.1, label = "a"), AdamW(lambda=0.1)), (; x=[12 34.]))
Optimisers.adjust!(os, "a", lambda = 0.3)
Optimisers.adjust!(os, "AdamW", lambda = 0.01) #"AdamW" is the default label for AdamW.
```

This would let you have two optimisers of the same type in a chain, but label them differently so you can adjust them independently (which my current PR doesn't allow). However I can't think of motivation for needing this, and this would mean we have to introduce changes to all the current optimisers (and might invalidate checkpointed opt states, depending on how it is done?).

Still prefer the current PR's solution, but I'm not picky as long as it is easy to use and visible.

mcabbott:
I agree 5 is messy. But the point is that the state tree is some simple nested Julia object, which you are free to manipulate however you like -- you do not have to work through any explicit API of this package, and hence the API does not have to expand to allow arbitrarily complex manipulations.

Adding labels seems way over the top to me.

> or just the rule that matches the type?

Obviously just the rule, surely? If you have setup `Momentum`, and do `adjust!(... AdamW...)` in any form, it can't replace the whole rule because it can't change the saved momenta -- that's the whole point of `adjust!` vs. a fresh `setup`.

murrellb:
> Obviously just the rule, surely? If you have setup Momentum, and do adjust!(... AdamW...) in any form, it can't replace the whole rule because it can't change the saved momenta -- that's the whole point of adjust! vs. a fresh setup.

This will only be "obvious" to a user who has both i) a very clear understanding of how this package does what it does, and ii) a mental model of the package devs. They also have to be very confident that there isn't some other reason (that they're currently unaware of) that doesn't eg. force them to have to re-specify the entire rule (eg. the package internals for some reason not handling automatic construction of the rest of the optimizer chain). Not obvious.

CarloLucibello:
I think the proposed solution is a light addition and intuitive addition to the interface and solves a concrete problem with no simple workaround, so we should go with it.

CarloLucibello (approved):

src/adjust.jl
```diff
@@ -144,3 +144,33 @@ function _adjust(r::T, nt::NamedTuple) where T <: AbstractRule
   end
   T(vals...)  # relies on having the default constructor
 end
+
+###
+#adjust with type control
+###
+
+adjust!(ℓ::Leaf, oT::Type, eta::Real) = (ℓ.rule = adjust(ℓ.rule, oT, eta); nothing)
```
ToucheSir, line 152: Can all the `Type`s here be changed to `Type{<:AbstractRule}`? It would provide some more information for users of the interface and ensure we're not promising support for arbitrary types.

src/adjust.jl
```diff
@@ -144,3 +144,33 @@ function _adjust(r::T, nt::NamedTuple) where T <: AbstractRule
...
+adjust!(ℓ::Leaf, oT::Type; kw...) = (ℓ.rule = adjust(ℓ.rule, oT; kw...); nothing)
+
+adjust(ℓ::Leaf, oT::Type, eta::Real) = Leaf(adjust(ℓ.rule, oT, eta), ℓ.state, ℓ.frozen)
+adjust(ℓ::Leaf, oT::Type; kw...) = Leaf(adjust(ℓ.rule, oT; kw...), ℓ.state, ℓ.frozen)
+
+adjust!(tree, oT::Type, eta::Real) = foreach(st -> adjust!(st, oT, eta), tree)
+adjust!(tree, oT::Type; kw...) = foreach(st -> adjust!(st, oT; kw...), tree)
+
+adjust(r::AbstractRule, oT::Type, eta::Real) = ifelse(isa(r, oT), adjust(r, eta), r)
```
ToucheSir, line 161: These overloads are allocating the output of `adjust(r, ...)` even if it doesn't end up being used. Unless `ifelse` brings a demonstrable type stability benefit over a plain conditional, I'd recommend using a ternary instead:
```suggestion
adjust(r::AbstractRule, oT::Type{<:AbstractRule}, eta::Real) = isa(r, oT) ? adjust(r, eta) : r
```

ToucheSir (commented):
Thanks for drafting this up. I think the only missing pieces are tests and an expanded docstring for `adjust[!]`?

src/adjust.jl
mcabbott: I would guess the branch is resolved when compiling?

I think you could also make dispatch do this, although whether it's a good idea I don't know:
```suggestion
adjust(r::C, oT::Type{A}, eta::Real) where {C <: A <: AbstractRule}  where {C<:AbstractRule} = adjust(r, eta)
adjust(r:: AbstractRule, oT::Type{<:AbstractRule}, eta::Real) = r
```

ablaom:
Please see also my "orthogonal" PR referenced above.

PR #203, opened 2024-12-21
Add Muon

murrellb:
This adds Muon (https://kellerjordan.github.io/posts/muon/), which uses an approximate orthogonalization before the update. There isn't a publication, but it gave a key improvement in the nanoGPT training "speedrun" attempt: https://github.com/KellerJordan/modded-nanogpt

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -646,7 +646,7 @@ function apply!(o::Muon, state, x::AbstractArray{T}, dx) where T
   else
     η, μ, λ = T(o.eta), T(o.mu), T(o.lambda)
     @.. state = μ * state + dx
-    Ot = _newton_schulz5(μ .* state .+ dx)
+    Ot = _newton_schulz5(μ .* state .+ dx) * T(sqrt(max(1, size(x,1)/nonfirstdims(x))))
     dx′ = @lazy η * (Ot + λ * x)
     return state, dx′
   end
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -624,13 +624,15 @@ In nanoGPT speedrun experiments, Muon is used for the internal layer >2D weights
 `Optimisers.adjust!(optimiser_state, η::Real)` will adjust the fallback optimizer's `eta` to `η * (opt.eta / eta)`, and Muon's `eta` to `η`, preserving their ratio,
 but `Optimisers.adjust!(optimiser, eta = η)` will only adjust Muon's learning rate (allowing you to adjust the fallback optimizer's learning rate separately).
 """
-@def struct Muon <: AbstractRule
-    opt = AdamW(eta = 0.0003, beta = (0.9,0.95), lambda = 0.01)
-    eta = 0.02
-    mu = 0.95
-    lambda = 0.01
-    fallback = Returns(false)
+struct Muon <: AbstractRule
+    opt::AbstractRule
+    eta::Float64
+    mu::Float64
+    lambda::Float64
+    fallback::Function
 end
+
+Muon(;opt = AdamW(eta = 0.0003, beta = (0.9,0.95), lambda = 0.01), eta = 0.02, mu = 0.95, lambda = 0.01, fallback = x -> false) = Muon(opt, eta, mu, lambda, fallback)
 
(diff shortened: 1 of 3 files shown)


src/rules.jl
```diff
@@ -599,14 +600,89 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+  end
+end
+
+function _newton_schulz5(G::AbstractMatrix{T}) where T
+    a, b, c = (T(3.4445f0), T(-4.7750f0), T(2.0315f0))
+    X = G / (norm(G) + eps(T))
+    transposed = size(G, 1) > size(G, 2)
+    if transposed
+        X = X'
+    end
```
mcabbott, line 663: I really don't like re-using X for different things, and sometimes the compiler feels the same way:
```suggestion
    X1 = G / (norm(G) + eps(T))
    transposed = size(G, 1) > size(G, 2)
    X2 = transposed ? transpose(X1) : X1
```

But also, I don't like creating type instability. So even better, the inner loop could be pulled out to another function:
```suggestion
    X = G / (norm(G) + eps(T))
    if size(G, 1) <= size(G, 2)
        _innerloop(X, a, b, c)
    else
        transpose(_innerloop(transpose(X), a, b, c))
    end
end

function _innerloop(X, a, b, c)
    for _ in 1:5
        A = X * X'
```

src/rules.jl
```diff
@@ -599,14 +600,89 @@ function apply!(o::AdaBelief, state, x::AbstractArray{T}, dx) where T
...
+- Learning rate (`η == eta`): Amount by which gradients are discounted before updating the weights
+- Momentum (`μ == mu`): Controls the acceleration of gradient descent in the prominent direction
+- Weight decay (`λ == lambda`): Controls the strength of ``L_2`` regularisation.
+- Fallback function (`fallback`): Function to control when, in addition to 1D arrays, the fallback optimizer should be used. Will be passed the parameter array and must return a boolean.
+
+Note: Works best with large batch sizes and may not be suitable for fine-tuning.
+In nanoGPT speedrun experiments, Muon is used for the internal layer >2D weights, and AdamW is used for the 1D weights, embeddings, and heads.
+
+`Optimisers.adjust!(optimiser_state, η::Real)` will adjust the fallback optimizer's `eta` to `η * (opt.eta / eta)`, and Muon's `eta` to `η`, preserving their ratio,
+but `Optimisers.adjust!(optimiser, eta = η)` will only adjust Muon's learning rate (allowing you to adjust the fallback optimizer's learning rate separately).
```
mcabbott, line 625: What it seems you really want is for `setup` to use `Muon` on some arrays, and `AdamW` on others. But instead this is rolled into this particular meta-optimisation rule, which is also called `Muon`.

One spelling would be `setup(OptimiserIfElse(fun, AdamW(), Muon()), model)` which does `fun(x) ? init(AdamW(), x) : init(Muon(), x)`, with some `stuct OptimiserIfElse <: AbstractRule`.

Another would be `setup(fun::Function, model::Any)` which is like
```julia
setup(model) do x
  ndims(x) == 1 ? AdamW() : Muon()
end
```
I'm sure we batted around such ideas when writing this package, but nobody had a concrete need.

murrellb: I strongly agree that this functionality would be good to have in general, but as you say it isn't quite clear what to switch on.

But in this case, when people say "Muon" they mean "Muon for >2D with an AdamW fallback for 1D" - see https://github.com/KellerJordan/Muon/blob/master/muon.py
And the only things they use to switch are directly inferable from the tensor itself, which we have access to inside the optimiser so we can get the same behavior. So I think it makes sense to, here, just call this Muon? Also, it supports the same (as the python version) way of adjusting the two eta values during warmup/cooldown (retaining their ratio) saving the user a little effort if they want to use it as-is.

murrellb:
FYI I've made this: https://github.com/MurrellGroup/CannotWaitForTheseOptimisers.jl which can house new optimisers until you folks are happy to have them included in Optimisers.jl (mostly because I really just need to be able to use them conveniently for now).

PR #204, opened 2024-12-21
Add `setup(::Function, model)`

mcabbott:
Quick sketch of one way to easily allow different rules for different arrays, by modifying `setup` -- see docstring.

### PR Checklist

- [ ] Tests are added

### DIFF
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -78,17 +78,28 @@ init
 
 """
     Optimisers.setup(rule, model) -> state_tree
+    Optimisers.setup(f::Function, model) -> state_tree
 
 Initialises the given optimiser for every trainable parameter within the model.
 Returns a tree of the relevant states, which must be passed to [`update`](@ref Optimisers.update)
 or [`update!`](@ref Optimisers.update!).
 
 # Example
-```jldoctest
-julia> m = (x = rand(3), y = (true, false), z = tanh);
+```jldoctest setup1
+julia> m = (x = rand(3), y = (true, false), z = randn(2, 2));
 
 julia> Optimisers.setup(Momentum(), m)  # same field names as m
-(x = Leaf(Momentum(0.01, 0.9), [0.0, 0.0, 0.0]), y = ((), ()), z = ())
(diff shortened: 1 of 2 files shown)


AntonOresten:
I think this is elegant and useful. I'm working on some improvements to #203. Muon is optimal for linear layers, but doesn't make as much sense for e.g. `Flux.Embedding`, even though it is linear-like, and since the linear decoder layer in LLMs is often tied to the input encoder layer, it's preferable to disable Muon for that layer as well. I imagine the cleanest way of differentiating between linear layers is with an `IdDict` inside the setup rule function. You'd for example create the function based on which layers are present in some `IdDict`, and in the same function embed rules for different array shapes.

AntonOresten:
One could do something like:
```julia
function fun_rule(model, rule=Muon(), fallback=Adam())
    skipped = Base.IdSet{Any}([model.encode.weight, model.decode.weight])
    fun(x::AbstractVector) = fallback
    fun(x::AbstractArray) = x in skipped ? fallback : rule
    return fun
end

opt_state = Optimisers.setup(fun_rule(model), model)
```
such that:
```julia
julia> model = (;
           encode=(; weight=rand(2,2)),
           other=(; weight=rand(2,2), bias=rand(2)),
           decode=(; weight=rand(2,2)));

julia> fun_rule(model)(model.encode.weight)
Adam(eta=0.001, beta=(0.9, 0.999), epsilon=1.0e-8)

julia> fun_rule(model)(model.other.weight)
Muon(0.02, 0.95, 0.01, 1.0e-7, true)

julia> fun_rule(model)(model.other.bias)
Adam(eta=0.001, beta=(0.9, 0.999), epsilon=1.0e-8)

julia> fun_rule(model)(model.decode.weight)
Adam(eta=0.001, beta=(0.9, 0.999), epsilon=1.0e-8)
```
I generally avoid closures, but this has a certain elegance to it. `Base.IdSet` is private, but the alternative is slightly cursed:
```julia
skipped = keys(IdDict([model.encode.weight, model.decode.weight] .=> nothing))
```

Issue #205, opened 2024-12-23
Type Constraints in the Rule Structs


avik-pal:
I was working on getting Optimisers to work with reactant, and it (mostly) does, but one of the current issues is that `eta` is forced to be Float64 in some of the structs.

But consider the following IR:

```julia
julia> pss = (; a = (rand(3) |> Reactant.to_rarray))
(a = ConcreteRArray{Float64, 1}([0.023549009580651203, 0.10813549621409191, 0.7874517465499301]),)

julia> st_opt = @allowscalar Optimisers.setup(opt, pss)
(a = Leaf(Adam(0.001, (0.9, 0.999), 1.0e-8), (ConcreteRArray{Float64, 1}([0.0, 0.0, 0.0]), ConcreteRArray{Float64, 1}([0.0, 0.0, 0.0]), (0.9, 0.999))),)

julia> @code_hlo Optimisers.update(st_opt, pss, pss)
module {
  func.func @main(%arg0: tensor<3xf64>, %arg1: tensor<3xf64>, %arg2: tensor<3xf64>) -> (tensor<3xf64>, tensor<3xf64>, tensor<3xf64>) {
    %cst = stablehlo.constant dense<1.000000e-03> : tensor<3xf64>
    %cst_0 = stablehlo.constant dense<1.000000e-08> : tensor<3xf64>
    %cst_1 = stablehlo.constant dense<0.0010000000000000009> : tensor<3xf64>
    %cst_2 = stablehlo.constant dense<0.99899999999999999> : tensor<3xf64>
    %cst_3 = stablehlo.constant dense<0.099999999999999978> : tensor<3xf64>
    %cst_4 = stablehlo.constant dense<9.000000e-01> : tensor<3xf64>
    %0 = stablehlo.multiply %cst_4, %arg0 : tensor<3xf64>
    %1 = stablehlo.multiply %cst_3, %arg2 : tensor<3xf64>
    %2 = stablehlo.add %0, %1 : tensor<3xf64>
    %3 = stablehlo.multiply %cst_2, %arg1 : tensor<3xf64>
    %4 = stablehlo.abs %arg2 : tensor<3xf64>
    %5 = stablehlo.multiply %4, %4 : tensor<3xf64>
    %6 = stablehlo.multiply %cst_1, %5 : tensor<3xf64>
    %7 = stablehlo.add %3, %6 : tensor<3xf64>
    %8 = stablehlo.divide %2, %cst_3 : tensor<3xf64>
[...]

PR #206, opened 2025-01-05
Reactant: add extension to prevent stackoverflow

wsmoses:
Add extension to prevent stackoverflow when using Reactant.

Origin of bug in https://github.com/FluxML/Fluxperimental.jl/pull/28

```
simple train!: Error During Test at /Users/wmoses/git/Fluxperimental.jl/test/reactant.jl:152
  Got exception outside of a @test
  StackOverflowError:
  Stacktrace:
    [1] mlirOperationCreate
      @ ~/git/Reactant.jl/src/mlir/libMLIR_h.jl:1005 [inlined]
    [2] create_operation(name::String, loc::Reactant.MLIR.IR.Location; results::Vector{Reactant.MLIR.IR.Type}, operands::Vector{Reactant.MLIR.IR.Value}, owned_regions::Vector{Reactant.MLIR.IR.Region}, successors::Vector{Reactant.MLIR.IR.Block}, attributes::Vector{Reactant.MLIR.IR.NamedAttribute}, result_inference::Bool)
      @ Reactant.MLIR.IR ~/git/Reactant.jl/src/mlir/IR/Operation.jl:319
    [3] constant(; output::Reactant.MLIR.IR.Type, value::Reactant.MLIR.IR.Attribute, location::Reactant.MLIR.IR.Location)
      @ Reactant.MLIR.Dialects.stablehlo ~/git/Reactant.jl/src/mlir/Dialects/StableHLO.jl:1134
    [4] constant
      @ ~/git/Reactant.jl/src/mlir/Dialects/StableHLO.jl:1126 [inlined]
    [5] constant(x::Array{Float32, 0}; location::Reactant.MLIR.IR.Location)
      @ Reactant.Ops ~/git/Reactant.jl/src/Ops.jl:74
    [6] constant(x::Array{Float32, 0})
      @ Reactant.Ops ~/git/Reactant.jl/src/Ops.jl:69
    [7] promote_to(::Type{Reactant.TracedRNumber{Float32}}, rhs::Int64)
      @ Reactant.TracedRNumberOverrides ~/git/Reactant.jl/src/TracedRNumber.jl:70
    [8] TracedRNumber
      @ ~/git/Reactant.jl/src/TracedRNumber.jl:56 [inlined]
    [9] convert
      @ ./number.jl:7 [inlined]
   [10] zero
      @ ./number.jl:309 [inlined]
   [11] float
      @ ./float.jl:311 [inlined]
   [12] _eps(T::Type{Reactant.TracedRNumber{Float32}}, e::Float64) (repeats 79970 times)
      @ Optimisers ~/.julia/packages/Optimisers/a4OnF/src/utils.jl:20
Test Summary:                           | Pass  Error  Broken  Total     Time
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,10 +13,12 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 [weakdeps]
 Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
+Reactant = "3c362404-f566-11ee-1572-e11a4b42c853"
 
 [extensions]
 OptimisersAdaptExt = ["Adapt"]
 OptimisersEnzymeCoreExt = "EnzymeCore"
+OptimisersReactantExt = "Reactant"
 
 [compat]
 Adapt = "4"
--- a/ext/OptimisersReactantExt.jl
+++ b/ext/OptimisersReactantExt.jl
@@ -1,0 +1,8 @@
+module OptimisersReactantExt
+
+import Optimisers
+import Reactant
+
+Optimisers._eps(T::Type{<:Reactant.TracedRNumber{<:AbstractFloat}}, e) = T(e)
+
+end


CarloLucibello:
For the problem with the learning rate in #205 instead, do we have to relax the type constraint?

wsmoses:
These are two different issues, but yes for that issue the learning rate will need to be relaxed. Essentially tracedrarrays/tracedrnumbers represent data being compiled. It is never convertible to a regular float (because during compile time the data doesn't exist, but it represents a potential input, or set of operations on that input). Thus if you do`f(TracedRValue)` for any f you'll get another traced return type (unless of course the function returns a constant or doesn't depend on the input).

For the learning rate, if you want the learning rate to be a parameter to the compiled function (and not baked into the compiled function), it needs to be a traced value. Of course you could alternatively have the learning rate be a regular Float64, but then the compiled function will always use whatever learning rate you passed in at compile time.

mcabbott:
On this issue alone, can you explain why this is needed? Or is there a write-up of the mental model we ought to have to think about such things?

(Also, the function `_eps` really needs a longer name, and a docstring or at least a comment explaining why it exists.)

mcabbott:
Do we really need an extension for this? The code is:
```
_eps(T::Type{<:AbstractFloat}, e) = T(e)
_eps(T::Type{<:Number}, e) = _eps(real(float(T)), e) 
```
written apparently assuming that `real(float(T))` must produce an AbstractFloat. That's apparently false here, but would also be false for say Dual numbers, or maybe unitful or who knows what else.

The second method could simply be this:
```
_eps(T::Type{<:Number}, e) = real(float(T))(e) 
```
which will make one attempt to convert & then give up. Why not do that?

wsmoses:
Yeah, honestly this is a bug in _eps here IMO.

So presently it is

```julia
_eps(T::Type{<:AbstractFloat}, e) = T(e)
# catch complex and integers
_eps(T::Type{<:Number}, e) = _eps(real(float(T)), e) 
# avoid small e being rounded to zero
_eps(T::Type{Float16}, e) = e == 0 ? T(0) : max(T(1e-7), T(e))
```

So essentially the float function takes whatever the type is and converts it into a floating point version of that type.

e.g. 

```julia
julia> float(3)
3.0

julia> float(Complex(3))
3.0 + 0.0im
```

We have a version of float that takes a traced number (e.g. TracedRNumber{Int} or TracedRNumber{Complex{Int}}` and this will convert it to a TracedRNumber{Float64} etc.

So the second case `_eps(T::Type{<:Number}, e) = _eps(real(float(T)), e)` if T is a TracedRNumber{Float64}, then `float(T)` will also be TracedRNumber{Float64}. Of course that is real so it's TracedRNumber{Float64} again. so literally the function 
`eps(TracedRNumber{Float64}, x) = eps(TracedRnumber{Float64}, x)` so it infinitely recurs.

Honestly if you want to catch the complex/int stuff, you could just special case it on int/complex instead of on number.

wsmoses:
I mean sure, go for it.

I'm not sufficiently into the insides of this package that I would've known that change to be okay (e.g. someone later down the line really assuming it needs to be a float).

Which.... eventually could be hit here: https://github.com/FluxML/Optimisers.jl/issues/205

mcabbott:
I guess I wish the description of what the problem actually is was in the first message (since it requires knowing nothing about fancy Reactant stuff, just number types) and that we don't try to merge the first draft at lightspeed.

Want to revert and simplify?

For #205, is there a clear writeup of all of this somewhere? Like at some point numbers became 0-dim arrays in some way that I didn't understand, maybe this has changed... what are we actually designing around?

wsmoses:
We no longer have that 0-dim array stuff. Instead now we just have two types TracedRNumbers [for numbers] and TracedRArrays [for arrays].

Essentially the same thing I said above applies to general traced types (and of course you can getindex a tracedrarray and get a tracedrnumber)

PR #207, opened 2025-01-05
Prevent recursion in _eps

wsmoses:
Alternate for https://github.com/FluxML/Optimisers.jl/pull/206

@mcabbott @CarloLucibello

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -15,8 +15,15 @@ end
 
 ofeltype(x, y) = convert(float(eltype(x)), y)
 
-_eps(T::Type{<:AbstractFloat}, e) = T(e)
-# catch complex and integers
+"""
+    _eps(Type{T}, val)
+
+Mostly this produces `real(T)(val)`, so that `_eps(Float32, 1e-8) === 1f-8` will
+convert the Float64 parameter epsilon to work nicely with Float32 parameter arrays.
+
+But for Float16, it imposes a minimum of `Float16(1e-7)`, unless `val==0`.
+This is basically a hack to increase the default epsilon, to help many optimisers avoid NaN.
+"""
 _eps(T::Type{<:Number}, e) = real(float(T))(e) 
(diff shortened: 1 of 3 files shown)


src/utils.jl
```diff
@@ -17,6 +17,6 @@ ofeltype(x, y) = convert(float(eltype(x)), y)
 
 _eps(T::Type{<:AbstractFloat}, e) = T(e)
 # catch complex and integers
```
mcabbott, line 19: Maybe we don't need the first method at all. But could add a docstring?
```suggestion
"""
    _eps(Type{T}, val)

Mostly this produces `real(T)(val)`, so that `_eps(Float32, 1e-8) === 1f-8` will
convert the Float64 parameter epsilon to work nicely with Float32 parameter arrays.

But for Float16, it imposes a minimum of `Float16(1e-7)`, unless `val==0`.
This is basically a hack to increase the default epsilon, to help many optimisers avoid NaN.
"""
```

src/utils.jl
```diff
@@ -17,6 +17,6 @@ ofeltype(x, y) = convert(float(eltype(x)), y)
...
-_eps(T::Type{<:Number}, e) = _eps(real(float(T)), e) 
+_eps(T::Type{<:Number}, e) = real(float(T))(e) 
 # avoid small e being rounded to zero
 _eps(T::Type{Float16}, e) = e == 0 ? T(0) : max(T(1e-7), T(e))
```
mcabbott, line 29: In which case we can also delete this comment.
```suggestion
_eps(T::Type{Float16}, e) = iszer(e) ? T(0) : max(T(1e-7), T(e))
```

or perhaps clearer
```suggestion
_eps(::Type{Float16}, e) = iszero(e) ? Float16(0) : max(Float16(1e-7), Float16(e))
```

mcabbott (approved):

mcabbott:
Thanks! Let's do this.

PR #210, opened 2025-03-08
feat: reactant support

avik-pal:
`Reactant` doesn't allow tracing conditionals directly right now. Add a hook rn to prevent reactant tracing.


src/interface.jl
```diff
@@ -294,6 +294,11 @@ macro def(expr)
         end |> esc
 end
 
+_assert_positive_eta(eta) = _assert_positive_eta(eta, eta < 0)
+function _assert_positive_eta(eta, cond::Bool)
+  cond && throw(DomainError(eta, "the learning rate cannot be negative"))
+end
```
ToucheSir, line 300: If this function is being overloaded anyways, could the `eta < 0` check not be moved inside `_assert_positive_eta`?

ToucheSir:
While I don't have any objections to this in principle, there are a couple of things to consider:
1. This is being added as internal API, which means we can remove it in any release. Can Reactant deal with that?
2. How many other places in the Optimisers.jl code need a similar change? If it's more than a handful, I think we need to step back and think about a more general solution (which may or may not live in Optimisers.jl itself).

avik-pal:
Added tests for the optimisers. We just need the assert check as a temporary means till we support throwing errors from inside MLIR

PR #224, opened 2026-02-24
Add  methods `adjust(state, rule)` and `adjust!(state, rule)`.

### LINKED ISSUE #223: Limitations of `adjust!` when there are nested field name collisions in chained optimisers
Here is a common use case for an optimiser which includes both L1 and L2 regularisation:

```julia
optimiser = OptimiserChain(
        SignDecay(),
        WeightDecay(),
        Adam(),
        )
# OptimiserChain(SignDecay(lambda=0.001), WeightDecay(lambda=0.0005), Adam(eta=0.001, beta=(0.9, 0.999), epsilon=1.0e-8))
```

Note this optimiser has *non-unique* nested field names, because both `SignDecay` and `WeightDecay` call their single parameter `:lambda`. 

Suppose, after updating `state = setup(optimiser, model)` for some epochs, I want to update the `:lambda` for `WeightDecay` to 0.3, without changing that of `SignDecay`. It appears the `adjust!(state; kwargs...)` mechanism cannot do this, because `adjust!(state, lambda=0.3)` adjusts *both* lambdas.  Is there public API to do what I want here?

Perhaps the API could be extended to provide `adjust!(state, new_optimiser)`, so that the whole of `new_optimiser` is replicated in `state`, the assumption being that `new_optimiser` has the same type as the `optimiser` appearing in the initialisation step above?

ablaom:
At present we have `adjust!(state, eta::Real)` to update a  learning rate, and there is a version with kwargs to update other (nested) parameters. This PR adds `adjust!(state, rule::AbstractRule)` and a non-mutating version, `adjust(state, rule::AbstractRule)` to allow updating the entire embedded rule, to resolve #223.

### PR Checklist

- [x] Tests are added

### DIFF
--- a/src/adjust.jl
+++ b/src/adjust.jl
@@ -101,15 +101,31 @@ julia> Optimisers.adjust(st2; beta = (0.777, 0.909), delta = 11.1)  # delta acts
 julia> Optimisers.adjust(st; beta = "no such field")  # silently ignored!
 (vec = Leaf(Nesterov(0.123, 0.9), Float32[-0.016, -0.088]), fun = ())
 ```
+
+The argument `η` may also be a new optimisation rule, of the same type as the rule
+currently embedded in the state tree, as in the next example:
+
+```
+julia> st = Optimisers.setup(Nesterov(0.123, 0.9), m)
+(vec = Leaf(Nesterov(eta=0.123, rho=0.9), Float32[0.0, 0.0]), fun = ())
+
+julia> Optimisers.adjust!(st, Nesterov(0.456, 0.8))
+
+julia> st
+(vec = Leaf(Nesterov(eta=0.456, rho=0.8), Float32[0.0, 0.0]), fun = ())
+```
 """
 adjust!(tree, eta::Real) = foreach(st -> adjust!(st, eta), tree)
 adjust!(tree; kw...) = foreach(st -> adjust!(st; kw...), tree)
+adjust!(tree, r::AbstractRule) = foreach(st -> adjust!(st, r), tree)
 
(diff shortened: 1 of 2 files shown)


ablaom:
Okay, I have now looked over #202, which seems to have different aim, namely convenience. Here I want complete control: Replace (in one hit)  the currently embedded rule (in general, an `OptimiserChain`) with a new one, remembering all the states. #202 does not appear to provide this. 

Fortunately, the current PR and #202 appear to offer non-colliding extensions of the current syntax: In #202  we are adding `adjust!(tree, oT::Type, eta::Real)`,  `adjust!(tree, oT::Type; kw...)`, 
`adjust(r::AbstractRule, oT::Type, eta::Real)`. This PR adds `adjust!(tree, r::AbstractRule)`, `adjust(state, rule::AbstractRule)`

I'm happy to rebase this PR off #202 to test there really are no method ambiguities, if there is support for this PR. 

If the internal layout of an optimiser state tree is public API (this wasn't obvious to me)  then I can live without this PR, but that doesn't seem to me the best solution.  By the way, my particular use case is https://github.com/FluxML/MLJFlux.jl/issues/322.

ablaom:
I've added a test that attempting to adjust the rule to one of different type throws an exception.

ablaom:
I realise there are a lot of open PRs, but could I get rough estimate of time for review?

Note this PR includes tests and doc updates.

CarloLucibello:
hi @ablaom, sorry for being very late here. This PR seems in good shape and could be merged. Did you give further thought to it?

ablaom:
Thanks @CarloLucibello for the feedback. I don't have anything to add.

PR #227, opened 2026-08-04
Add OptimisersReactantMLDataDevicesExt for Reactant-compatible optimisers

CarloLucibello:
Cc @avik-pal @wsmoses 

Upstreams Lux's `ReactantCompatibleOptimisers` shim into Optimisers.jl as a new weak-dependency extension `OptimisersReactantMLDataDevicesExt` (triggered on Reactant + MLDataDevices).

Adds a public `Optimisers.make_reactant_compatible(rule, dev)` that wraps a rule so its scalar hyper-parameters (learning rate, momenta, step counters) are stored as tracked on-device numbers rather than baked into the compiled program as constants — so `adjust!` works without recompilation. The existing `OptimisersReactantExt` (traced `apply!` for RAdam/AccumGrad) is unchanged; the two extensions cooperate.

Also in this PR:
- run the Reactant test suite unconditionally and add Windows/macOS to CI, since Reactant now supports Julia 1.12 and Windows
- declare Reactant + MLDataDevices in `test/Project.toml` (dropping the runtime `Pkg.add`)
- replace the `VERSION`-guarded `public` eval hack with `Compat.@compat public`

Verified locally on Julia 1.12.6: the new extension precompiles and its tests pass (Adam, RAdam, `OptimiserChain(AccumGrad, Descent)`, learning-rate-not-a-constant, `adjust!`), and the existing `reactant.jl` still passes.

Once released, Lux can drop its shim and call `Optimisers.make_reactant_compatible` directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


CarloLucibello:
Possibile followup, automatically detect reactant model in setup

PR #231, opened 2026-08-05
Implement Muon Optimizer

### LINKED ISSUE #230: Implement Muon Optimizer
### Motivation and description

Related to https://github.com/FluxML/Optimisers.jl/issues/213

The Muon optimizer, alongside its variants, are getting on the spotlight for their applications on LLM training (https://arxiv.org/abs/2502.16982).

I think it would be a nice addition to `Optimisers.jl` and it can set the precedent for other 2D optimisers, such as LiMuon.

**References:**
- https://kellerjordan.github.io/posts/muon/
- https://github.com/KellerJordan/Muon
- https://arxiv.org/abs/2502.16982

### Possible Implementation

The reference implementation (https://github.com/KellerJordan/Muon) can be used.

I've started development on a fork of mine https://github.com/LucasGrasso/Optimisers.jl and created a pull request #231.

LucasGrasso:
Implements the Muon optimizer as per the [reference implementation](https://github.com/KellerJordan/Muon).

Closes https://github.com/FluxML/Optimisers.jl/issues/230

This implementation follows the reference implementation closely, and allows only for 2D inputs.
Reshaping non 2D inputs could be a possibility but I think it shouldn't be the responsibility of the optimiser.

Regarding testing, a matrix-specific ruleset was created and a test was generalized for matrix input, whereas applicable tests test both the old `RULES` ruleset alongisde the new `MATRIX_RULES` one. Generalization of tests whenever possible is also a possibility.

**References:**
- https://github.com/KellerJordan/Muon
- https://kellerjordan.github.io/posts/muon/

### PR Checklist

- [x] Tests are added

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -21,6 +21,7 @@ Optimisers.NAdam
 Optimisers.AdamW
 Optimisers.AdaBelief
 Optimisers.Lion
+Optimisers.Muon
 ```
 
 In addition to the main course, you may wish to order some of these condiments:
--- a/src/Optimisers.jl
+++ b/src/Optimisers.jl
@@ -25,7 +25,8 @@ include("rules.jl")
 export Descent, Adam, Momentum, Nesterov, Rprop, RMSProp,
        AdaGrad, AdaMax, AdaDelta, AMSGrad, NAdam, AdamW, RAdam, OAdam, AdaBelief,
        WeightDecay, SignDecay, ClipGrad, ClipNorm, OptimiserChain, Lion,
-       AccumGrad, MixedPrecision
+    AccumGrad, MixedPrecision,
+    Muon
 
 """
     Optimisers.make_reactant_compatible(rule::AbstractRule, [dev]) -> rule
--- a/src/rules.jl
+++ b/src/rules.jl
@@ -21,7 +21,7 @@ struct Descent{T} <: AbstractRule
   eta::T
 end
 
-Descent(; eta=1f-1) = Descent(eta)
+Descent(; eta = 1f-1) = Descent(eta)
 
 init(o::Descent, x::AbstractArray) = nothing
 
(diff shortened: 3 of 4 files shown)


AntonOresten:
See also https://github.com/FluxML/Optimisers.jl/pull/203. That PR got some comments (e.g. regarding type stability) that may be relevant to this one too.

IIRC Muon has a few variants going around with e.g. different normalization factors and I think the PyTorch one deviated a bit at one point. There are some successors that made it closer to production like [NorMuon](https://arxiv.org/abs/2510.05491), [MuonClip](https://arxiv.org/abs/2507.20534), and [Aurora](https://arxiv.org/abs/2606.27715). A faithful port of the original reference is justifiable, but it is rarely a simple drop-in replacement for e.g. AdamW. There are also low-hanging fruit like [Polar Express](https://arxiv.org/pdf/2505.16932) for more robust orthogonalization that actually converges towards the polar factor w.r.t. step count.

Worth mentioning I made [a package](https://github.com/MurrellGroup/MatrixSign.jl) for orthogonalizing matrices with different methods like SVD, and quintic Newton-Schulz iteration last year (including Jordan Keller's and the Polar Express coefficients), and by [switching to mutating GEMMs and explicit buffers](https://github.com/MurrellGroup/MatrixSign.jl/blob/main/src/newtonschulz5/naive.jl), memory use goes down by an order of magnitude.

Moreover there's [an optimization](https://github.com/MurrellGroup/MatrixSign.jl/blob/main/src/newtonschulz5/fused.jl) in the Polar Express (J: INITIALIZATION FOR MATRICES WITH LARGE SPECTRAL GAPS) that can be done for rectangular matrices of higher aspect ratios.

IIRC one could even ~halve the number of FLOPs used by switching from GEMM to SYRK for the Gram matrices. I experimented with a https://github.com/JuliaGPU/cuTile.jl kernel for this at some point.

I don't wanna make perfect be the enemy of good. Just thought some extra context could be productive.

